PR #771, opened 2015-03-19
Include 'report_xform' when removing permissions

jnm:
Hello, this fixes `onadata.apps.main.views.set_perm()` so that it becomes possible to remove the `report_xform` permission. Thanks for your consideration.


jnm:
It looks like this was included as part of a larger merge 🎉

https://github.com/onaio/onadata/blame/32a777f56e566df185f0c8e03ebe117fe5646284/onadata/apps/main/tests/test_form_permissions.py#L227-L243

https://github.com/onaio/onadata/blame/32a777f56e566df185f0c8e03ebe117fe5646284/onadata/apps/main/views.py#L1175

PR #792, opened 2015-09-14
Small docs fix on Form ID

smn:
The `Form ID` in the documentation is actually the `id_string` and that tripped us up quite a bit during dev.


pld:
:+1:

smn:
Yay! I don't have commit flags so please merge :)

PR #816, opened 2016-10-05
validate password during email update

### LINKED ISSUE #812: Password validation should be enforced while changing an email
**Environment**
Preview/Chrome

**To reproduce**
- Enter an incorrect password so as to change your email 
- The email changes regardless of entering an incorrect password

**Expected result**
- Password should be correct before an email is changed.

denniswambua:
closes #812

### DIFF
--- a/onadata/libs/serializers/user_profile_serializer.py
+++ b/onadata/libs/serializers/user_profile_serializer.py
@@ -147,7 +147,7 @@ class UserProfileSerializer(serializers.HyperlinkedModelSerializer):
     def update(self, instance, validated_data):
         params = validated_data
 
-        # Check password if email is beig updated
+        # Check password if email is being updated
         if 'email' in params and 'password1' not in params:
             raise serializers.ValidationError(
                 _(u'Your password is required when updating your email '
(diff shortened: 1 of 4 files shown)


onadata/libs/serializers/user_profile_serializer.py
```diff
@@ -147,11 +147,14 @@ def _get_params(self, attrs):
     def update(self, instance, validated_data):
         params = validated_data
 
-        # Check password if email is beig updated
+        # Check password if email is being updated
         if 'email' in params and 'password1' not in params:
             raise serializers.ValidationError(
                 _(u'Your password is required when updating your email '
                   u'address.'))
+        password = params.get("password1")
```
pld: if you move this above line 151, you can have the second conjuct in that conditional be `password`

onadata/libs/serializers/user_profile_serializer.py
```diff
@@ -167,7 +170,7 @@ def update(self, instance, validated_data):
         instance.user.save()
 
         # force django-digest to regenerate its stored partial digests
-        update_partial_digests(instance.user, params.get("password1"))
+        update_partial_digests(instance.user, password)
```
pld: There's an error here that needs a test, if you add a test that passes no `email` and no `password` in params this should execute with `password` set to `nil` and set the partial digests to a bad state.

I'd suggest writing that test and then getting it passing by adding if `password` here

denniswambua: This has been addressed

onadata/libs/serializers/user_profile_serializer.py
denniswambua: addressed

skambo:
@denniswambua Theres validation, but the error msg is displayed as below when incorrect..

screen shot 2016-10-06 at 12 41 52 pm

Error when password is left blank

screen shot 2016-10-06 at 12 44 29 pm

pld:
That part is a zebra problem in handling the API response, this branch is
good to merge.

PR #826, opened 2016-10-18
include _ in the group delimeters

### LINKED ISSUE #825: Support under score group delimiter in exports
Linked dataset pull csv functionality does not work with current supported group delimiter `.` and `\`.
- [x] include `_` to the list

denniswambua:
closes #825 
Signed-off-by: Dennis Wambua dennis.wambua@gmail.com


onadata/libs/utils/export_tools.py
```diff
@@ -741,7 +741,7 @@ def parse_request_export_options(params):
     else:
         options["remove_group_name"] = False
 
-    if params.get("group_delimiter") in ['.', DEFAULT_GROUP_DELIMITER]:
+    if params.get("group_delimiter") in ['.', DEFAULT_GROUP_DELIMITER, '_']:
```
pld: why not use the new constant here?

denniswambua: will use constant

denniswambua:
Ready for second review

PR #842, opened 2016-11-11
Not have to send a request for each user we share a form with.

### LINKED ISSUE #840: XForm sharing should receive a list of dicts.
With regards to https://api.ona.io/static/docs/forms.html#share-a-form-with-a-specific-user

We can share in this way
`curl -X POST -d '{"username": "sluggie", "role": "readonly"}' http://localhost:8000/api/v1/forms/36/share -u urbanslug:pass -v -H "Content-Type: application/json"`

but it would be better if we could bundle many dicts in one lists in a vector/list for example:
`curl -X POST -d '[{"username": "urban", "role": "editor"}, {"username": "sluggie", "role": "readonly"}]' http://localhost:8000/api/v1/forms/36/share -u urbanslug:pass -v -H "Content-Type: application/json"`

urbanslug:
Allows us to send a request like so 

`curl -u urbanslug:<password> -X POST -d '{"username":"wolfie, sluggie, thor", "role":"editor"}' "http://localhost:8000/api/v1/forms/36/share" -H "Content-Type: application/json" -v`

closes #840


onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -474,12 +474,13 @@ def retrieve(self, request, *args, **kwargs):
     def share(self, request, *args, **kwargs):
         self.object = self.get_object()
 
-        data = {}
-        for key, val in request.data.iteritems():
-            data[key] = val
-        data.update({'xform': self.object.pk})
+        data_list = []
+        for username in request.data.get("users").split(','):
```
pld: make backwards compatible with `request.data.get("users", request.data.get("username")).split(',')`

pld:
documentation needs to be updated

onadata/apps/api/viewsets/xform_viewset.py
pld: also this will entirely throw an exception is that key does not exist

pld: I also think this should be called `usernames` since that's what it is :)

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -474,12 +474,17 @@ def retrieve(self, request, *args, **kwargs):
...
+        usernames_str = request.data.get("users", request.data.get("username"))
+        if not usernames_str:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
 
-        serializer = ShareXFormSerializer(data=data)
+        data_list = []
+        for username in usernames_str.split(","):
+            data_list.append({"xform":    self.object.pk,
+                              "username": username,
+                              "role":     request.data.get("role")})
```
pld: this only needs to be fetched once, even though it's O(1) no the fetch cost is O(n), n=number of users

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -474,12 +474,17 @@ def retrieve(self, request, *args, **kwargs):
...
-        for key, val in request.data.iteritems():
-            data[key] = val
-        data.update({'xform': self.object.pk})
+        usernames_str = request.data.get("users", request.data.get("username"))
+        if not usernames_str:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
 
-        serializer = ShareXFormSerializer(data=data)
+        data_list = []
+        for username in usernames_str.split(","):
```
pld: so iterable

pld:
@denniswambua what's up with that flakey e-tag test? do you know how to make it consistent

docs/metadata.rst
```diff
@@ -224,6 +224,21 @@ Link XForm or Dataview as a media example:
         "url": "https://api.ona.io/api/v1/metadata/7121.json"
         }
 
+Create xform meta permissions
```
pld: xform should be capitalized like `XForm`

docs/metadata.rst
```diff
@@ -224,6 +224,21 @@ Link XForm or Dataview as a media example:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Setting meta permissions for a specific form by passing two roles that are pipe delimited.
```
pld: Setting => Set

skambo:
@denniswambua On stage, the assigned permission roles are not reflecting

skambo:
@denniswambua While switching through permissions successfully for a while, at some point , the permissions stopped updating on zebra UI. The error is

screen shot 2016-11-23 at 11 50 51 am

skambo:
@denniswambua Noticed that at times the 'submitted by' field on the table view is empty after submitting via Enketo and this makes the permissions not get filtered correctly. Also if the dataset records are via a csv import, this 'submitted by' field will be empty.

Using ODK Collect though, the 'submitted by' field is always updated.

denniswambua:
@skambo I think will need to set require_auth  true on the form to ensure user authenticates when submitting.

pld:
@skambo yeah that's a known limitation not going to be solved other than by what @denniswambua suggested, so I wouldn't consider that a problem/blocker from merge.

pld:
the switching permissions problem, what's going on there? @skambo maybe after that's figured out it'll be good to merge?

skambo:
@pld new permissions were being created instead of being updated, so it was causing conflicts. just checking again on that, that if the permission exists, its updated.

pld:
Cool, reassign to you, if all looks well let's get this in

onadata/libs/models/share_project.py
```diff
@@ -28,6 +29,18 @@ def save(self, **kwargs):
...
+                        .filter(data_type='xform_meta_perms')
+                    if meta_perms:
+                        meta_perm = meta_perms[0].data_value.split("|")
+
+                        if len(meta_perm) > 1:
+                            if role == EditorRole:
+                                role = ROLES.get(meta_perm[0])
+
+                            elif role == DataEntryRole:
+                                role = ROLES.get(meta_perm[1])
```
ukanga: Would it be correct to say the original role gets overridden here? Let's say the role that had been applied was not a meta-permission, this seems to override those permissions when you have a form/project that has meta-permission. Probably a test to show that this is not the case would be helpful.

onadata/libs/permissions.py
```diff
@@ -357,10 +363,36 @@ def get_team_project_default_permissions(team, project):
...
+                ret_query = json.dumps(query)
+
+        except (ValueError, AttributeError):
+            query_list = list()
+            query_list.append({"_submitted_by": user.username})
+            query_list.append(query)
+
+            ret_query = json.dumps(query_list)
+    else:
+        ret_query = NoRecordsPermission()
```
ukanga: I think Exceptions are meant to be raised you should probably have a 
```python

raise NoRecordsPermission()
```

You can then catch the exception in `xform_viewset.py`  as:
```python
try:
    query = filter_queryset_xform_meta_perms_sql(self.get_object(), self.request.user, query)
except NoRecordsPermission:
    pass
```

onadata/apps/api/tests/viewsets/test_data_viewset.py
```diff
@@ -565,9 +639,17 @@ def test_data_start_limit_sort_json_field(self):
         self.assertEqual(response.status_code, 200)
         data = json.loads(''.join([c for c in response.streaming_content]))
         self.assertEqual(len(data), 2)
-        self.assertEqual(sorted([i['_uuid'] for i in data]),
-                         [u'5b2cc313-fc09-437e-8149-fcd32f695d41',
-                          u'9c6f3468-cfda-46e8-84c1-75458e72805d'])
+
+        data = {
```
denniswambua: The test was flakey and I had to change it to fix it.

denniswambua:
@skambo implemented the meta perms filter in the exports too

skambo:
@dwambua export restrictions work, this is the error displayed to a user who is not allowed to export data.

screen shot 2016-11-25 at 3 16 28 pm

PR #852, opened 2016-11-28
Support submitted by charts

### LINKED ISSUE #851: Add support for _submitted_by to charts endpoint
The form _submitted_by meta field is not supported by the charts endpoint, unlike the _submission_time.

denniswambua:
closes #851

### DIFF
--- a/onadata/libs/utils/chart_tools.py
+++ b/onadata/libs/utils/chart_tools.py
@@ -29,6 +29,15 @@ DATA_TYPE_MAP = {
     'calculate': 'numeric',
 }
 
+FIELD_DATA_MAP = {
+    common_tags.SUBMISSION_TIME:
+        ('Submission Time', '_submission_time', 'datetime'),
+    common_tags.SUBMITTED_BY:
+        ('Submission By', '_submitted_by', 'text'),
+    common_tags.DURATION:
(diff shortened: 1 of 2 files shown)


onadata/libs/utils/chart_tools.py
```diff
@@ -208,6 +208,10 @@ def build_chart_data_for_field(xform, field, language_index=0, choices=None,
         field_label = 'Submission Time'
         field_xpath = '_submission_time'
         field_type = 'datetime'
+    elif isinstance(field, basestring) and field == common_tags.SUBMITTED_BY:
```
pld: why check that this is a string?

ukanga: I think this is a pattern already laid out before, see most of the if statements already in the file, also a field tends to be an object most of the time.

pld: can we remove it? if it's an object this would fail in the 2nd conjunct, why's the first part of this needed?

pld:
@denniswambua we're also going to want to add `_duration`, is that included already, if not can you do that too?

The conditional blocks have a repeating pattern, it looks like there's probably a better way to abstract this.

denniswambua:
`_duration` has not been implemented let me add it.

PR #857, opened 2016-12-07
Allow "readonly" and "readonly no download" to edit widgets

### LINKED ISSUE #831: Allow readonly users to edit saved chart titles.
We want to fix an error on the widgets endpoint that forbids readonly users from editing saved chart metadata, since they are already able to save charts they should be able to edit the saved chart metadata.

urbanslug:
closes #831 
closes https://github.com/onaio/zebra/issues/4685

### DIFF
--- a/onadata/apps/api/permissions.py
+++ b/onadata/apps/api/permissions.py
@@ -8,6 +8,7 @@ from onadata.libs.permissions import (
     CAN_ADD_XFORM_TO_PROFILE,
     CAN_CHANGE_XFORM,
     CAN_DELETE_SUBMISSION)
+from onadata.libs.permissions import ReadOnlyRoleNoDownload
 
 from onadata.apps.api.tools import get_user_profile_or_none, \
     check_inherit_permission_from_project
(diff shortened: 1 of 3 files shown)


onadata/apps/api/permissions.py
```diff
@@ -277,6 +277,9 @@ def has_object_permission(self, request, view, obj):
                 isinstance(obj.content_object, DataView)):
             return False
 
+        if 'key' in request.query_params or view.action == 'partial_update':
```
pld: how does this work? is it still validating the user has permissions on the form this widget is attached to?

urbanslug: No it is not. I am looking for a way to find the user's permission in the parent Xform and restrict that in the creation and changing of the widget.

denniswambua:
Updated to check xform perms

onadata/apps/api/tests/viewsets/test_widget_viewset.py
```diff
@@ -210,6 +211,81 @@ def test_list_widgets(self):
...
+                              self.xform.pk,
+            'description': "Test widget",
+            'aggregation': "Sum",
+            'widget_type': "charts",
+            'view_type': "horizontal-bar",
+            'column': "age",
+            'group_by': ''
+        }
+
+        # to do: test random user with auth
```
pld: can we do this now?

denniswambua: done

denniswambua:
yes, i have added a test case for user without perms.

PR #858, opened 2016-12-08
4762 formbuilder phase 1

ivermac:


ivermac:
I have added a test for when invalid text xls form is submitted

pld:
You should have added that as a separate test

PR #864, opened 2017-01-03
validate org account when adding org members

### LINKED ISSUE #843: Add a check for whether an account username is for an org account or personal account
At the moment any username/account can be added as a member of an org, regardless of whether it is an org too or a personal account.

A user can add an org account as a member of another org which shouldn't be the case. The check will assist in verifying whether a particular username belongs to an org or personal a/c before adding as a member of an org.

denniswambua:
closes #843 
closes https://github.com/onaio/zebra/issues/4672

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
@@ -236,7 +236,7 @@ class TestOrganizationProfileViewSet(TestAbstractViewSet):
         response = view(request, user='denoinc')
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.data,
-                         {u'username': [u"This field is required."]})
+                         {u'username': [u"This field may not be null."]})
 
     def test_add_members_to_org_user_does_not_exist(self):
         self._org_create()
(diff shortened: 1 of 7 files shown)


onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
```diff
@@ -269,6 +271,28 @@ def test_add_members_to_org(self):
...
+
+        data = {'username': username}
+        request = self.factory.post(
+            '/', data=json.dumps(data),
+            content_type="application/json", **self.extra)
+
+        response = view(request, user='denoinc')
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data,
+                         u"Cannot add org account `second_inc` as member.")
```
ukanga: We probably should have this functionality and everything in members handled by serializers. I would not expect this response from the API, it usually is a JSON response and not a string.

skambo:
@denniswambua org accounts are validated, however, if you change the member roles, the change doesn't reflect.

denniswambua:
Migrating all members functionality to a serializer

denniswambua:
@skambo you can test again

skambo:
@denniswambua the good is there are now validation error messages if you try add an org as a member, however you still can't change member roles.
screen shot 2017-01-10 at 5 44 55 pm

skambo:
@denniswambua Creating an org is now working..

PR #871, opened 2017-01-12
Retrieve username from request object and not xform object

### LINKED ISSUE #870: Incorrect enketo preview url returned by core
The username variables in [this line](https://github.com/onaio/onadata/blob/master/onadata/apps/api/viewsets/xform_viewset.py#L370) and [this line](https://github.com/onaio/onadata/blob/master/onadata/apps/api/viewsets/xform_viewset.py#L382) should be the username in the request object.

ivermac:
closes #870

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -281,6 +281,14 @@ The following is a jquery code snippet on how to make a CORS request.
        },
    });
 
+Formlist for preview urls
+--------------------------
+To generate a preview url for enketo, use the following formlist url format
+::
+
+    https://api.ona.io/<username>/formList
+
+
 
 Quick start
 -----------
--- a/onadata/apps/api/tests/viewsets/test_xform_list_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_list_viewset.py
@@ -8,7 +8,9 @@ from django.core.urlresolvers import reverse
 
 from onadata.apps.api.tests.viewsets.test_abstract_viewset import\
     TestAbstractViewSet
-from onadata.apps.api.viewsets.xform_list_viewset import XFormListViewSet
+from onadata.apps.api.viewsets.xform_list_viewset import (
+    XFormListViewSet, PreviewXFormListViewSet
+)
 from onadata.apps.api.viewsets.project_viewset import ProjectViewSet
 from onadata.libs.permissions import DataEntryRole
 from onadata.libs.permissions import ReadOnlyRole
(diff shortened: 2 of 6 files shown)


ukanga:
I would have thought the correct Enketo URI is the owner of the form and not the requesting user. This has the likely hood of generating the wrong records in Enketo that may end up clashing with a users own forms. Suppose the user B is accessing a form from user A with the `id_string` `tutorial`, User B also has a form with `id_string` `tutorial`. Suppose that the two forms have different content, which form will be shown?

ivermac:
You make a good point and I agree with you that this approach needs to be reconsidered but at the same time preview urls are not working for users with `can view and download` as well as `can view` permissions - I tested this out after @msschroeder reported and I replicated the issue (`<id_string> not found in formlist` issue). What I have here fixes the preview url problem as illustrated in [enketo's preview url doc] (https://apidocs.enketo.org/v2#/get-survey-preview). 
As a side note, after working on the formbuilder, I think I now understand why kobo decided to go with the approach of generating an id_string for the user; the generated id_string contains random characters which seems to be unique not only in a person's account but the xform table as well. With this, one would never be worried about having duplicate id_strings in a formList even if forms from different accounts were shared with a particular user.
The other thing I had discussed with @royrutto and @denniswambua last week was that I don't see the reason why core/onadata is the middle service when zebra wants an enketo url or enketo preview url. We have `forms/<id>/enketo` endpoint which returns both an enketo url as well as an enketo preview url. For example, the request format for getting a preview url is the following:
`http -a <enketo-api-token>: POST "https://enketo.ona.io/api_v2/survey/preview?server_url=<user-specific-formlist-endpoint>&form_id=<id-string>"`
I think, but I stand to be corrected, zebra can acquire all the 3 variables above and make a request to enketo without having core make the request on zebra's behalf.

pld:
That's a good point, it looks like zebra could generate that on its own.

What are some examples of user-specific-formlist-endpoint?

ivermac:
An example would be `https://api.ona.io/pld/formList`

pld:
Cool I would say that yes zebra should definitely do this. It will reduce traffic and new faster. @ivermac can you create a zebra issue for this?

Regardless we should close this core issue so @ukanga please suggest a way forward for that.

ivermac:
@pld I have the [zebra issue](https://github.com/onaio/zebra/issues/4925)

ukanga:
@ivermac what is the reason that those with permissions to the form have no access to the form?

ivermac:
For context, @ukanga and I decided to add new formList endpoint that would be used only for enketo preview urls. This is because at the moment, the formlist endpoint only allows users without `can view` or `can view and download` permission to access that endpoint - it's a permission issue which @ukanga can give a better explanation if i'm not being clear.

msschroeder:
Is there any update when this issue will be deployed? The customer was asking when this will be resolved. I was under the impression that it would be deployed this week.

ivermac:
@msschroeder I can't really merge it since I worked on it. @denniswambua or @ukanga should be able to review this PR and give a status update.

msschroeder:
@denniswambua What else needs to be done with this? Can you give a timeline for when this will be done?

denniswambua:
@msschroeder Its ready for QA

ivermac:
@denniswambua I have added documenation

skambo:
`Can view` and `Can view and download` users now get Enketo preview URL

PR #874, opened 2017-01-14
Order repeat fields as they appear on XLSForms

ukanga:


denniswambua:
Tested in stage and looks good to me

PR #877, opened 2017-01-17
Add xform metadata for forms published by kpi

### LINKED ISSUE #876: Add info to show form was published by formbuilder
This info should be used by zebra to know if it should add a formbuilder edit link for a form or not. At the moment, it only makes sense to have edit links for forms that were published/deployed by the formbuilder.

ivermac:
closes #876

### DIFF
--- a/onadata/apps/api/viewsets/project_viewset.py
+++ b/onadata/apps/api/viewsets/project_viewset.py
@@ -34,6 +34,7 @@ from onadata.settings.common import (
     SHARE_PROJECT_SUBJECT)
 from onadata.apps.api.tools import get_baseviewset_class
 from onadata.libs.mixins.profiler_mixin import ProfilerMixin
+from onadata.apps.main.models.meta_data import MetaData
 
 
 BaseViewset = get_baseviewset_class()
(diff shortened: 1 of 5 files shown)


onadata/apps/api/viewsets/project_viewset.py
```diff
@@ -91,6 +92,13 @@ def forms(self, request, **kwargs):
 
                 serializer = serializer_cls(survey,
                                             context={'request': request})
+
+                published_by_formbuilder = request.data.get(
+                    'published_by_formbuilder'
+                )
+                if published_by_formbuilder == u'True':
```
denniswambua: There is a utility function for parsing Boolean. You should use it here

ivermac: Where do I find it? Also, `published_by_formbuilder` has string value, how will that utility function be helpful?

onadata/libs/serializers/project_serializer.py
```diff
@@ -168,12 +169,22 @@ class Meta:
...
+            'published_by_formbuilder',
             'last_submission_time',
             'date_created',
             'url',
             'last_updated_at'
         )
 
+    def get_published_by_formbuilder(self, obj):
+        md = obj.metadata_set.filter(
+            data_type='published_by_formbuilder'
```
denniswambua: You can use `unique_type_for_form` function from the metadata to retrieve the value. It has implemented this already

skambo:
@denniswambua pls review

ivermac:
Note: The management command should be tested on both stage and production. Testing involves ensuring that forms that were initially deployed using the formbuilder should have an edit url in zebra.

skambo:
@ivermac The edit button doesn't display for existing forms.  Also, the form builder questions added don't display on charts 

screen shot 2017-01-24 at 10 31 22 am

PR #882, opened 2017-01-19
Add share users to users when projects filtered by username

### LINKED ISSUE #869: List of projects filtered by owner/user lacks share users.
When we make the request described here https://api.ona.io/static/docs/projects.html#list-of-projects-filter-by-owner-organization `/api/v1/projects?owner=<username>` the projects list returned has members of the org under `teams` but doesn't have collaborators/share users i.e users with whom each project has been shared with. 

A way to get this would be to them make a request to `/api/v1/projects/<projectid>.json` for each object in the json array but it would be just doing more work on both the client and server.

Expectation:
- The request to projects filtering by username should include users with whom the project has been shared with (collaborators).

urbanslug:
List the collaborators when the projects endpoint is filtered by owner.

For example: `/api/v1/projects?owner=<username>` could have among it's users non-owners i.e collaborators
```
{...
 "projectid": 5
 "users": [{...}
           {"user": "sluggie"
            "role":   "readonly"}
           {...}]
...}
```

Compared to `api/v1/projects/<project-id>` which will not list collaborators.

closes #869

### DIFF
--- a/onadata/libs/serializers/project_serializer.py
+++ b/onadata/libs/serializers/project_serializer.py
@@ -209,11 +209,11 @@ class BaseProjectSerializer(serializers.HyperlinkedModelSerializer):
         return get_starred(obj, self.context['request'])
 
     def get_users(self, obj):
-        if 'request' in self.context \
-                and "owner" in self.context['request'].GET:
-            return get_users(obj, self.context)
-        else:
-            return get_users(obj, self.context, False)
+        owner_query_param_in_request = \
+          'request' in self.context and "owner" in self.context['request'].GET
+        return get_users(obj,
+                         self.context,
+                         owner_query_param_in_request)
 
     @profile("get_project_forms.prof")
     @check_obj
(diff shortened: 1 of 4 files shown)


onadata/libs/serializers/project_serializer.py
```diff
@@ -209,7 +209,11 @@ def get_starred(self, obj):
         return get_starred(obj, self.context['request'])
 
     def get_users(self, obj):
-        return get_users(obj, self.context, False)
+        if 'request' in self.context \
```
pld: why not use the conditional here to set the third arg to `get_users`? Like

```python
get_users(..., 'request' in self.context and ...)
```

urbanslug: Yeah that's clever.

pld:
@urbanslug can you check the docs to make sure they match with your changes and update them if needed?

skambo:
@urbanslug It's working well

PR #884, opened 2017-01-20
xml form upload

### LINKED ISSUE #848: Add the ability to upload XML XForms to the /api/forms endpoint
Modify the `/api/forms` endpoint to add the ability to accept XML XForms and create datasets from them.

Following the current pattern, https://api.ona.io/static/docs/forms.html#upload-xlsform, @ukanga suggests we add the parameter `xml_xform`.

Apparently this can be done via briefcase so this is a matter of refactoring and exposing the code that does that.

Docs need to be updated to reflect the new endpoint.

denniswambua:
closes #848

### DIFF
--- a/onadata/apps/main/forms.py
+++ b/onadata/apps/main/forms.py
@@ -255,9 +255,14 @@ class QuickConverterTextXlsForm(forms.Form):
         label=ugettext_lazy('XLSForm Representation'), required=False)
 
 
+class QuickConverterXmlFile(forms.Form):
+    xml_file = forms.FileField(
+        label=ugettext_lazy(u'XML File'), required=False)
+
+
 class QuickConverter(QuickConverterFile, QuickConverterURL,
                      QuickConverterDropboxURL, QuickConverterTextXlsForm,
-                     QuickConverterCsvFile):
+                     QuickConverterCsvFile, QuickConverterXmlFile):
     project = forms.IntegerField(required=False)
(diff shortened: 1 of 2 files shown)


onadata/apps/main/forms.py
```diff
@@ -289,15 +295,16 @@ def publish(self, user, id_string=None, created_by=None):
                     default_storage.save(
                         upload_to(None, rand_name, user.username),
                         ContentFile(csv_data))
-            else:
+            if 'xls_file' in self.cleaned_data and\
```
pld: if `xls_file in self.cleaned_data` is true, isn't `self.cleaned_data['xls_file']` also true? can't we remove the second part

onadata/apps/main/forms.py
```diff
@@ -324,6 +331,13 @@ def publish(self, user, id_string=None, created_by=None):
             else:
                 project = self._project
 
+            if 'xml_file' in self.cleaned_data and\
```
pld: same as above, if `xml_file in self.cleaned_data` is true, isn't `self.cleaned_data['xml_file']` also true? can't we remove the second part

denniswambua: The key might exist but have an empty string, thus the need for the second check.

pld: cool, since the same pattern is repeated you could move it to a helper function

PR #886, opened 2017-01-23
Easy install.

urbanslug:
Make running onadata easier.
- Update install instructions.
- Instructions for using Onadata with Postgres 9.6 running in a Docker container.

Depends https://github.com/onaio/docker-builds/pull/12


Makefile
```diff
@@ -1,7 +1,8 @@
...
-	@echo "Downloading ona core dependencies and running migrations for you (hopefully into a virtualenv)... \n (This will take a while, go grab a coffee or something.)\n\n"
-	python2 -m pip install -r requirements/makefile_dependencies.pip
-	python2 -m pip install -e .
-	python2 manage.py syncdb --noinput
-	python2 manage.py migrate
-	@echo "Start the app with \`python2 manage.py runserver --nothreading\`"
+  @echo "Downloading ona core dependencies and running migrations for you (hopefully into a virtualenv)... \n (This will take a while, go grab a coffee or something.)\n\n"
+  python2 -m pip install -r requirements/base.pip
+  python2 -m pip install -r requirements/dev.pip
+  # what does syncdb do?
```
pld: looks like nothing anymore, https://stackoverflow.com/questions/29683494/what-should-i-use-instead-of-syncdb-in-django-1-9. I think it's fine to remove these 2 lines

install.md
```diff
@@ -1,4 +1,4 @@
-# Ubuntu installation instructions
+# Installation instructions.
```
pld: I don't think you should change the title, that ubuntu script is still just for ubuntu, https://github.com/onaio/onadata/blob/master/script/install/ubuntu

install.md
```diff
@@ -1,4 +1,4 @@
...
 ## Prepare Os
```
pld: should be `OS` in caps

install.md
pld: Or, better probably, have a new sub section below this so

```
# Installation instructions

## Database Setup

### Using Ubuntu

[lines 3-12 without line 5]

### Using docker

[lines 15-42]
```

remove line 14

install.md
```diff
@@ -11,6 +11,36 @@ Replace username and db name accordingly.
     sudo su postgres -c "psql -d onadata -c \"CREATE EXTENSION IF NOT EXISTS postgis;\""
     sudo su postgres -c "psql -d onadata -c \"CREATE EXTENSION IF NOT EXISTS postgis_topology;\""
 
+**Alternatively** you can use docker to set up the DB.
```
pld: remove alternatively and make this a section `### Using docker`

install.md
urbanslug: Yeah I totally agree, however I am thinking of something like. 

### In base OS

### In Docker

Since Docker will be using ubuntu base image in this case.

Makefile
urbanslug: LOL how do I forget to remove these comments that I add as I read code 😁

install.md
pld: we can do that and mention on the line with the ubuntu install script that this line is only for ubuntu, otherwise we assume you've installed postgres already, that script is the only reason to keep ubuntu in the title

PR #888, opened 2017-01-24
catch the assert error

denniswambua:
Fixes csv export errors

Signed-off-by: Dennis Wambua <dennis.wambua@gmail.com>

### DIFF
--- a/onadata/libs/utils/csv_builder.py
+++ b/onadata/libs/utils/csv_builder.py
@@ -366,7 +366,11 @@ class CSVDataFrameBuilder(AbstractDataFrameBuilder):
                 # this dict
                 if type(item) is dict:
                     # order repeat according to xform order
-                    item = get_ordered_repeat_value(key, item)
+                    try:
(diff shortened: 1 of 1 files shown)


onadata/libs/utils/csv_builder.py
```diff
@@ -366,7 +366,11 @@ def get_ordered_repeat_value(xpath, repeat_value):
                 # this dict
                 if type(item) is dict:
                     # order repeat according to xform order
-                    item = get_ordered_repeat_value(key, item)
+                    try:
+                        item = get_ordered_repeat_value(key, item)
+                    except AssertionError:
+                        pass
```
pld: why bother throwing the assert error if we're not going to do anything?

denniswambua: I am not sure why it was added. I am removing it and test if breaks something

denniswambua:
@pld assert error removed

PR #893, opened 2017-01-27
Grant new users formbuilder perms

### LINKED ISSUE #892: Grant formbuilder object permissions on creation of a user profile
New users are the getting the following error when clicking the formbuilder link on zebra because they do not have object permissions. These permissions should be granted when new accounts/profiles are created.

ivermac:
closes #892

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
@@ -1,7 +1,11 @@
 import json
+import requests
 
 from mock import patch
 from django_digest.test import DigestAuth
+
+
+from httmock import all_requests, HTTMock
 
 from onadata.apps.api.tests.viewsets.test_abstract_viewset import\
     TestAbstractViewSet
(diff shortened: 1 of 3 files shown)


ivermac:
I can't seem to find a better way to test a `post_save` signal - there aren't any (or I haven't seen) any signal tests on the repo. Using `@patch` decorator on [this test case](https://github.com/onaio/onadata/blob/master/onadata/apps/main/tests/test_user_profile.py#L42) and checking if the mock was called i.e `self.assertTrue(mock.called)` failed for me - returned false. If anyone has a better way to handle this, please mention it here.

skambo:
@ivermac on stage, after creating a new acc, the error is still there

screen shot 2017-01-30 at 12 31 10 pm

skambo:
@ivermac @denniswambua It works now for new users

ivermac:
See the second comment on the PR

ivermac:
I think what should be tested is whether the signal function was called or not.

pld:
@ivermac​ sounds good, please add tests for what Dennis and you identified
as testable

PR #897, opened 2017-02-06
Use xpath as key for valueLabels dict

royrutto:
closes https://github.com/onaio/core/issues/1305


skambo:
@royrutto Some forms work, like the one you're testing with, but on this dataset (access with onasupport) https://stage.ona.io/onasupport/20009/140968# @faith-mutua and I get different errors

screen shot 2017-02-07 at 4 55 30 pm
 
Her error is
screenshot_020717_045538_pm

onadata/libs/utils/export_builder.py
```diff
@@ -737,10 +739,25 @@ def _get_sav_value_labels(self):
                     name = choice['name'].strip()
                     label = self.get_choice_label_from_dict(choice['label'])
                     _value_labels[name] = label.strip()
-                self._sav_value_labels[q['name']] = _value_labels
+                self._sav_value_labels[var_name or q['name']] = _value_labels
```
denniswambua: Should `var_name` be `q['name']` when `xpath_var_names` in None?

royrutto: yes

PR #902, opened 2017-02-08
Dataviews endpoint should not require auth for public dataviews.

### LINKED ISSUE #891: Public filtered datasets require auth.
Accessing filtered datasets requires auth even though they are both in a public project and created from a public form.

For example https://api.ona.io/api/v1/data/168803.json is accessible but https://api.ona.io/api/v1/dataviews/1466.json which is a filtered dataset created from the former and both being in a public project requires auth.

urbanslug:
The dataviews endpoint should not require auth for public dataviews, 
`api/v1/dataviews/<public-dataview-id>`, filtered by the dataview id but should require auth for a request for all dataviews, `api/v1/dataviews`.

closes #891

### DIFF
--- a/onadata/apps/api/permissions.py
+++ b/onadata/apps/api/permissions.py
@@ -235,10 +235,18 @@ class DataViewViewsetPermissions(AlternateHasObjectPermissionMixin,
 
     model_classes = [Project]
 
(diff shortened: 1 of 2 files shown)


urbanslug:
@denniswambua  Done.

PR #927, opened 2017-02-16
Add organization teams permissions to xform.

urbanslug:
Org members permissions are added to the project and transitively the xform.
However, this doesn't happen for org *teams* which has admins as the initial team (teams[0]) and every other element in teams. 

This adds the admins/owners team permissions to the xform permission.


skambo:
@urbanslug 
- An admin user still needs to be added explicitly on the projects page on the Zebra share modal, to be able to view/download the forms on Ona and ODK Collect

PR #933, opened 2017-02-27
CSV Export optimization

ukanga:


skambo:
@ukanga the CSV file is downloaded, but it's still blank

PR #938, opened 2017-03-02
Add org admins as owners in all projects

### LINKED ISSUE #924: Org projects should be shared with org members by default.
When a user is added to an org they should get related permissions in all org projects. For example, org admins should be owners of all org projects.

urbanslug:
Adding a person as an admin to an org should add them as an owner in all all org projects and removing them from the org should remove them as owners in all org projects.

closes #924

### DIFF
--- a/onadata/apps/api/tools.py
+++ b/onadata/apps/api/tools.py
@@ -27,6 +27,7 @@ from registration.models import RegistrationProfile
 from rest_framework import exceptions
 from taggit.forms import TagField
 
+from onadata.libs.models.share_project import ShareProject
 from onadata.libs.permissions import get_role
 from onadata.libs.permissions import is_organization
 from onadata.apps.api.models.organization_profile import OrganizationProfile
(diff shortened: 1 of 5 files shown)


urbanslug:
@denniswambua Thanks for the last commit to clear cache after removing a member.

onadata/apps/api/tools.py
```diff
@@ -172,6 +173,11 @@ def remove_user_from_organization(organization, user):
     owners_team = get_organization_owners_team(organization)
     remove_user_from_team(owners_team, user)
 
+    role = get_role_in_org(user, organization)
+    # Remove user from all org projects
+    for project in organization.user.project_org.all():
+        ShareProject(project, user.username, role, remove=True).remove_user()
```
ukanga: Should you be calling `ShareProject(project, user.username, role, remove=True).save()`, it will have the same effect and do any other cleanup required.

skambo:
@urbanslug it works well for org admins accessing projects and forms, but if you have no access to a project, a user can still access the form on the apps.

skambo:
@urbanslug It's working well now, please rebase.

PR #941, opened 2017-03-02
add open data viewset

### LINKED ISSUE #931: Add endpoint to retrieve tableau data
I had already started on this but hadn't created an issue for this. As discussed:
- The endpoint will be public so that users won't have to authenticate when they add the link to tableau's web data connector functionionality. 
- We will use a unique identifier - UUID in this case - on the new endpoint to retrieve data for a specific form.
- The object with the UUID attributed will have to be added and it can also be updated, activated, deactivated and deleted.

ivermac:
closes #931


onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,287 @@
...
+                if header == quest_name or header.endswith('_%s' % quest_name):
+                    append_to_tableau_colulmn_headers(header, quest_type)
+                    break
+            else:
+                append_to_tableau_colulmn_headers(header)
+
+        return tableau_colulmn_headers
+
+    def create(self, request, *args, **kwargs):
+        if request.user.is_anonymous():
```
ukanga: See my comment on permissions.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,287 @@
...
+            for quest_name, quest_type in self.flattened_dict.items():
+                if header == quest_name or header.endswith('_%s' % quest_name):
+                    append_to_tableau_colulmn_headers(header, quest_type)
+                    break
+            else:
+                append_to_tableau_colulmn_headers(header)
+
+        return tableau_colulmn_headers
+
+    def create(self, request, *args, **kwargs):
```
ukanga: This functionality should be handled by the serializer, considering this a `ModelViewSet`.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,287 @@
...
+            xform = self.object.content_object
+            qs_kwargs = {'xform': xform}
+            if gt:
+                qs_kwargs.update({'id__gt': gt})
+
+            instances = Instance.objects.filter(**qs_kwargs)
+            csv_df_builder = CSVDataFrameBuilder(
+                xform.user.username, xform.id_string, include_images=False
+            )
+            data = csv_df_builder._format_for_dataframe(
```
ukanga: I foresee a case where we will have issues with large dataset, and I suppose this also does not seem to support any form of paging.

onadata/libs/utils/common_tools.py
```diff
@@ -19,3 +20,10 @@ def get_boolean_value(str_var, default=None):
         return str_to_bool(str_var)
 
     return str_var if default else False
+
+
+def getUUID():
+    '''
+    Return a 32-character-long UUID
+    '''
+    return str(uuid.uuid1()).replace('-', '')
```
ivermac: Done. I have changed it to return `uuid.uuid4().hex`.

onadata/libs/serializers/open_data_serializer.py
```diff
@@ -0,0 +1,47 @@
...
+    object_id = serializers.IntegerField(required=False)
+    active = serializers.BooleanField(required=False)
+
+    class Meta:
+        model = OpenData
+        exclude = ('date_created', 'date_modified', 'content_type', 'id')
+
+    def create(self, validated_data):
+        name = validated_data.get('name')
+        clazz = validated_data.get('clazz')
```
ivermac: Done. I have changed it to `data_type`.

onadata/apps/api/viewsets/open_data_viewset.py
ivermac: Done.

onadata/apps/api/viewsets/open_data_viewset.py
ivermac: Done. I have set the value of `permission_classes` to `(OpenDataViewSetPermissions,)` - `OpenDataViewSetPermissions` is the new permissions class.

onadata/apps/api/viewsets/open_data_viewset.py
ivermac: I have added a `MAX_INSTANCES_PER_REQUEST` constant whose value is the maximum number of records returned per request. The current value is 1000.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,287 @@
...
+                DataInstanceSerializer(instances, many=True).data
+            )
+            data = replace_special_characters_with_underscores(data)
+
+        return Response(data)
+
+    def partial_update(self, request, *args, **kwargs):
+        if request.user.is_anonymous():
```
ivermac: Done. See my comment above.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,287 @@
...
+            return Response(data=data, status=status.HTTP_200_OK)
+
+        return Response(status=status.HTTP_404_NOT_FOUND)
+
+    @list_route(methods=['GET'])
+    def uuid(self, request, *args, **kwargs):
+        clazz = request.query_params.get('clazz')
+        object_id = request.query_params.get('object_id')
+
+        if not clazz or not object_id:
```
ivermac: The validation functionality from the above link is used when you are using a serializer class unless I misunderstood the docs. This action doesn't use the serializer class anywhere so I don't think we'll need the serializer's validation functionality or may be I'm missing the point.

ivermac:
@ukanga I have responded to each of your comments with the changes I have made. I usually add documentation once an endpoint has been approved because there might be changes during review - it's convenient to the implementer in my opinion.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,185 @@
...
+                return Response(status=status.HTTP_404_NOT_FOUND)
+
+            xform = self.object.content_object
+            qs_kwargs = {'xform': xform}
+            if gt:
+                qs_kwargs.update({'id__gt': gt})
+
+            instances = Instance.objects.filter(
+                **qs_kwargs
+            )[:self.MAX_INSTANCES_PER_REQUEST]
```
ukanga: How do you pick the next 1000 records?

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,185 @@
...
+                ),
+                'table_alias': xform.title
+            }
+
+            return Response(data=data, status=status.HTTP_200_OK)
+
+        return Response(status=status.HTTP_404_NOT_FOUND)
+
+    @list_route(methods=['GET'])
+    def uuid(self, request, *args, **kwargs):
```
ukanga: This seems to indicate that any authenticated user can get access to any form data as long as they have an object_id. This should not be the case, this should be only available to an authenticated user with permissions to the form and data.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -0,0 +1,185 @@
...
+                data="Query params data_type and object_id are required",
+                status=status.HTTP_400_BAD_REQUEST
+            )
+
+        if data_type == 'xform':
+            xform = get_object_or_404(XForm, id=object_id)
+            ct = ContentType.objects.get_for_model(xform)
+            _open_data = OpenData.objects.filter(
+                object_id=object_id, content_type=ct
+            ).first()
```
ukanga: I would suggest the use of `get_object_or_404` here.

onadata/apps/api/permissions.py
```diff
@@ -313,3 +313,14 @@ def has_object_permission(self, request, view, obj):
             return super(OrganizationProfilePermissions,
                          self).has_object_permission(
                 request=request, view=view, obj=obj)
+
+
+class OpenDataViewSetPermissions(IsAuthenticated):
```
ukanga: After the comment below, this should probably be extending `ViewDjangoObjectPermissions` or `DjangoObjectPermissionsAllowAnon`.

onadata/apps/api/viewsets/open_data_viewset.py
ivermac: You provide instance pk which is retrieved in [this line](https://github.com/onaio/onadata/pull/941/files#diff-15288ddaa4e9bd5c34d1480dc549a7c8R101). The pk is used to retrieve records which have a pk greater than the provided pk. This value is provided by tableau on refresh hence convenient in our case.

onadata/apps/api/viewsets/open_data_viewset.py
ivermac: Done. Checking if authenticated user has permissions to the form.

onadata/apps/api/viewsets/open_data_viewset.py
ivermac: Done.

onadata/apps/api/permissions.py
ivermac: Done. Using `DjangoObjectPermissionsAllowAnon` now.

onadata/apps/api/viewsets/open_data_viewset.py
ukanga: I am opposed to this approach, it does not make much sense to me. Please have a look at http://www.django-rest-framework.org/api-guide/pagination/. I can work on this with you when I get back to the office.

ivermac: Cool! I'll wait for you to get back.

PR #945, opened 2017-03-06
reject any malicious javascript in submissions

denniswambua:
Signed-off-by: Dennis Wambua <dennis.wambua@gmail.com>


ukanga:
Will this also catch an input that is HTML encoded? e.g `<age>34 &lt;script&gt;alert(document.cookie)&lt;/script&gt;</age` this does show a popup on classic view.

pld:
Is it the APIs responsibility to be concerned about this? I'm not convinced it is, if someone what's to upload script tags that's on them.

IMO the problems coming from this are caused by how the viewer incorrectly trusting the data submitted in forms.

PR #947, opened 2017-03-10
Add special handling for Instance Metadata deletions

geoffreymuchai:
Fix bug that occurs when deleting xform instance metadata objects due to permissions being handled at a form level and not an instance level


geoffreymuchai:
@denniswambua tests have been added

PR #958, opened 2017-03-16
export task progress

### LINKED ISSUE #951: Keep track of export progress
- pending, started % completed, success/failure

denniswambua:
closes #951

### DIFF
--- a/onadata/libs/utils/async_status.py
+++ b/onadata/libs/utils/async_status.py
@@ -3,14 +3,18 @@ from celery import states
 PENDING = 0
 SUCCESSFUL = 1
 FAILED = 2
+PROGRESS = 3
+RETRY = 4
+STARTED = 5
 
-status_msg = {PENDING: 'PENDING', SUCCESSFUL: 'SUCCESS', FAILED: 'FAILURE'}
+status_msg = {PENDING: 'PENDING', SUCCESSFUL: 'SUCCESS', FAILED: 'FAILURE',
+              PROGRESS: 'PROGRESS', RETRY: 'RETRY', STARTED: 'STARTED'}
 
 
 def celery_state_to_status(state):
-    status_map = {states.PENDING: PENDING, states.STARTED: PENDING,
-                  states.RETRY: PENDING, states.SUCCESS: SUCCESSFUL,
-                  states.FAILURE: FAILED}
+    status_map = {states.PENDING: PENDING, states.STARTED: STARTED,
+                  states.RETRY: RETRY, states.SUCCESS: SUCCESSFUL,
+                  states.FAILURE: FAILED, 'PROGRESS': PROGRESS}
     return status_map[state] if state in status_map else FAILED
 
 
--- a/onadata/libs/utils/export_builder.py
+++ b/onadata/libs/utils/export_builder.py
@@ -159,6 +159,12 @@ def dict_to_joined_export(data, index, indices, name, survey, row,
 
 
 def track_task_progress(additions, total=None):
+    """
+    Updates the current export task with number of submission processed
+    :param additions:
+    :param total:
+    :return:
+    """
     try:
         meta = {'progress': additions}
         if total:
--- a/onadata/libs/utils/export_builder.py
+++ b/onadata/libs/utils/export_builder.py
@@ -31,6 +31,7 @@ from onadata.libs.utils.mongo import _is_invalid_for_mongo,\
 # the bind type of select multiples that we use to compare
 MULTIPLE_SELECT_BIND_TYPE = u"select"
 GEOPOINT_BIND_TYPE = u"geopoint"
+DEFAULT_UPDATE_BATCH = 100
 
 
 def current_site_url(path):
@@ -169,7 +170,7 @@ def track_task_progress(additions, total=None):
     """
     try:
         if additions % getattr(settings, 'EXPORT_TASK_PROGRESS_UPDATE_BATCH',
-                               100) == 0:
+                               DEFAULT_UPDATE_BATCH) == 0:
             meta = {'progress': additions}
             if total:
                 meta.update({'total': total})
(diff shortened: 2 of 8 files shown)


onadata/libs/utils/export_builder.py
```diff
@@ -153,6 +158,23 @@ def dict_to_joined_export(data, index, indices, name, survey, row,
...
+    Updates the current export task with number of submission processed
+    :param additions:
+    :param total:
+    :return:
+    """
+    try:
+        meta = {'progress': additions}
+        if total:
+            meta.update({'total': total})
+        current_task.update_state(state='PENDING',
```
ukanga: Do we have any other started other than `pending`? For example, `started`? http://docs.celeryproject.org/en/latest/reference/celery.states.html#all-states

onadata/libs/utils/export_builder.py
```diff
@@ -542,6 +565,7 @@ def write_row(row, csv_writer, fields):
                             self.pre_process_row(child_row, section),
                             csv_writer, fields)
             index += 1
+            track_task_progress(i, total_records)
```
ukanga: Would it be better to make this updates in chunks/batches of 10 or 100 or X? Something configurable in settings. Could be an overkill when you are iterating over 10k or 100k records.

onadata/libs/utils/export_tools.py
```diff
@@ -163,8 +163,10 @@ def generate_export(export_type, xform, export_id=None, options=None,
     if options.get("dataview_pk"):
         dataview = DataView.objects.get(pk=options.get("dataview_pk"))
         records = dataview.query_data(dataview, all_data=True)
+        total_records = dataview.query_data(dataview, count=True)
     else:
         records = query_data(xform, query=filter_query, start=start, end=end)
+        total_records = xform.num_of_submissions
```
ukanga: Does this take into account the filter query? An export could be using filters where the number of records is less than the total number of submissions.

onadata/libs/utils/export_builder.py
denniswambua: addressed

onadata/libs/utils/export_tools.py
```diff
@@ -163,9 +163,17 @@ def generate_export(export_type, xform, export_id=None, options=None,
...
         dataview = DataView.objects.get(pk=options.get("dataview_pk"))
         records = dataview.query_data(dataview, all_data=True)
+        total_records = dataview.query_data(dataview,
+                                            count=True)[0].get('count')
     else:
         records = query_data(xform, query=filter_query, start=start, end=end)
 
+        if filter_query:
+            total_records = query_data(xform, query=filter_query, start=start,
+                                       end=end, count=True)[0].get('count')
```
ukanga: This makes a huge assumption that we will always receive a list of at least size one.

onadata/libs/utils/async_status.py
```diff
@@ -3,14 +3,18 @@
 PENDING = 0
 SUCCESSFUL = 1
 FAILED = 2
+PROGRESS = 3
+RETRY = 4
+STARTED = 5
```
pld: what's the difference between this and `PROGRESS`?

onadata/libs/utils/export_builder.py
```diff
@@ -153,6 +158,26 @@ def dict_to_joined_export(data, index, indices, name, survey, row,
...
+    Updates the current export task with number of submission processed.
+    Updates in batches of settings EXPORT_TASK_PROGRESS_UPDATE_BATCH defaults
+    to 100.
+    :param additions:
+    :param total:
+    :return:
+    """
+    try:
+        if additions % getattr(settings, 'EXPORT_TASK_PROGRESS_UPDATE_BATCH',
+                               100) == 0:
```
pld: I'd put this in a `DEFAULT_UPDATE_BATCH` var, @ukanga what do you think?

onadata/libs/utils/export_builder.py
pld: you could also only batch for large datasets, i.e. if it's < 10k, do not batch

onadata/libs/utils/export_tools.py
pld: well, since count is `True`, looks like this will be a list of length 1 unless the query results in a generator, in which case maybe you'd need to call `len` on it? Is this guaranteed to be on line 15 of the `apps.viewer.models.parsed_instance` function? Not related, but why does the condition on lines 10-11 of `apps.viewer.models.parsed_instance` exist at all? this would put a lot of objects in memory and probably segfault on large datasets

onadata/libs/utils/async_status.py
denniswambua: Between `PENDING`? `PENDING` is a unkown state, I have added `PROGRESS` to indicate that the task has started and is processing.
Docs: http://docs.celeryproject.org/en/latest/reference/celery.states.html#all-states

onadata/libs/utils/export_builder.py
ukanga: I agree.

onadata/libs/utils/async_status.py
pld: that makes sense, I am asking about the difference between, `PROGRESS` and `STARTED`

denniswambua: There is no big difference  `STARTED` means the job has started and `PROGRESS` (custom task state) is how the task is progressing. `STARTED` is fired when the task is started but is cleared and we require a custom state when updating the task meta field.

pld:
Is there a reason not to combine these into a single state constant?

denniswambua:
I introduced `PROGRESS` mainly because of http://docs.celeryproject.org/en/latest/reference/celery.states.html#celery.states.state

Especially here
```
>>> state('PROGRESS') > state(STARTED)
True

>>> state('PROGRESS') > state('SUCCESS')
False
```

pld:
cool, got it

PR #965, opened 2017-03-22
project soft delete

### LINKED ISSUE #959: Implement project soft deletion

denniswambua:
closes #959 
And fixes xform soft delete bug


onadata/apps/logger/models/project.py
```diff
@@ -93,6 +96,23 @@ def clean(self):
...
 
+    def soft_delete(self):
+        """
+        Soft deletes a project by adding a deleted_at timestamp and renaming
+        the project name by adding a deleted-at and timestamp.
+        Also soft deletes the associated forms.
+        :return:
+        """
+
+        soft_deletion_time = datetime.now()
```
ukanga: You probably should use a timezone aware `datetime` since we are saving this to the database,  we make use of this throughout the code base.
```python
from django.utils import timezone

soft_deletion_time = timezone.now()
```

ukanga: Please apply the same fix on `xform.soft_delete()`

PR #974, opened 2017-03-28
Handle XLSFormError exceptions on form parial_update

### LINKED ISSUE #953: Unhandled XLSFormError on form title edit
```python
Request Method: PATCH
Request URL: https://api.ona.io/api/v1/forms/[pk].json
Django Version: 1.9.5
...
File "./onadata/apps/logger/models/xform.py" in save
  764.             self._set_title()

File "./onadata/apps/logger/models/xform.py" in _set_title
  741.             raise XLSFormError(_("Title shouldn't have any invalid xml "

Exception Type: XLSFormError at /api/v1/forms/[pk].json
Exception Value: Title shouldn't have any invalid xml characters ('>' '&' '<')
Request information:
```

ukanga:
Catching `XLSFormError` exception and raising `ParseError` allows rest-framework to return a response with `status_code` 400 instead of failing with a `500` `status_code`.

Fixes #953

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_xform_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_viewset.py
@@ -1389,7 +1389,7 @@ class TestXFormViewSet(TestAbstractViewSet):
             view = XFormViewSet.as_view({
                 'patch': 'partial_update'
             })
-            title = u'مرحب'
+            title = u'Hello & World!'
             description = 'DESCRIPTION'
             data = {'public': True, 'description': description, 'title': title,
                     'downloadable': True}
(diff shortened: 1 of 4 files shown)


onadata/apps/api/tests/viewsets/test_xform_viewset.py
```diff
@@ -1389,7 +1389,7 @@ def test_partial_update(self):
             view = XFormViewSet.as_view({
                 'patch': 'partial_update'
             })
-            title = u'مرحب'
```
pld: are we testing that Arabic characters work somewhere else? Is that worth having?

ukanga: I think there is plenty of places where Unicode characters are tested. I don't think it's worth having here.

PR #989, opened 2017-04-06
[DNM] XForm version limits

### LINKED ISSUE #982: Limit form versions to 10 digits to match form design guidelines
See https://opendatakit.org/help/form-design/guidelines/

ukanga:
Fixes #982 

We probably should not merge this until the form builder is generating valid versions.


onadata/apps/logger/models/xform.py
```diff
@@ -124,6 +124,16 @@ class DuplicateUUIDError(Exception):
     pass
 
 
+def create_xform_version(release=1):
+    date_part = datetime.utcnow().strftime("%Y%m%d")
+
+    if release is None:
+        release = 1
+    release_part = str(release).zfill(2)[:2]
```
pld: I personally like this better, but am happy w/it as is if you do not:

```python
release_part = str(release or 1).zfill(2)[:2]
```

onadata/apps/logger/models/xform.py
```diff
@@ -124,6 +124,16 @@ class DuplicateUUIDError(Exception):
     pass
 
 
+def create_xform_version(release=1):
```
pld: this needs a docstring

onadata/apps/logger/models/xform.py
```diff
@@ -616,15 +626,33 @@ def _check_version_set(self, survey):
         the default version in this datetime (yyyymmddhhmm) format.
         """
 
+        change_version = False
         # get the json and check for the version key
         survey_json = json.loads(survey.to_json())
-        if not survey_json.get("version"):
+        version = survey_json.get("version")
+        if version is None:
             # set utc time as the default version
```
pld: I think this comment should be moved to the `create_xform_version` docstring

onadata/apps/logger/models/xform.py
```diff
@@ -616,15 +626,33 @@ def _check_version_set(self, survey):
...
+            change_version = True
+            version = create_xform_version()
+        else:
+            try:
+                int(version)
+            except ValueError:
+                raise XLSFormError(
+                    _(u"version should be a string of upto 10 numbers that "
+                      "describe this revision. You could use a string of the "
+                      "form 'yyyymmddrr'. For example, 2017021501 is the 1st "
```
pld: Is `rr` a standard? I think some text describing that is in order since I don't expect to see that in what looks like a date format string.

Also, if we think this is the best I think we should be more forceful then "could", e.g. replacing the sentance beginning with "You could" to address both:

> Please use a string of the form 'yyyymmddrr', where `rr` is the revision number with a leading zero if less than 10.

onadata/apps/logger/models/xform.py
```diff
@@ -616,15 +626,33 @@ def _check_version_set(self, survey):
...
+                      "revision from Feb 15th, 2017."))
+            else:
+                if len(version) > 10:
+                    change_version = True
+
+        if change_version:
+            survey_json['version'] = version[:10]
```
pld: What case is this handling? Would it perhaps to also throw an error here? I'm thinking of someone who mistakenly using a 16 digit version with hours, minutes, and seconds in their format. We'd silently truncate it and end up w/overlapping versions

pld:
@ukanga yea agreed, if it's the case formbuilder doesn't generate versions in this format, would we end up with any forms formbuilder tries to generate being rejected? We'll need to allocate time for someone to modify formbuilder to use this version format, is there a timeline for when we want to get this in?

onadata/apps/logger/models/xform.py
ukanga: This I picked from the XLSForm documentation https://github.com/XLSForm/xlsform.github.io/pull/82.

pld:
Cool, I still think it's good to explain it

pld (dismissed):

PR #1004, opened 2017-04-11
switch to using django storage for csv import processing

### LINKED ISSUE #1002: CSV imports do not work when celery runs on a different server
This is due to the use of a temporary file that is not accessible from a different server.

ukanga:
Fixes #1002


msschroeder:
@ukanga The CSV import works. It now shows the total number of submissions. Though it did take a long time to load. However, after I hit refresh, it displayed the number of total submissions. The Map view now displays the questions and allows you to filter the data

The SAV export worked fine. 
The CZIP did not work.

ukanga:
The CSV ZiP issue you faced has to do with zebra, @royrutto is creating an issue for it.

PR #1007, opened 2017-04-19
upgrade to django 1.11

ukanga:


onadata/apps/viewer/tests/test_export_builder.py
```diff
@@ -455,8 +455,10 @@ def test_xls_export_with_hxl_adds_extra_row(self):
         wb = load_workbook(temp_xls_file.name)
         children_sheet = wb.get_sheet_by_name("hxl_example")
         self.assertTrue(children_sheet)
+
         # we pick the second row because the first row has xform fieldnames
-        hxl_row = [a.value for a in children_sheet.rows[1]]
+        rows = [row for row in children_sheet.rows]
+        hxl_row = [a.value for a in rows[1]]
```
pld: This change definitely looks like if would have longer run time

onadata/libs/serializers/project_serializer.py
```diff
@@ -228,15 +227,21 @@ def get_users(self, obj):
                          self.context,
                          owner_query_param_in_request)
 
-    @profile("get_project_forms.prof")
     @check_obj
     def get_forms(self, obj):
+        forms = cache.get('{}{}'.format(PROJ_FORMS_CACHE, obj.pk))
```
pld: Is cache being cleared whenever the xforms change?

onadata/settings/common.py
```diff
@@ -411,8 +408,8 @@ def configure_logging(logger, **kwargs):
 after_setup_logger.connect(configure_logging)
 
 GOOGLE_STEP2_URI = 'http://ona.io/gwelcome'
-GOOGLE_CLIENT_ID = '617113120802.onadata.apps.googleusercontent.com'
-GOOGLE_CLIENT_SECRET = '9reM29qpGFPyI8TBuB54Z4fk'
+GOOGLE_OAUTH2_CLIENT_ID = '617113120802.onadata.apps.googleusercontent.com'
+GOOGLE_OAUTH2_CLIENT_SECRET = '9reM29qpGFPyI8TBuB54Z4fk'
```
pld: This is a public secret?

onadata/apps/logger/models/widget.py
```diff
@@ -84,10 +83,13 @@ def query_data(cls, widget):
...
             field_label = get_field_label(field)
 
-        columns = [SimpleField(field="json->>'%s'" % unicode(column),
-                               alias='"{}"'.format(column)),
-                   CountField(field="json->>'%s'" % unicode(column),
-                              alias='"count"')]
+        columns = [
+            SimpleField(
+                field="json->>'%s'" % unicode(column),
+                alias='{}'.format(column)), CountField(
```
ukanga: This is the work of yapf, I miss them sometimes.

onadata/apps/viewer/tests/test_export_builder.py
ukanga: Not longer really, this is in tests. `children_sheet.rows` due to package upgrade is now a generator/iterator and we can no longer use an index to get the second row.

pld:
Cool

onadata/settings/common.py
ukanga: We override this in deployment, I will check if empty strings do not affect tests.

pld:
Are these all non-active IDs and secrets then?

If empty strings aren't OK, we can use something like 'REPLACE ME'

onadata/settings/common.py
ukanga: `REPLACE_ME` will be more ideal.

onadata/libs/serializers/project_serializer.py
ukanga: Yes, whenever permissions change or a form is deleted. Check [`XForm`]( https://github.com/onaio/onadata/blob/master/onadata/apps/logger/models/xform.py#L945) model

onadata/apps/viewer/tests/test_export_builder.py
pld: OK agree doesn't matter since in tests, nevermind. But still has 2n vs n runtime

PR #1019, opened 2017-05-15
Allow form version filtering on exports

### LINKED ISSUE #1020: Version filters on exports is returning blank files
What's happening is when exporting forms using form version filters and you switch between different versions, the file that is exported is empty as the version value being passed is blank.

ivermac:
closes #1020


onadata/apps/viewer/models/parsed_instance.py
```diff
@@ -165,7 +178,7 @@ def _get_sort_fields(sort):
 
 def get_sql_with_params(xform, query=None, fields=None, sort=None, start=None,
                         end=None, start_index=None, limit=None, count=None):
-    records = _get_instances(xform, start, end)
+    records, query = _get_instances(xform, start, end, query)
```
ukanga: I would have thought this is already being handled by `get_where_clause` even though the current implementation will return a JSON field filter rather than an `Instance.version` field filter. The following changes may be required for this field to be used.

```patch
index 38a64120d..475286a13 100644
--- a/onadata/apps/viewer/parsed_instance_tools.py
+++ b/onadata/apps/viewer/parsed_instance_tools.py
@@ -7,7 +7,8 @@ from onadata.libs.utils.common_tags import MONGO_STRFTIME
 KNOWN_DATES = ['_submission_time']
 NONE_JSON_FIELDS = {
     '_submission_time': 'date_created',
-    '_id': 'id'
+    '_id': 'id',
+    '_version': 'version'
 }
 
 
@@ -55,8 +56,12 @@ def _parse_where(query, known_integers, or_where, or_params):
                 else:
                     where_params.extend((field_key, unicode(_v)))
         else:
-            where.append(u"json->>%s = %s")
-            where_params.extend((field_key, unicode(field_value)))
+            if field_key in NONE_JSON_FIELDS:
+                where.append("{} = %s".format(NONE_JSON_FIELDS[field_key]))
+                where_params.extend([unicode(field_value)])
+            else:
+                where.append(u"json->>%s = %s")
+                where_params.extend((field_key, unicode(field_value)))
 
     return where + or_where, where_params + or_params
```

PR #1034, opened 2017-05-30
Update Postgres install docs to use docker-compose.

urbanslug:
Make the Postgres setup simpler and easy to follow.

Depends https://github.com/onaio/docker-builds/pull/14

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -43,16 +43,9 @@ Connect to postgres using psql with:
 
     psql -h localhost -p 5432 -U postgres
 
-In psql:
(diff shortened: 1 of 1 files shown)


urbanslug:
@pld  I would argue that they are right above the docker instructions starting from L14 https://github.com/onaio/onadata/pull/1034/files#diff-62c330cd8da31d476bb63b51bb8b1307R14

pld:
I stand corrected

ukanga:
@urbanslug I think next is a dockerfile and docker compose for onadata itself. Thanks for this.

PR #1056, opened 2017-06-16
Add export permission to allow delete

ivermac:


onadata/apps/api/permissions.py
```diff
@@ -57,6 +57,30 @@ class ViewDjangoObjectPermissions(DjangoObjectPermissions):
     }
 
 
+class ExportDjangoPermission(IsAuthenticated,
```
ukanga: Can you rename this to `ExportDjangoObjectPermission` to be consistent with the others.

PR #1071, opened 2017-07-22
Proper submission count on CSV imports

### LINKED ISSUE #1065: CSV Imports, update submission count correctly after successful import
- [x] update `XFrom.num_of_submissions` when csv import completes
- [x] update accordingly after every X(lets start with 100/50?) records.
- [x] if an exception happens, send email to admins for further analysis

ukanga:
closes #1065

### DIFF
--- a/onadata/libs/utils/csv_import.py
+++ b/onadata/libs/utils/csv_import.py
@@ -1,5 +1,6 @@
 import cStringIO
 import json
+import sys
 import uuid
 from collections import defaultdict
 from copy import deepcopy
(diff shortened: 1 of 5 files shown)


onadata/libs/utils/logger_tools.py
```diff
@@ -1,59 +1,48 @@
...
 from onadata.apps.logger.xform_instance_parser import (
-    InstanceEmptyError,
-    InstanceInvalidUserError,
-    InstanceMultipleNodeError,
-    NonUniqueFormIdError,
-    DuplicateInstance,
-    clean_and_parse_xml,
-    get_uuid_from_xml,
-    get_deprecated_uuid_from_xml,
-    get_submission_date_from_xml)
```
pld: I actually prefer this style, but either consistently is fine, would be nice to define this in the style checker

ukanga: This is the work of [yapf](https://github.com/google/yapf), I will add a config for it.

onadata/libs/utils/csv_import.py
```diff
@@ -266,29 +269,44 @@ def submit_csv(username, xform, csv_file):
...
-                        meta={
-                            'progress': additions,
-                            'total': num_rows,
-                            'info': addition_col
-                        })
-                except:
-                    pass
+                if additions % getattr(settings,
+                                       'EXPORT_TASK_PROGRESS_UPDATE_BATCH',
+                                       100) == 0:
```
ukanga: Noted, will set a static variable for this.

onadata/libs/utils/logger_tools.py
ukanga: Using the default pep8 style I think.

Issue #1069, opened 2017-07-21
Instance.get_notes() does not check if a created_by is null


ukanga:
Results in an exception on data edits with notes that are not linked to any user. This is hidden by the `AttributeError` caught in the `xform_submission_viewset.py`.

PR #1081, opened 2017-08-03
Check for and handle cases where created_by is None for a Note object

moshthepitt:
Fixes: #1069

We now return an empty string in place of created_by when a Note is not
linked to any user.

### DIFF
--- a/onadata/apps/logger/tests/models/test_note.py
+++ b/onadata/apps/logger/tests/models/test_note.py
@@ -1,0 +1,22 @@
+from onadata.apps.main.tests.test_base import TestBase
+from onadata.apps.logger.models import Instance, Note
+
+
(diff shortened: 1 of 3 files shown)


onadata/apps/logger/tests/models/test_note.py
```diff
@@ -0,0 +1,22 @@
...
+        instance = Instance.objects.first()
+        note = Note(
+            instance=instance,
+            instance_field="",
+            created_by=None,
+        )
+        note.save()
+        try:
+            note.get_data()
+        except AttributeError:
```
ukanga: Could we have checked that we get an empty string for `owner` instead? I don't think `try ...  except` in tests is a good pattern.

Issue #1057, opened 2017-06-19
Add merged datasets functionality


ukanga:
Allows data from two or more forms to be represented as one form/endpoint.

The model will inherit from existing xform model. The difference are:  
- It will have its schema be a created from a combination of N+ xforms
- It will not accept submissions

*Requirements*
- [x] Add functionality in forms endpoint for creating merged dataset (may not be required)
- [ ] Generate Schema from existing xforms
- [x] Query data from child xforms
- [x] Use existing endpoints (`projects`, `forms`, `data`) to access/modify the form
- [x] Allows exports
- Integrations
  - [ ] Webhooks
  - [ ] Google sheets
  - [ ] Rapidpro
- [ ] HXL support
- [x] Will reflect changes (updates/deletions) in the parent datasets

urbanslug:
# Implementation plan<a id="sec-1" name="sec-1"></a>

## Merged datasets<a id="sec-1-1" name="sec-1-1"></a>

### To do:<a id="sec-1-1-1" name="sec-1-1-1"></a>

-   Add a MergedXForm model
    -   [X] Subclass the XForm model
    -   [X] Extend it with an \`xforms\` field

-   Add MergedXFormViewset
    -   [X] Add a merged-dataset endpoint

-   Add MergedXFormSerializer
    -   [X] GET Create - takes {"xforms": ["list-of-xform-urls"], "name": "form-name-string", "project": "url-to-parent-project"}
    -   [X] POST Retreive a user's merged datasets - list and detail (data endpoint)
    -   [ ] PATCH Update - we can submit to a merged xform
    -   [ ] Expose an endpoint to retrieve merged dataset JSON and XML
    -   [ ] Expose a merged-dataset data endpoint to give data for specific merged dataset

-   Validate that: (serializer validation and model before save)
    -   [ ] name is a required field
    -   [ ] project is a required field
    -   [ ] A merged XForm is made of at least two xforms

-   Add permissions and authourization
    -   [ ] Inherit permissions from the project

urbanslug:
### `ManyToManyField` v `OneToMany` for `xforms` in MergedXForm model.

---

##### OneToMany field
In this case django would create an merged xform relation with a foreign key field to reference a specific xform. It would however not be able to store more then one xform id. This means a merged xform would only reference a single xform. However, what we want is for a single merged xform to reference an arbitrary number of xforms.

`xform table`

| id    |  other | xform  | columns  | 
|-----|:------:|:------:|:---------:|
[...]

PR #1082, opened 2017-08-06
add support for merged datasets

ukanga:
Implements merged datasets as described in #1057


ivermac:
Hey @ukanga I was testing out the the `last_submission_time` on merged dataset and it seems not to be returned. On further investigation, it seems a merged dataset object is [created](https://github.com/onaio/onadata/compare/merged-datasets#diff-ab8d7875ff4cf8338ac5076d56ff3d47R119) before setting the last_submission_time variable - which I believe was what [this function](https://github.com/onaio/onadata/compare/merged-datasets#diff-ab8d7875ff4cf8338ac5076d56ff3d47R87) was meant to do. We could add a utility function that returns the last submission time outside the serializer class (borrowed this snippet from what is in the `get_last_submission_time` function):
```python
def get_last_submission_time(xforms):
    values = [
        x.last_submission_time
        for x in xforms
        if x.last_submission_time
    ]
    if values:
        return sorted(values, reverse=True)[0]
```

Then inside the `create` serializer function, we could set the `last_submission_time` in the `validated_data` variable:
```python
...
validated_data['last_submission_time'] = get_last_submission_time(
            validated_data['xforms']
        )
```

The above snippets seem to work for me but if there is a better implementation, that would be great.

ukanga:
The idea was that it will be calculated on the fly. I will check on it.

ukanga:
@ivermac I have made the changes on the forms endpoint, 8692573.

ivermac:
Cool, thanks @ukanga. I'll check it out

ivermac:
I would like `is_merged_dataset` field to be added to [this line](https://github.com/onaio/onadata/blob/merged-datasets/onadata/libs/serializers/project_serializer.py#L154) to allow zebra to only list non-merged forms when a user wants to select forms that can be merged i.e. the form listing on the screenshot below should only have normal forms - no dataviews or merged datasets.

ukanga:
@ivermac I will add the field.

onadata/apps/api/viewsets/charts_viewset.py
```diff
@@ -53,21 +62,23 @@ def get_content(self, renderer, data,
         return content
 
 
-class ChartsViewSet(AnonymousUserPublicFormsMixin,
-                    AuthenticateHeaderMixin,
-                    CacheControlMixin,
-                    ETagsMixin,
+# pylint: disable=R0901
+class ChartsViewSet(AnonymousUserPublicFormsMixin, AuthenticateHeaderMixin,
```
ukanga: Pylint `R0901` has to do with "Too many ancestors". 

The formatting at this time was under the mercy of `yapf` command, I tend to take its recommendations. I should probably look into and check-in yapf's configuration.

ukanga:
This PR now includes service integrations support for merged datasets.

onadata/libs/serializers/merged_xform_serializer.py
```diff
@@ -0,0 +1,179 @@
...
+
+    merged_xform_dict = json.loads(xforms[0].json)
+    children = merged_xform_dict.pop('children')
+    merged_xform_dict['children'] = []
+
+    intersect = set(xform_sets[0]).intersection(*xform_sets[1:])
+    intersect = set([__ for (__, ___) in intersect])
+
+    merged_xform_dict['children'] = _get_elements(children, intersect)
+    del merged_xform_dict['_xpath']
```
ivermac: @ukanga This could be re-written as
```python
...
merged_xform_dict.get("_xpath") and merged_xform_dict.pop("_xpath")
```
On stage and locally, saving a merged dataset triggers the following error
screen shot 2017-08-21 at 14 19 18

Issue #1077, opened 2017-08-02
Users without permission to public project cannot export data


ivermac:
### Environment (local, stage, preview, production)
All

### Expected behavior
- A public project should allow users without permission to it to export and download data
- Exporting data when it’s public should not require authentication

### Actual behavior
- A public project does not allow users without permission to it to export and download data

### Steps to reproduce the behavior
Access a public project and try to download data
---

### Implementation plan

- 
- 
- 

### Blocking or related [onadata](https://github.com/onaio/onadata/) issues

I'm tagging @KipSigei here so that when the issue is done, he can be alerted

moshthepitt:
Fix seems to not work as expected.

moshthepitt:
Please dont close this.  I'll close it.

JohnMwashuma:
@moshthepitt If the authorization header is present the api returns an empty list

PR #1085, opened 2017-08-07
Allow unrestricted exports on public project

moshthepitt:
Fixes: #1077

Ensure that exports of public projects are allowed for all users whether
they are authenticated or not.

### DIFF
--- a/onadata/apps/api/permissions.py
+++ b/onadata/apps/api/permissions.py
@@ -16,6 +16,7 @@ from onadata.apps.api.tools import get_user_profile_or_none, \
 from onadata.apps.logger.models import XForm, Instance
 from onadata.apps.logger.models import Project
 from onadata.apps.logger.models import DataView
+from onadata.apps.viewer.models.export import Export
 
 SAFE_METHODS = ('GET', 'HEAD', 'OPTIONS')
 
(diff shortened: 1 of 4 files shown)


onadata/apps/api/permissions.py
```diff
@@ -62,6 +63,17 @@ class ExportDjangoObjectPermission(IsAuthenticated,
                                    ViewDjangoObjectPermissions):
 
     def has_permission(self, request, view):
+        export_pk = view.kwargs.get('pk', None)
```
ukanga: I think this check is in the wrong place, it should be at the `has_object_permission` level. `has_permission` only applies when you are requesting for the export list, in this case the request is for downloading a specific export.

onadata/apps/api/viewsets/export_viewset.py
```diff
@@ -38,7 +40,7 @@ class ExportViewSet(DestroyModelMixin, ReadOnlyModelViewSet):
     permission_classes = [ExportDjangoObjectPermission]
 
     def retrieve(self, request, *args, **kwargs):
-        export = self.get_object()
+        export = get_object_or_404(self.queryset, pk=kwargs.get('pk'))
```
ukanga: Why? I would think `self.get_object()` does exactly this, I will say this is straying from the way Django rest framework does it.

moshthepitt: So my decision was guided by two things:

1.  In DRF's documents, their retrieve method example uses get_object_or_404, here: http://www.django-rest-framework.org/api-guide/viewsets/

2.  I wasn't able to get self.get_object() to work in my test.  It consistently failed whereas get_object_or_404 just worked

Issue #1068, opened 2017-07-20
If an export has been pending for more than 6 hours mark as failed


pld:
I'm fine with adjusting 6 hours, but we have some exports in the pending state for months, after a hard max pending cut-off time we should stop the celery process and mark long exports as failed.

ukanga:
Also apply to failed exports.

PR #1086, opened 2017-08-09
Cancel or delete stale pending and failed exports

moshthepitt:
Fixes: #1068

Create periodic tasks which:

- Find old pending exports and mark them as failed.

- Find old failed exports and delete them

The 'oldness' is determined by a configurable setting.

### DIFF
--- a/onadata/apps/viewer/tasks.py
+++ b/onadata/apps/viewer/tasks.py
@@ -1,6 +1,8 @@
 import sys
+from datetime import datetime, timedelta
 
 from celery import task
+from celery.task.schedules import crontab
 from django.conf import settings
 from django.shortcuts import get_object_or_404
 from requests import ConnectionError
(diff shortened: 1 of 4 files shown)


onadata/apps/viewer/tasks.py
```diff
@@ -349,3 +351,39 @@ def delete_export(export_id):
...
+@task.periodic_task(
+    run_every=(crontab(hour='*/7')),
+    ignore_result=True
+)
+def delete_old_failed_exports():
+    """
+    Delete old failed exports
+    """
+    h = settings.EXPORT_TASK_LIFESPAN
+    time_threshold = datetime.now() - timedelta(hours=h)
```
ukanga: You should probably use the time zone aware `django.utils.timezone.now()`.
The `h` could probably be more descriptive like `task_lifespan`.

onadata/apps/viewer/tasks.py
```diff
@@ -349,3 +351,39 @@ def delete_export(export_id):
...
+    exports = Export.objects.filter(internal_status=Export.FAILED,
+                                    created_on__lt=time_threshold)
+    for export in exports:
+        delete_export.delay(export.id)
```
ukanga: Is there a reason you chose to delete the export in a different task considering this is already task? Possible use `queryset` deletion, `exports.delet()`.

onadata/apps/viewer/tests/test_tasks.py
```diff
@@ -38,3 +40,11 @@ def test_create_async(self):
             self.assertTrue(export.id)
             self.assertIn("username", options)
             self.assertEquals(options.get("id_string"), self.xform.id_string)
+
+    def test_check_pending_exports(self):
+        result = check_pending_exports.delay()
+        self.assertTrue(result)
```
ukanga: We should probably check that the function deletes exports and not that a task is created.

onadata/apps/viewer/tasks.py
moshthepitt: I had seen that there was a task created to specifically delete exports that was used elsewhere, and so I chose to use it - to keep things consistent.

onadata/apps/viewer/tests/test_tasks.py
moshthepitt: Agree.  And implemented.

Issue #1049, opened 2017-06-08
Keep track of whether all attachments have been submitted.


ukanga:
- [ ] have a toggle field that will state if all attachments have been received.
- [ ] add a data filter on the API data endpoint for this field

PR #1089, opened 2017-08-15
Keep track of whether all attachments have been submitted

moshthepitt:
Fixes: #1049

Add new fields to the Instance model to keep track of:

- How many attachments, if any are expected in a submission
- How many attachments have been actually received
- Whether all attachments have been received


onadata/libs/tests/utils/test_logger_tools.py
```diff
@@ -24,6 +36,68 @@ def test_generate_content_disposition_header(self):
...
+                                   media_files=[media_file])
+        self.assertFalse(instance.json[MEDIA_ALL_RECEIVED])
+        self.assertEquals(instance.json[TOTAL_MEDIA], 2)
+        self.assertEquals(instance.json[MEDIA_COUNT], 1)
+        self.assertEquals(instance.json[TOTAL_MEDIA], instance.total_media)
+        self.assertEquals(instance.json[MEDIA_COUNT], instance.media_count)
+        self.assertEquals(instance.json[MEDIA_ALL_RECEIVED],
+                          instance.media_all_received)
+        xml2_string = """
+        <data id="{}">
```
ukanga: You should use the same submission XML as above, you can add the `instanceID`, I see no need for the two.

onadata/libs/utils/logger_tools.py
```diff
@@ -196,6 +195,15 @@ def save_attachments(xform, instance, media_files):
             mimetype=content_type,
             extension=extension)
 
+    instance.total_media = len([m for m in instance.get_dict().keys() if m in
```
ukanga: Could `198-206` be refactored to a function that does this?

ukanga: I think this may not handle images in a repeat. the structure is slightly different. I will suggest looping through each and every key in the submission JSON.

moshthepitt:
@ukanga we now deal with repeats recursively

Issue #983, opened 2017-04-03
SAV Exports failing due to duplicate name in the same choice list name


bmalina:
Env Production

SAV Exports failing for a form due a duplicate name in the same choice list.

![image](https://cloud.githubusercontent.com/assets/20789710/24555849/41fb022c-163b-11e7-83bd-6ee0c23b9cc8.png)

CSV and Excel export working as expected.
Form can be accessed [here](https://ona.io/forcier_south_sudan/32199/190327#) 

Expected Results

SAV exports should ignore the duplicate and export as expected or give an error indicating the failure.

moshthepitt:
Form where issue is reproducible: https://ona.io/acme/37113/230306#

PR #1092, opened 2017-08-17
Prevent SAV Exports From Failing When Choices Lists Have Duplicate Names

moshthepitt:
Prevent the export builder from failing when data from a form with
multiple choices that have duplicate names by ignoring all but one of the
choices that have a duplicated names.

Fix: #983

### DIFF
--- a/onadata/apps/viewer/tests/test_export_builder.py
+++ b/onadata/apps/viewer/tests/test_export_builder.py
@@ -713,7 +713,7 @@ class TestExportBuilder(TestBase):
         |         | animals   | 4    | Sheep      |
         |         | animals   | 5    | Donkey     |
         |         | animals   | 6    | Hen        |
-        |         | animals   | 6    | Other      | <--- duplicate name
+        |         | animals   | 6    | Other      | # <--- duplicate name
         |         | animals   | 8    | None       |
         |         | animals   | 9    | Don't Know |
         """
--- a/onadata/apps/viewer/tests/test_export_builder.py
+++ b/onadata/apps/viewer/tests/test_export_builder.py
@@ -646,7 +646,7 @@ class TestExportBuilder(TestBase):
                        returnHeader=True) as reader:
             rows = [r for r in reader]
             self.assertTrue(len(rows) > 1)
-            self.assertEqual(rows[1][0],  "1")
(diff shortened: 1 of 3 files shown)


onadata/libs/utils/export_builder.py
```diff
@@ -206,6 +206,22 @@ def track_task_progress(additions, total=None):
...
+
+def get_element_from_xpath(xpath, elements):
+    return next((el for el in elements if el['xpath'] == xpath), None)
+
+
+def get_element_type_from_xpath(xpath, elements):
+    element = get_element_from_xpath(xpath, elements)
+    if element:
+        return element['type']
+    return "unknown"
```
ukanga: "unknown" should be constant.

We probably should not return "unknown",  returning None or empty string would be more consistent elements in XForm dict.

onadata/apps/viewer/tests/test_export_builder.py
```diff
@@ -701,6 +701,45 @@ def test_zipped_sav_export_with_values_split_select_multiple(self):
 
         shutil.rmtree(temp_dir)
 
+    def test_zipped_sav_export_with_duplicate_name_in_choice_list(self):
```
ukanga: The mark down does not seem to show the duplicate name, how is this identified?

onadata/libs/utils/export_builder.py
```diff
@@ -911,7 +934,12 @@ def _is_numeric(xpath, element_type, data_dictionary):
             [(_var_types[item],
                 0 if item in ['_id', '_index', '_parent_index',
                               SUBMISSION_TIME] else 255)
-                for item in self.EXTRA_FIELDS]
+                for item in self.EXTRA_FIELDS] +
+            [(x[1],
+              0 if _is_numeric(x[0],
+                               get_element_type_from_xpath(x[0], elements),
+                               self.dd) else 255)
```
pld: can we do a find and search to replace `255` with a constant (under the no-magic-numbers justification)? I see at least once place right above here, line 936, seems reasonable to call this `MAX_TEXT_LENGTH` if that's what this denotes

moshthepitt:
@ukanga @pld the latest push addresses your points above

PR #1094, opened 2017-08-18
Add Instance Model Fields As Data Endpoints Filters

moshthepitt:
Add the ability to filter data endpoints using Instance model fields.
Done through a django-filter FilterSet.  Implement filters for the
following fields:

- version
- date_created
- date_modified
- status

Fix: #1048


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -347,6 +347,12 @@ def list(self, request, *args, **kwargs):
             tags = self.request.query_params.get('tags')
             not_tagged = self.request.query_params.get('not_tagged')
 
+            self.object_list = filters.InstanceFilter(
```
ukanga: I would have expected this to be added to the `DataViewSet.filter_backends` property. Why is this not the case?

moshthepitt: DataViewSet is set up to deal primarily with XForm objects.  If we added InstanceFilter to the filter_backends property, then it would filter on XForm objects, which is not what we want.

The Instance objects are dealt with in the list method of DataViewSet - which is also the same place where Data filters are implemented, and this is where (and why) we are adding InstanceFilter at that point.

PR #1102, opened 2017-08-24
Return appropriate filename for URL based media

ukanga:


onadata/libs/serializers/xform_serializer.py
```diff
@@ -406,4 +410,14 @@ def get_filename(self, obj):
         if len(parts) > 2:
             filename = u'%s.csv' % parts[2]
```
pld: not a fan of this extra line break

onadata/libs/serializers/xform_serializer.py
```diff
@@ -406,4 +410,14 @@ def get_filename(self, obj):
...
 
+        else:
+            try:
+                URLValidator()(filename)
+            except ValidationError:
+                pass
+            else:
+                urlparts = urlparse(obj.data_value)
+                filename = os.path.basename(urlparts.path)
```
pld: how about combining this with the below line by rewriting as:

```python
filename = os.path.basename(urlparts.path) or urlparts.netloc
```

ukanga: Forgot I could and did have tests for this, was worried about a boolean.

Issue #948, opened 2017-03-10
Exception Value: 'long' object has no attribute 'split' at /username/forms/id_string/stats


ukanga:
```python
Traceback:

File "/.../.virtualenvs/onadata/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  149.                     response = self.process_exception_by_middleware(e, request)

File "/.../.virtualenvs/onadata/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  147.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "./onadata/apps/viewer/views.py" in charts
  826.     summaries = build_chart_data(xform, lang_index, page)

File "./onadata/libs/utils/chart_tools.py" in build_chart_data
  313.             for field in fields]

File "./onadata/libs/utils/chart_tools.py" in build_chart_data_for_field
  254.                                          choices=choices)

File "./onadata/libs/utils/chart_tools.py" in _use_labels_from_field_name
  163.                     choices, item[truncated_name])

File "./onadata/libs/utils/chart_tools.py" in get_choice_label
  120.                       for name in string.split(" ")]

Exception Type: AttributeError at /[username]/forms/[id_string]/stats
Exception Value: 'long' object has no attribute 'split'
```

PR #1104, opened 2017-08-25
Test get_choice_label to ensure it does not fail for non-string values

moshthepitt:
Fix: #948

### DIFF
--- a/onadata/libs/tests/utils/test_chart_tools.py
+++ b/onadata/libs/tests/utils/test_chart_tools.py
@@ -405,6 +405,14 @@ class TestChartTools(TestBase):
 
         self.assertEqual(get_choice_label(choices, string), [string])
 
+        string = 200
+
+        self.assertEqual(get_choice_label(choices, string), [])
+
+        string = long(200)
+
+        self.assertEqual(get_choice_label(choices, string), [])
+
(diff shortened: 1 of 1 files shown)


moshthepitt:
@ukanga ```isinstance(string, six.string_types)``` below does not allow you to get to string.split(" ") for non-string values, so this seems to have been fixed at some point.

The PR includes tests to confirm this.

```python
def get_choice_label(choices, string):
    """
    `string` is the name value found in the choices sheet.

    Select one names should not contain spaces but some do and this conflicts
    with Select Multiple fields which use spaces to distinguish multiple
    choices.

    A temporal fix to this is to search the choices list for both the
    full-string and the split keys.
    """
    labels = []

    if string and isinstance(string, six.string_types) and choices:
        label = find_choice_label(choices, string)

        if label:
            labels.append(label)
        else:
            # Try to get labels by splitting the string
            labels = [
                find_choice_label(choices, name) for name in string.split(" ")
            ]

            # If any split string does not have a label it is not a multiselect
            # but a missing label, use string
            if None in labels:
                labels = [string]
    elif isinstance(string, list) and string:
        # most likely already translated
        labels = string
    elif not choices:
        labels = [string]

    return labels
```

PR #1122, opened 2017-09-10
XML submission checksum

ukanga:


onadata/libs/utils/logger_tools.py
```diff
@@ -279,10 +285,11 @@ def create_instance(username,
     xml = xml_file.read()
     xform = get_xform_from_submission(xml, username, uuid)
     check_submission_permissions(request, xform)
+    checksum = md5(xml).hexdigest()
```
pld: can we use `sha256` here? I know we're not using it for something were collisions are a big deal right now, but since it's possible to construct collisions intentionally in `md5` I'd prefer we do not use it anywhere

onadata/libs/utils/logger_tools.py
```diff
@@ -297,7 +304,7 @@ def create_instance(username,
         #    has already been submitted for that user.
         return DuplicateInstance()
 
-    # get new and depracated uuid's
+    # get new and deprecated uuid's
```
pld: while we're at it, `UUIDs` instead of `uuid's`, this is not a possessive

onadata/libs/utils/logger_tools.py
ukanga: I think for the functionality we want, `sha256` might be an overkill, the collision will be limited to the single form when it does occur.

pld: Unless there's a big performance penalty, it's worth it to reduce the chance of collisions and because it's concerning that someone can intentionally create a colliding document if we use md5

PR #1124, opened 2017-09-12
Merged dataset fixes

### LINKED ISSUE #1123: Merge Dataset Issues
- [x] merged datasets being created without being linked.
- [ ] deleted xforms affecting a merged datasets
- [x] handle the case when a merged datasets has no form

ukanga:
FIxes #1123


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -346,8 +346,10 @@ def list(self, request, *args, **kwargs):
             xform_id, is_merged_dataset = qs[0] if qs else (lookup, False)
             pks = [xform_id]
             if is_merged_dataset:
-                pks = [__ for __ in MergedXForm.objects.values_list('xforms',
-                                                                    flat=True)]
+                pks = [__ for __ in MergedXForm.objects.filter(
```
pld: why 2 underscores? is the `if` to filter out nils?

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -346,8 +346,10 @@ def list(self, request, *args, **kwargs):
...
+                    pk=xform_id).values_list('xforms', flat=True) if __]
+                if not pks:
```
pld: feel like this is clearer as an `or` clause after the above assignment

onadata/libs/serializers/merged_xform_serializer.py
```diff
@@ -179,4 +178,13 @@ def create(self, validated_data):
             __.instances_with_geopoints for __ in validated_data.get('xforms')
         ])
 
-        return super(MergedXFormSerializer, self).create(validated_data)
+        with transaction.atomic():
+            instance = super(MergedXFormSerializer,
+                             self).create(validated_data)
+
+        if instance.xforms.all().count() != len(xforms):
```
pld: is this to guard against some synchronization problem? Like if a new xform is created by another user on line 184.

Either way, it seems like this is something we, as the server might want to retry a couple times (with increasing pause times) before returning a validation error to the client

onadata/apps/api/viewsets/data_viewset.py
ukanga: A single underscore I might override the `uggettext as _`, the double underscore also keeps pylint silent.
Yes, when a merged dataset has no form linked to it, it returns a nil value here, we need to filter that out.

onadata/libs/serializers/merged_xform_serializer.py
ukanga: The serializer already validates we have the correct number of forms to create a merged dataset, but due to the nature of how Django handles many to many relationships, we have had incidences in preview where a merged dataset was created but the forms were not linked. I have yet to find a way to replicate the issue, a check after the fact I hope will catch the issue. Nothing to do with user synchronization.

onadata/apps/api/viewsets/data_viewset.py
pld: oh right, cool

onadata/libs/serializers/merged_xform_serializer.py
pld: Cool, then I think we should retry this in the code a couple times with back-off before returning an error to the user

ukanga: I think I will take the approach of adding the forms again, it's not clear what/why it failed in the `create()` method. Retrying the whole function might create more objects that are not necessary. The idea of the exception was to force a rollback on the request but I suppose we will have to wait and see if it will reoccur and hopefully have a better understanding of why it is behaving this way.

PR #1134, opened 2017-09-22
Add new 'hash' field to XForm model

moshthepitt:


onadata/apps/logger/models/xform.py
```diff
@@ -815,6 +816,9 @@ def save(self, *args, **kwargs):
         if 'skip_xls_read' in kwargs:
             del kwargs['skip_xls_read']
 
+        # update hash field
+        self.hash = self.get_hash()
```
pld: does this run on every save? can we run this only when we save and the new xml differs from the existing xml?

moshthepitt:
@ukanga I am concerned about cases where xform.save() is called, and xform.title has changed.  

As far as I know, the XML is only updated is you call xform.save(update_fields=['title']).  If this is the case then everything is good, because I only change the hash - as far as the title is concerned - when update_fields['title'] is present.

However, if this is not the case, this needs further work to detect when xform.title changes.  

Thoughts?

pld:
Either way, seems like some more tests are needed for those edge cases

onadata/apps/api/viewsets/xform_list_viewset.py
```diff
@@ -44,8 +44,10 @@ class XFormListViewSet(ETagsMixin, BaseViewset,
     content_negotiation_class = MediaFileContentNegotiation
     filter_backends = (filters.XFormListObjectPermissionFilter,
                        filters.XFormListXFormPKFilter)
-    queryset = XForm.objects.filter(downloadable=True, deleted_at=None,
-                                    is_merged_dataset=False)
+    queryset = XForm.objects.filter(
+        downloadable=True, deleted_at=None,
+        is_merged_dataset=False).only('id_string', 'title', 'version', 'uuid',
+                                      'description', 'user__username')
```
ukanga: You should add the field `hash` here and make the relevant change in the `XFormListSerializer`.

Issue #1143, opened 2017-10-09
RapidPro to Ona form submission


nateMJ:
- [x] Write a RapidPro submission test
- [x] Refactor XML and JSON submissions to use a serializer
- [x] Write a RapidPro serializer class
- [x] Get Rapidpro serializer class on xform submission viewset if the request content type matches `'application/x-www-form-urlencoded'`

PR #1144, opened 2017-10-09
Rapidpro to Ona Submissions

nateMJ:
Closes https://github.com/onaio/onadata/issues/1143

### DIFF
--- a/onadata/apps/api/viewsets/xform_submission_viewset.py
+++ b/onadata/apps/api/viewsets/xform_submission_viewset.py
@@ -25,7 +25,8 @@ from onadata.libs.mixins.authenticate_header_mixin import \
     AuthenticateHeaderMixin
 from onadata.libs.mixins.openrosa_headers_mixin import OpenRosaHeadersMixin
 from onadata.libs.renderers.renderers import TemplateXMLRenderer
-from onadata.libs.serializers.data_serializer import SubmissionSerializer, JSONSubmissionSerializer
+from onadata.libs.serializers.data_serializer import (
+    SubmissionSerializer, JSONSubmissionSerializer,RapidProSubmissionSerializer)
 from onadata.libs.utils.logger_tools import dict2xform, safe_create_instance
 from onadata.apps.api.tools import get_baseviewset_class
 
@@ -60,6 +61,8 @@ class XFormSubmissionViewSet(AuthenticateHeaderMixin,
             self.request.accepted_renderer = JSONRenderer()
             self.request.accepted_media_type = 'application/json'
             return JSONSubmissionSerializer
+        if 'application/x-www-form-urlencoded' in self.request.content_type.lower():
+            return RapidProSubmissionSerializer
 
         return SubmissionSerializer
     
--- a/onadata/apps/api/viewsets/xform_submission_viewset.py
+++ b/onadata/apps/api/viewsets/xform_submission_viewset.py
@@ -18,6 +18,7 @@ from rest_framework.renderers import BrowsableAPIRenderer, JSONRenderer
 
 from onadata.apps.logger.models import Instance
 from onadata.apps.main.models.user_profile import UserProfile
+from onadata.apps.api.permissions import IsAuthenticatedSubmission
 from onadata.libs import filters
 from onadata.libs.authentication import DigestAuthentication
 from onadata.libs.authentication import EnketoTokenAuthentication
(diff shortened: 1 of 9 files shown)


onadata/apps/api/viewsets/xform_submission_viewset.py
```diff
@@ -109,72 +49,24 @@ class XFormSubmissionViewSet(AuthenticateHeaderMixin,
...
-
-        return Response({'error': error_msg},
-                        headers=self.get_openrosa_headers(request),
-                        status=status_code)
+    def get_serializer_class(self):
+        if 'application/json' in self.request.content_type.lower():
+            self.request.accepted_renderer = JSONRenderer()
+            self.request.accepted_media_type = 'application/json'
+            return JSONSubmissionSerializer
+        if 'application/x-www-form-urlencoded' in self.request.content_type.lower():
```
pld: this is repeated on line 60, can refactor and express the structure better w/if else, e.g.

```python
content_type = self.request.content_type.ower()

if 'application/json' in content_type:
    ...
elif 'application/x-www-form-urlencoded' in content_type:
    ...
else
   return Submission Serializer
```

onadata/apps/api/viewsets/xform_submission_viewset.py
```diff
@@ -109,72 +49,24 @@ class XFormSubmissionViewSet(AuthenticateHeaderMixin,
...
+            return RapidProSubmissionSerializer
+
+        return SubmissionSerializer
+    
+    def handle_exception(self, exc):
+        if hasattr(exc, 'response'):
+            return exc.response
+        return super(XFormSubmissionViewSet, self).handle_exception(exc)
```
pld: could be a ternary

```python
return exc.response if hasattr(exc, 'response') else super(XFormSubmissionViewSet, self).handle_exception(exc)
```

onadata/libs/serializers/data_serializer.py
```diff
@@ -113,3 +146,87 @@ def to_representation(self, obj):
...
+            raise serializers.ValidationError({
+                'submission': _(u"Received empty submission. No instance was created")
+            })
+        return super(JSONSubmissionSerializer, self).validate(attrs)
+
+    def create(self, validated_data):
+        request = self.context['request']
+        view = self.context['view']
+
+        username = view.kwargs.get('username')
```
pld: here to 174 is the exact same as 81-85 and 215-219, let's make a helper

onadata/libs/serializers/data_serializer.py
```diff
@@ -113,3 +146,87 @@ def to_representation(self, obj):
...
+        except AttributeError:
+            raise serializers.ValidationError(_(u'Incorrect format, see format details here,'
+                                                u'https://api.ona.io/static/docs/submissions.html.'))
+        xml_string = dict2xform(submission_joined, request.data.get('id'))
+
+        xml_file = StringIO.StringIO(xml_string)
+
+        error, instance = safe_create_instance(
+            username, xml_file, [], None, request)
+        if error and error.status_code >= 400:
```
pld: why does this care about `error.status_code`, but neither the normal submission nor the RapidPro submission do?

code from 183-192 and 224-232 are the same except for the above, looks like it could be refactored into a helper

onadata/libs/utils/dict_tools.py
```diff
@@ -94,3 +96,42 @@ def csv_dict_to_nested_dict(a):
...
+        if isinstance(v, list) and all([isinstance(e, basestring) for e in v]):
+            d[k] = ' '.join(v)
+        elif isinstance(v, dict):
+            d[k] = dict_lists2strings(v)
+
+    return d
+
+
+def dict_paths2dict(d):
+    result = {}
```
pld: This function is missing a docstring. So line 124 is creating a nested dict?  think an example input and output in the docstring would help clarify how it's working

onadata/libs/serializers/data_serializer.py
ukanga: For JSON submission we need to return a JSON response, we don't need to that for RapidPro or XML Submission. `error` is an actual `HTTPResponse` object, and in this other cases we simply append the response to `APIException` class, which we look for in the `XFormSubmissionViewSet.handle_exceptiion(exc)`. As such, the handling is different. This struck the right balance of maintaining legacy code without getting into a lot of re-writing at this time.

pld: Makes sense, there are cases when an error is created with a `< 400` status code?

PR #1146, opened 2017-10-10
Use native Django function refresh_from_db() instead of custom reload()

moshthepitt:


pld:
@moshthepitt can you remove the custom `reload` function too?

moshthepitt:
@pld @ukanga  please review

Issue #1118, opened 2017-09-07
Handle a multi-row single-submission/run in spec result format into Ona datasets


pld:
Modify the submission endpoint to receive a submission or submissions in the format specified here, (bottom) https://github.com/FLOIP/flow-results/blob/master/specification.md#resource-data-found-at-external-path

Todo

* [ ] handle conversion from FLOIP format to Ona/XForm format for `select_many`, `geopoint`, `datetime`, `date`, `time`
* [ ] change mimetype to `application/vnd.org.flowinterop.results+json`

pld:
To approach this, we'll start with handle a single row submission, e.g.:

```
[ "2017-05-23T13:35:37.119-04:00", 20394823948, 923842093, "ae54d3", "female", {"option_order": ["male","female"]} ]
```

Multi-row issue is https://github.com/onaio/onadata/issues/1151

pld:
We are assuming that a single set of rows represents a single flow. We have a question around whether this should support multiple flows at any point.

> Note that we need to handle conversions from FLOIP format to our format for the following column types: `select_many`, `geopoint`, `datetime`, `date`, `time`

> Note that we are dropping all data except the Response data now, later we will need a facility for storing the other data, or the raw submission, so that we can expose it in an export

> Note that we are assuming every submission is only a single flow run and there is no facility for handling a request with multiple runs

I've asked about mimetype and multiple runs on the tracking FLOIP issue, https://github.com/FLOIP/flow-results/issues/9

pld:
Heard back from FLOIP, we can use the draft mimetype `application/vnd.org.flowinterop.results+json` for now, it is listed here https://github.com/FLOIP/flow-results

PR #1160, opened 2017-10-23
Add FLOIP submission endpoint

nateMJ:
Closes https://github.com/onaio/onadata/issues/1118

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
@@ -473,8 +473,7 @@ class TestXFormSubmissionViewSet(TestAbstractViewSet, TransactionTestCase):
         response = self.view(request, username=self.user.username,
                              xform_pk=self.xform.pk)
         self.assertContains(response,
-                            "{u'non_field_errors': [u'All rows must have 6 "
-                            "values, row 0 does not.']}",
+                            "All rows must have 6 values, row 0 does not",
                             status_code=400)
 
     def test_floip_format_submission_not_list(self):
--- a/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
@@ -495,3 +495,23 @@ class TestXFormSubmissionViewSet(TestAbstractViewSet, TransactionTestCase):
                             "{u'non_field_errors': [u'Invalid format. "
                             "Expecting a list.']}",
                             status_code=400)
+    
+    def test_floip_format_multiple_rows_submission(self):
+        """
+        Test FLOIP multiple rows submission
+        """
+        # pylint: disable=C0301
+        data = '[["2017-05-23T13:35:37.119-04:00", 20394823948, 923842093, "ae54d3", "female", {"option_order": ["male", "female"]}], ["2017-05-23T13:35:47.822-04:00", 20394823950, 923842093, "ae54d7", "chocolate", null ]]'  # noqa
+        request = self.factory.post(
+            '/submission', data,
+            content_type='application/vnd.org.flowinterop.results+json')
+        response = self.view(request)
+        self.assertEqual(response.status_code, 401)
+        auth = DigestAuth('bob', 'bobbob')
+        request.META.update(auth(request.META, response))
+        response = self.view(request, username=self.user.username,
+                             xform_pk=self.xform.pk)
+        self.assertContains(response, 'Successful submission', status_code=201)
+        self.assertTrue(response.has_header('Date'))
+        self.assertEqual(response['Location'], 'http://testserver/submission')
+
--- a/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
@@ -529,7 +529,6 @@ class TestXFormSubmissionViewSet(TestAbstractViewSet, TransactionTestCase):
         response = self.view(request, username=self.user.username,
                              xform_pk=self.xform.pk)
         instance_json = Instance.objects.last().json
-        instance_json_values = [item for sublist in instance_json.values()
-                                for item in sublist]
         data_responses = [i[4] for i in json.loads(data)]
-        self.assertTrue(set(data_responses).issubset(instance_json_values))  
+        self.assertTrue(any(i in data_responses
+                            for i in instance_json.values()))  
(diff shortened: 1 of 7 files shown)


onadata/apps/api/viewsets/xform_submission_viewset.py
```diff
@@ -37,18 +42,33 @@ class XFormSubmissionViewSet(AuthenticateHeaderMixin,  # pylint: disable=R0901
...
                         BrowsableAPIRenderer)
     serializer_class = SubmissionSerializer
     template_name = 'submission.xml'
+    parser_classes = (FLOIPParser, JSONParser, FormParser, MultiPartParser)
+
+    def get_serializer(self, *args, **kwargs):
+        """
+        Pass many=True flag if data is a list.
+        """
+        if "data" in kwargs:
```
pld: could change lines 60-64 to be:

```python
data  = kwargs.get("data")

if isinstance(data, list):
    kwargs["many"] = True
```

onadata/libs/serializers/data_serializer.py
```diff
@@ -303,3 +303,36 @@ def create(self, validated_data):
...
+    FLOIP SubmmissionSerializer - Handles a row of FLOIP specification format.
+    """
+    def to_internal_value(self, data):
+        """
+        Overrides validating rows in list data.
+        """
+        if not isinstance(data, list):
+            raise serializers.ValidationError(_(u'Invalid format. '
+                                                u'Expecting a list.'))
+        atleast_five = len(data) >= 5
```
pld: doesn't look like it needs a var, can be made inline

pld: can also remove the not

onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
```diff
@@ -437,3 +436,62 @@ def test_head_submission_anonymous(self):
...
+            '/submission', data,
+            content_type='application/vnd.org.flowinterop.results+json')
+        response = self.view(request)
+        self.assertEqual(response.status_code, 401)
+        auth = DigestAuth('bob', 'bobbob')
+        request.META.update(auth(request.META, response))
+        response = self.view(request, username=self.user.username,
+                             xform_pk=self.xform.pk)
+        self.assertContains(response,
+                            "{u'non_field_errors': [u'Invalid format. "
```
pld: This test isn't passing, not sure if the test is wrong or the `validate` method in the serializer

onadata/apps/api/viewsets/xform_submission_viewset.py
```diff
@@ -60,6 +82,12 @@ def get_serializer_class(self):
         if 'application/x-www-form-urlencoded' in content_type:
             return RapidProSubmissionSerializer
 
+        if 'application/vnd.org.flowinterop.results+json' in content_type:
```
ukanga: Can we have this in a constant variable, it's being used in a couple of places in this file.

```python
FLOIP_RESULTS_CONTENT_TYPE = 'application/vnd.org.flowinterop.results+json'
```
Then
```python
if FLOIP_RESULTS_CONTENT_TYPE in content_type:
    ....
    request.accepted_media_type = FLOIP_RESULTS_CONTENT_TYPE
```

onadata/libs/serializers/data_serializer.py
```diff
@@ -303,3 +303,48 @@ def create(self, validated_data):
...
+    """
+    def validate(self, attrs):
+        data = self.context['request'].data
+        error_msg = None
+
+        if not isinstance(data, list):
+            error_msg = u'Invalid format. Expecting a list.'
+        else:
+            for row_i in xrange(len(data)):
+                if len(data[row_i]) != 6:
```
ukanga: Should `6` be a constant ?

onadata/libs/serializers/data_serializer.py
```diff
@@ -303,3 +303,48 @@ def create(self, validated_data):
...
+                    error_msg = u"All rows must have 6 values, "\
+                                u"row %s does not." % row_i
+                    break
```
ukanga: Could we state what the 6 values should be?

Also, you should probably apply `ugettext()` here for this to be available in translation files.
```python
_("All rows must have 6 values, row %(row)s does not." % {'row': row_i})
```
Applying on the variable should work when it is being translated but the string will not be included in the translation files.

onadata/libs/serializers/data_serializer.py
```diff
@@ -303,3 +303,48 @@ def create(self, validated_data):
...
+        if error_msg:
+            raise serializers.ValidationError(_(error_msg))
+
+        return super(FLOIPSubmissionSerializer, self).validate(attrs)
+
+    def to_internal_value(self, data):
+        """
+        Overrides validating rows in list data.
+        """
+        data = {data[1]: data}
```
ukanga: I feel like we should have a check first that data has at least one row and is probably not `None`.

onadata/libs/serializers/data_serializer.py
```diff
@@ -303,3 +303,48 @@ def create(self, validated_data):
...
+        return data
+
+    def create(self, validated_data):
+        """
+        Returns object instances based on the validated data.
+        """
+        request, username = get_request_and_username(self.context)
+        xform_pk = self.context['view'].kwargs['xform_pk']
+        xform_id = get_object_or_404(XForm, pk=xform_pk).id_string
+        xform_headers = XForm.objects.get(id_string=xform_id).get_keys()
```
ukanga: Why to queries here when we can do with only one?

Do note that `id_string` is not a unique field, it is possible to have the same id_string in different user accounts.

```python
xform = get_object_or_404(XForm, pk=xform_pk)
xform_headers = xform.get_keys()
```

onadata/libs/serializers/data_serializer.py
nateMJ: By this point, we get a row of data already. I can add the check at this point to confirm that.

```
elif len(data) > 0:
    for row_i in xrange(len(data)):
        if len(data[row_i]) != 6:
            error_msg = _(u"All rows must have 6 values, "\
                          u"row %(row)s does not." % {'row': row_i})
```

ukanga: More likely,
```python
elif data:
    for row_i, row in enumerate(data):
        if len(row) != 6:
            error_msg = _(u"All rows must have 6 values, "\
                          u"row %(row)s does not." % {'row': row_i})
```

pld:
Eesh, yeah that's so much better than my code

PR #1170, opened 2017-11-04
Publish FLOIP results data package

### LINKED ISSUE #1116: Convert a FLOIP spec into a restricted XForm/XLSForm
Given an JSON flow result spec as specified here, https://github.com/FLOIP/flow-results#example, convert it to an XLSForm such that the data collected using that spec can now be entered into the XLSForm.

If there's anything in the spec that XLSForm cannot support comment here and lets see what the bet way around this is.

ukanga:
Fix #1116.


ukanga:
Docs will be in #1171

PR #1174, opened 2017-11-07
1173 add bulk delete functionality

### LINKED ISSUE #1173: Add ability to delete multiple instances at once
- [ ] provide xform id on the data endpoint when deleting multiple instances and
- [ ] comma-separated list of ids as a param

ivermac:
closes #1173

### DIFF
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -283,13 +283,25 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                     instance_ids.split(',')
                 )
 
-                instances = Instance.objects.filter(
-                    id__in=instance_ids,
-                    xform=self.object
-                )
+                if not instance_ids:
+                    raise ParseError(_(u"Invalid data ids were provided."))
 
-                for instance in instances:
-                    instance.set_deleted()
+                initial_count = self.object.submission_count()
+                Instance.objects.filter(
+                    id__in=instance_ids,
+                    xform=self.object,
+                    # do not update this timestamp when the record have
+                    # already been deleted.
+                    deleted_at__isnull=True
+                ).update(deleted_at=timezone.now())
+                # updates the num_of_submissions for the form.
+                after_count = self.object.submission_count(force_update=True)
+                number_of_records_deleted = initial_count - after_count
+
+                return Response(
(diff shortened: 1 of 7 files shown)


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -271,10 +271,26 @@ def enketo(self, request, *args, **kwargs):
...
+                    lambda x: x.isdigit(),
+                    instance_ids.split(',')
+                )
+
+                instances = Instance.objects.filter(
+                    id__in=instance_ids,
+                    xform=self.object
+                )
+
+                for instance in instances:
```
ukanga: We may want to actually delete in bulk here rather than for every single submission. `set_deleted()` calls submission count every time you call it.  Perhaps:

```python
from django.utils import timezone
...
initial_count = self.object.submission_count()
instances = Instance.objects.filter(
    id__in=instance_ids,
    xform=self.object,
    deleted_at__isnull=True  # do not update this timestamp when the record have already been deleted. 
).update(deleted_at=timezone.now())
after_count = self.object.submission_count(force_update=True)  # updates the num_of_submissions for the form.

number_of_records_deleted = initial_count - after_count
...
```

This does include the notion of indicating the number of records that have been deleted, we may want to respond with a `status_code = 200` in this case since we are returning this info in the response.

ivermac: Cool, that makes sense. I'll make the change.

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -271,10 +271,38 @@ def enketo(self, request, *args, **kwargs):
...
+                    # do not update this timestamp when the record have
+                    # already been deleted.
+                    deleted_at__isnull=True
+                ).update(deleted_at=timezone.now())
+                # updates the num_of_submissions for the form.
+                after_count = self.object.submission_count(force_update=True)
+                number_of_records_deleted = initial_count - after_count
+
+                return Response(
+                    data="%d records were deleted" % number_of_records_deleted,
```
ukanga: I think this needs to be an object that gets serialized by JSON. Not sure what would be ideal but something like:
```python
{
    "message": "%d records were deleted" % number_of_records_deleted,
"%d records were deleted" % number_of_records_deleted",
    "rows": number_of_records_deleted
}
```

PR #1196, opened 2018-01-02
Prevent duplacate extra columns

### LINKED ISSUE #1155: Additional meta fields in exported data is appearing twice
@jmunene reports that a linked dataset has the error `[_xform_id,_xform_id] match!` in ODK Collect. This turns out to the CSV having more than one column with `_xform_id` on the first row.

ukanga:
closes #1155

### DIFF
--- a/onadata/libs/utils/csv_builder.py
+++ b/onadata/libs/utils/csv_builder.py
@@ -132,7 +132,8 @@ class AbstractDataFrameBuilder(object):
     # fields NOT within the form def that we want to include
     ADDITIONAL_COLUMNS = [
         ID, UUID, SUBMISSION_TIME, TAGS, NOTES, VERSION, DURATION,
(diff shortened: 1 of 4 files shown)


onadata/libs/utils/csv_builder.py
```diff
@@ -561,6 +561,7 @@ def export_to(self, path, dataview=None):
         columns_with_hxl = self.include_hxl and get_columns_with_hxl(
             self.dd.survey_elements)
 
+        print columns
```
pld: ✂️

ukanga: done in c091ca28abfbfa92583e2239b1bfaf9a65d2882f

ukanga:
I have added the test in https://github.com/onaio/onadata/pull/1196/files#diff-5097b703179c589290b37667c8eb07d1R735

PR #1222, opened 2018-01-31
Return exports for all public forms to authenticated users

moshthepitt:
Fix: #1077


moshthepitt:
@JohnMwashuma this is the PR for #1077

Issue #1216, opened 2018-01-23
SAV export will fail when two columns have the same name


moshthepitt:
The SAV export this form https://ona.io/acme/29002/276446#/overview will fail because we have one column named "Groupe" and another named "groupe"

This export will fail with:

```
Traceback (most recent call last):
  File "<console>", line 5, in <module>
  File "/home/onadata/app-versioned/1516012440/onadata/libs/utils/export_builder.py", line 986, in to_zipped_sav
    **sav_options)
  File "/home/onadata/.virtualenvs/onadata/local/lib/python2.7/site-packages/savReaderWriter/savWriter.py", line 220, in __init__
    self.varNamesTypes = self.varNames, self.varTypes
  File "/home/onadata/.virtualenvs/onadata/local/lib/python2.7/site-packages/savReaderWriter/header.py", line 200, in varNamesTypes
    checkErrsWarns(msg, retcode)
  File "/home/onadata/.virtualenvs/onadata/local/lib/python2.7/site-packages/savReaderWriter/error.py", line 120, in checkErrsWarns
    raise SPSSIOError(msg, retcode)
SPSSIOError: Problem setting variable name 'groupe' [SPSS_DUP_VAR]
```

Issue #1226, opened 2018-02-01
SAV exports fail when the data dictionary is empty


moshthepitt:
When trying to create a SAV export for the form with pk = 275210, the export will fail.

For some reason we get a data dictionary that is empty which in turn causes a TypeError exception when building the export.  

The specific TyeError is:

```
argument 3: <type 'exceptions.TypeError'>: wrong type
```

This can be replicated when:
            1. You have duplicate fields in the same group
            2. One of those fields is in a repeat
            3. export_builder.TRUNCATE_GROUP_TITLE = True

moshthepitt:
@faith-mutua

PR #1232, opened 2018-02-05
KJ | JW: Test That SAV Exports With Duplicate Column Names Fail

moshthepitt:
Add a test to ascertain that SAV exports fail when the form has duplicate column names.

Fix: #1216


moshthepitt:
Fix: #1226

This is because we catch exceptions for all failed SAV exports.  The addition in this PR of reporting the exception means we can get the reason why individual SAV exports failed.

PR #1235, opened 2018-02-06
Handle rabbitmq connection errors when applying permissions.

ukanga:


onadata/libs/permissions.py
```diff
@@ -300,47 +347,57 @@ class OwnerRole(Role):
 # Memoize a class to permissions dict.
 for role in ROLES.values():
     role.class_to_permissions = defaultdict(list)
-    [role.class_to_permissions[k].append(p) for p, k in role.permissions]
+    for p, k in role.permissions:
```
pld: This is to not create another object? This may have been written the way it was because we thought list comprehension was faster, did you profile this change? Definitely the comprehension would mean memory allocation (at least briefly until the function got returned, but I suppose even it that case it might not be released but merely marked for GC)

onadata/libs/permissions.py
```diff
@@ -300,47 +347,57 @@ class OwnerRole(Role):
...
 
 
 def get_role(permissions, obj):
-    for role in reversed(ROLES_ORDERED):
-        if role.has_role(permissions, obj):
-            return role.name
+    """
+    Return the user role for the given obj permissions.
+    """
+    for _role in reversed(ROLES_ORDERED):
```
pld: why change? this is not a reserved word, are you concerned it'll conflict/confuse with the class name? The case is different, but I suppose that's fair

onadata/libs/permissions.py
```diff
@@ -300,47 +347,57 @@ class OwnerRole(Role):
 # Memoize a class to permissions dict.
 for role in ROLES.values():
```
pld: this should be `ROLES_ORDERED` instead of `ROLES.values()`

onadata/libs/permissions.py
```diff
@@ -300,47 +347,57 @@ class OwnerRole(Role):
...
 
 def get_role(permissions, obj):
-    for role in reversed(ROLES_ORDERED):
-        if role.has_role(permissions, obj):
-            return role.name
+    """
+    Return the user role for the given obj permissions.
+    """
+    for _role in reversed(ROLES_ORDERED):
+        if _role.has_role(permissions, obj):
```
pld: On another note, can we avoid this `O(len(ROLES_ORDERED))` loop? I imagine this is affecting performance. If the possible values of `permissions` and `obj` are closed (like we can enumerate them) then we should be able to construct a map once when loading this namespace for the first time, and get constant time access.

Hmm, but we can't do this because `obj` can be anything?

pld: Oh wait, but we're looking at the `type` of object to figure this out, is that a closed set?

onadata/libs/permissions.py
```diff
@@ -300,47 +347,57 @@ class OwnerRole(Role):
...
+    return None
 
 
 def get_role_in_org(user, organization):
+    """
+    Return the user role in the organization.
+    """
     perms = get_perms(user, organization)
 
     if 'is_org_owner' in perms:
```
pld: why are these lines even necessary? shouldn't line 385 alone return the correct role?

onadata/libs/permissions.py
```diff
@@ -300,47 +347,57 @@ class OwnerRole(Role):
...
+
+    return get_role(perms, organization) or MemberRole.name
 
 
 def _get_group_users_with_perms(obj, attach_perms=False, user_perms=None):
     """
     Returns a list of users in the groups with permissions on the object obj.
     """
     if isinstance(obj, XForm):
-        Klass = XFormGroupObjectPermission  # pylint: disable=C0103
```
pld: I like the old version better, less repetition, but using a lowercase `klass`

onadata/libs/permissions.py
```diff
@@ -369,13 +426,14 @@ def _get_users_with_perms(obj, attach_perms=False, with_group_users=None):
     Returns a list of users with their permissions on an object obj.
     """
     if isinstance(obj, XForm):
-        Klass = XFormUserObjectPermission  # pylint: disable=C0103
+        user_obj_perms = XFormUserObjectPermission.objects.filter(
```
pld: alright, now this is getting out of hand, let's make a helper to replace lines 392-397 and 428-433:

```python
def _get_user_obj_perms(obj):
    if isinstance(obj, XForm):
        klass = XFormUserObjectPermission
    elif isinstance(obj, Project):
        klass = ProjectUserObjectPermission
    
    return klass and klass.objects.filter(content_object_id=obj.pk)
```

then use as:

```python
def _get_users_with_perms(obj, attach_perms=False, with_group_users=None):
    user_obj_perms = _get_user_obj_perms(obj)

    if not user_obj_perms:
                 return get_users_with_perms(obj, attach_perms=attach_perms,
                                     with_group_users=with_group_users)
    ...
```

onadata/libs/utils/project_utils.py
```diff
@@ -49,29 +58,45 @@ def set_project_perms_to_xform(xform, project):
...
+        # make a report only on the 3rd try.
+        if self.request.retries > 2:
+            msg = '%s: Setting project %d permissions to form %d failed.' % (
+                type(e), project_id, xform_id)
+            report_exception(msg, e, sys.exc_info())
+        # let's retry if the record may still not be available in read replica.
+        self.retry(countdown=60 * self.request.retries)
+    except IntegrityError:
+        # nothing to do, fail silently.
+        pass
```
pld: why is it OK to fail silently here?

onadata/apps/api/tools.py
```diff
@@ -360,7 +354,11 @@ def id_string_exists_in_account():
                 xform.save()
         except IntegrityError:
             raise exceptions.ParseError(_(msg))
-        set_project_perms_to_xform_async.delay(xform.pk, project.pk)
+        try:
+            set_project_perms_to_xform_async.delay(xform.pk, project.pk)
+        except librabbitmq.ConnectionError:
+            # Apply permissions synchrounously
+            set_project_perms_to_xform(xform, project)
```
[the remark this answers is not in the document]
ukanga: No, we cannot connect to the broker, we have no access to a queue in order to queue the background retry. 

We have had the most complaints about permissions on form publishing and sharing, in most cases applying in them in the foreground as a backup will be sufficient. For those with a lot of users who have access to the project, it might take a little bit longer and the client might drop the connection but the permissions will still be applied. Trying to avoid the situation where we have to schedule a cron job to reapply permission every x minutes/hours.

onadata/libs/permissions.py
ukanga: The main reason for this change had to do with Pylint complaining, the comprehension needed to be assigned to something.

```vim
pylint: [expression-not-assigned] Expression "[role.class_to_permissions[k].append(p) for (p, k) in role.permissions]" is assigned to nothing [W0106]
```
I will revisit and see if it can be done differently.

onadata/libs/permissions.py
ukanga: Pylint, it actually has to do with the code with the list comprehension.

onadata/libs/permissions.py
ukanga: I suppose it was not possible to return an `owner` role from the line 385. I will check if that has changed since this section is 4 years old code.

onadata/libs/permissions.py
ukanga: There is a difference, the models are different one affects the group and the other affects the user.

onadata/libs/permissions.py
ukanga: I suppose I hear the screams of pylint loader. I will revisit.

onadata/libs/permissions.py
pld: got it, we can move any work here to this issue too, https://github.com/onaio/onadata/issues/1236, about a general permissions namespace refactor

PR #1237, opened 2018-02-07
Handle connection error on async exports

ukanga:


onadata/apps/viewer/tasks.py
```diff
@@ -82,7 +93,12 @@ def _create_export(xform, export_type, options):
 
     # start async export
     if export_type in export_types:
-        result = export_types[export_type].apply_async((), kwargs=options)
+        try:
+            result = export_types[export_type].apply_async((), kwargs=options)
+        except librabbitmq.ConnectionError as e:
+            export.delete()
```
pld: why do we delete the export? could you add a comment

ukanga: The export did not start, I consider this failure as the export service being down but not a failure related to the data being exported. I think it is important we log the failure but it's not necessary to maintain the record. @moshthepitt seems it's important to have the record, I am going to remove the delete.

PR #1238, opened 2018-02-09
Display Links To Images With Spaces In Filenames Correctly

Wambere:
- Added 'name' to attachment
- Edited function that returns attachment uri to replace
 spaces with underscores

### DIFF
--- a/onadata/libs/tests/utils/test_export_tools.py
+++ b/onadata/libs/tests/utils/test_export_tools.py
@@ -268,7 +268,7 @@ class TestExportTools(TestBase):
             'photo_type_in_repeat_group.xlsx')
         self._publish_xls_file_and_set_xform(path)
 
-        filename = u'bob/attachments/1 2 3.jpg'
+        filename = u'bob/attachments/1_2_3.jpg'
         download_url = u'/api/v1/files/1?filename=%s' % filename
 
         # used a smaller version of row because we only using _attachmets key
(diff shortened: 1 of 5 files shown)


onadata/libs/utils/export_builder.py
```diff
@@ -77,10 +77,8 @@ def get_value_or_attachment_uri(
             if a.get('filename') and a.get('filename').endswith(value)
         ]
         if attachments:
-            value = current_site_url(
-                attachments[0].get('download_url', '')
-            )
-
+            value = current_site_url(attachments[0].get('download_url', ''))
```
ukanga: I don't think this addresses the issue, as we had discussed the check needs to be applied on line 77, replacing it with:
```python
if a.get('name') == value
```
The current issue is that `filename` already has the spaces as underscores. We should be using the newly introduced `name` column that will always have the exact same name as the `value` when it was captured during data collection.

onadata/libs/tests/utils/test_export_tools.py
```diff
@@ -261,6 +262,44 @@ def test_get_value_or_attachment_uri(self):
         self.assertTrue(val_or_url)
         self.assertEqual(value, val_or_url)
 
+    def test_get_attachment_uri_for_filename_with_space(self):
+        path = os.path.join(
+            os.path.dirname(__file__), 'fixtures',
+            'photo_type_in_repeat_group.xlsx')
+        self._publish_xls_file_and_set_xform(path)
+
+        filename = u'bob/attachments/1 2 3.jpg'
```
ukanga: `filename` should actually be `bob/attachments/1_2_3.jpg` to properly address the issue.

PR #1244, opened 2018-02-12
Return Manifest url Only When Object Has Media

Wambere:
- Return manifest url if object has media and none if it does not
- Edited formList.xml to have empty manifest url tags as expected


Wambere:
Fixes issue #1245

PR #1248, opened 2018-02-13
Deal With Duplicate Fields in Different SAV Sections

moshthepitt:
- Get correct value labels for select questions with the same name across different SAV sections.
- Ensure that we do not attempt to handle duplicate fields that appear across SAV sections.  They are not an issue because they represent different SPSS 'sheets'.

Fix: #1226

### DIFF
--- a/onadata/libs/utils/export_builder.py
+++ b/onadata/libs/utils/export_builder.py
@@ -795,7 +795,8 @@ class ExportBuilder(object):
         self._sav_value_labels = {}
 
         for q in choice_questions:
-            if q.get_abbreviated_xpath() not in xpath_var_names:
+            if (xpath_var_names and
+                    q.get_abbreviated_xpath() not in xpath_var_names):
                 continue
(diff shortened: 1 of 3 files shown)


onadata/libs/utils/export_builder.py
```diff
@@ -791,36 +791,38 @@ def _get_sav_value_labels(self, xpath_var_names=None):
...
+            is_numeric = is_all_numeric([c['name'] for c in choices])
+            for choice in choices:
+                name = choice['name'].strip()
+                # should skip select multiple and zero padded numbers e.g
+                # 009 or 09, they should be treated as strings
+                if q.type != u'select all that apply' and is_numeric:
+                    try:
+                        name = float(name) \
+                            if (not name.startswith('0') and
+                                float(name) > int(name)) else int(name)
```
pld: why also the inequality condition?

moshthepitt: @pld I looked at it some more and I think the check 

```
not name.startswith('0')
```

is not necessary because it is already done in *is_all_numeric*

PR #1281, opened 2018-03-01
Add slack support

RemiDesgrange:
This is a WIP. Implementation is incomplete. 

Ressources for slack : https://api.slack.com/docs/message-formatting
https://api.slack.com/docs/messages/builder?msg=%7B"text"%3A"I%20am%20a%20test%20message%20http%3A%2F%2Fslack.com"%2C"username"%3A"Ona%20Bot"%2C"channel"%3A"%23general"%2C"icon_url"%3A"https%3A%2F%2Fona.io%2Fimg%2Ffavicon-32x32.png"%2C"attachments"%3A%5B%7B"text"%3A"And%20here’s%20an%20attachment!"%7D%5D%7D

### DIFF
--- a/onadata/apps/restservice/services/slack.py
+++ b/onadata/apps/restservice/services/slack.py
@@ -1,0 +1,35 @@
+import json
+
+import requests
+from six import string_types
+
+from onadata.apps.main.models import MetaData
+from onadata.apps.restservice.RestServiceInterface import RestServiceInterface
+from onadata.libs.utils.common_tags import TEXTIT
+from onadata.settings.common import METADATA_SEPARATOR
+
+
+class ServiceDefinition(RestServiceInterface):
+    id = SLACK
+    verbose_name = u'Slack Webhook'
+    channel = "#general"
+    username = "Ona Bot"
+    avatar = 'https://ona.io/img/favicon-32x32.png'
+
+    def send(self, url, submission_instance):
+        """
+        Sends the submission to the configured rest service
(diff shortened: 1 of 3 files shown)


ukanga:
Looks like a good start, do check on the flake8 errors we have a test for this hence the fail in this one.

I think you may be looking for a place where to store the slack channel and username for the slack webhook. I will suggest you use the `MetaData` model, https://github.com/onaio/onadata/blob/master/onadata/apps/main/models/meta_data.py. The model links a form to a specific webhook/restservice and has `data_type` which could be `slack` in this case and a `data_value` which could be `slack_webhook_url|channel|username` where `|` is the METADATA_SEPARATOR. The `.send()` method will need to pull this before posting.

onadata/apps/restservice/services/slack.py
```diff
@@ -0,0 +1,35 @@
...
+        
+        ona_data = json.dumps(submission_instance.json)
+        post_data = {
+            "channel": self.channel
+            "username": self.username
+            "icon_url": self.avatar
+            "text": "A new form {form_name} has been submitted by {form_user}".format(form_name=ona_data[], form_user=ona_data[])
+        }
+        headers = {"Content-Type": "application/json"}
+        requests.post(url, headers=headers, data=post_data)
```
ukanga: You may want to wrap this in a `try ... except` block to handle any connection issues among other exceptions that would occur when communicating on the wire.

PR #1282, opened 2018-03-01
Replace Index Tags in CSV

### LINKED ISSUE #1181: Error when pulling data from a XLSForm with repeat groups
**Environment:** Production

image

Wambere:
- Replace the default square brackets with underscores when an index_tag is passed
- Add test to check the replacing

This is to replace the brackets in indices with an underscore in csv which makes it valid xml

Fix #1181


onadata/libs/utils/csv_builder.py
```diff
@@ -145,7 +151,8 @@ def __init__(self, username, id_string, filter_query=None,
                  start=None, end=None, remove_group_name=False, xform=None,
                  include_labels=False, include_labels_only=False,
                  include_images=True, include_hxl=False,
-                 win_excel_utf8=False, total_records=None):
+                 win_excel_utf8=False, total_records=None,
+                 index_tag=DEFAULT_INDEX_TAGS):
```
ukanga: Should this be plural `index_tags`?

onadata/libs/utils/export_builder.py
```diff
@@ -223,6 +223,9 @@ class ExportBuilder(object):
     GROUP_DELIMITER = GROUP_DELIMITER_SLASH
     GROUP_DELIMITERS = [GROUP_DELIMITER_SLASH, GROUP_DELIMITER_DOT]
 
+    # index tags
+    REPEAT_INDEX_TAGS = ['_', '_']
```
ukanga: Can we have this as a tuple instead?
```python
REPEAT_INDEX_TAGS = ('_', '_')
```

onadata/libs/utils/export_tools.py
```diff
@@ -864,4 +872,9 @@ def parse_request_export_options(params):
     if value_select_multiples and value_select_multiples in boolean_list:
         options['value_select_multiples'] = str_to_bool(value_select_multiples)
 
+    if params.get("repeat_index_tags") == DEFAULT_INDEX_TAGS:
```
ukanga: This is super confusing, it does seem like if a user specifies their set of tags then we will use the default ones instead of the user defined.
Was the intention to validate the user-supplied tags instead?

PR #1285, opened 2018-03-05
Add form filter for shared forms

### LINKED ISSUE #1280: formID filter for formList fails when require authentication is true
**Environment**

Stage

**Actual behavior**

formID filter for formList doesn't return any results when the account has require authentication turned on. 

**Expected behavior**

formID filter for formList should return the form in the search results even when require authentication for the account is on.

Wambere:
Fix #1280


onadata/apps/api/viewsets/xform_list_viewset.py
```diff
@@ -94,6 +94,9 @@ def filter_queryset(self, queryset):
             if self.action == 'list' and profile and xform_pk is None:
                 forms_shared_with_user = get_forms_shared_with_user(
                     profile.user)
+                if self.request.query_params.get('formID'):
+                    forms_shared_with_user = forms_shared_with_user.filter(
+                        id_string=self.request.query_params.get('formID'))
```
pld: Since we're extracting this twice, and it's not a trivial extraction, what do you think about pulling it out into a var above then referencing that var in these two locations, e.g.:

```python
id_string = self.request.query_params.get('formID')

if id_string:
    forms_shared_with_user = forms_shared_with_user.filter(id_string=id_string)
```

PR #1286, opened 2018-03-05
Handle session_id for flow results responses.

ukanga:


onadata/libs/renderers/renderers.py
```diff
@@ -22,14 +22,25 @@
 IGNORE_FIELDS = ['meta/instanceID', 'formhub/uuid']
 
 
+def pairing(val1, val2):
+    """
+    Pairing function, encodes two natural numbers into a single natural number.
+
+    Reference: https://en.wikipedia.org/wiki/Pairing_function
+    """
+    return 0.5 * (val1 + val2) * (val1 + val2 + 1) + val2
```
pld: I wonder if it'd be faster to avoid decimals with something like

```python
((val1 + val2) * (val1 + val2 + 1) >> 1) + val2
```

ukanga: For the pairing function, the 0.5 is part of it, though we do convert this into a long when calling the function from below.

pld: Yes, I know this gives the same result but it may perform better

PR #1287, opened 2018-03-06
Improve query performance

### LINKED ISSUE #1188: Gateway timeout error thrown when fetching data from a dataset
**Environment** 

API

**Expected behavior** 

Fetching a small number of records using the GET command with a limit operator should return the number of records specified in the query. 

**Actual behavior**

A Gateway timeout error is thrown when fetching a small number of records from a dataset. 

**Steps to reproduce**

- Log into the onasupport account. 
- Fetch X number of records from this [dataset](https://ona.io/akros_health/44743/244266) using the GET command with a limit operator.

Wambere:
Fix #1188

### DIFF
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -384,7 +384,7 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                     .values_list('xforms', flat=True)
                 pks = [pk for pk in xforms if pk] or [xform_id]
             self.object_list = Instance.objects.filter(
-                xform_id__in=pks, deleted_at=None).only('json').order_by('id')
+                xform_id__in=pks, deleted_at=None).only('json')
             xform = self.get_object()
             self.object_list = \
                 filter_queryset_xform_meta_perms(xform, request.user,
@@ -452,9 +452,8 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                 limit = limit if start is None or start == 0 else start + limit
                 self.object_list = filter_queryset_xform_meta_perms(
                     self.get_object(), self.request.user, self.object_list)
-                self.object_list = \
-                    self.object_list.order_by('pk')[start: limit]
-                self.total_count = self.object_list.count()
+                self.object_list = self.object_list[start: limit]
+                self.total_count = xform.submission_count()
             elif (sort or limit or start or fields) and not is_public_request:
                 try:
                     query = \
@@ -464,10 +463,7 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                     self.object_list = query_data(xform, query=query,
                                                   sort=sort, start_index=start,
                                                   limit=limit, fields=fields)
-                    self.total_count = query_data(
-                        xform, query=query, sort=sort, start_index=start,
-                        limit=limit, fields=fields, count=True
-                    )[0].get('count')
+                    self.total_count = xform.submission_count()
 
                 except NoRecordsPermission:
                     self.object_list = []
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -453,7 +453,7 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                 self.object_list = filter_queryset_xform_meta_perms(
                     self.get_object(), self.request.user, self.object_list)
                 self.object_list = self.object_list[start: limit]
-                self.total_count = xform.submission_count()
+                self.total_count = self.object_list.count()
             elif (sort or limit or start or fields) and not is_public_request:
                 try:
                     query = \
@@ -463,7 +463,10 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                     self.object_list = query_data(xform, query=query,
                                                   sort=sort, start_index=start,
                                                   limit=limit, fields=fields)
-                    self.total_count = xform.submission_count()
+                    self.total_count = query_data(
+                        xform, query=query, sort=sort, start_index=start,
+                        limit=limit, fields=fields, count=True
+                    )[0].get('count')
 
                 except NoRecordsPermission:
                     self.object_list = []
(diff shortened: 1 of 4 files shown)


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -452,8 +452,7 @@ def set_object_list_and_total_count(
                 limit = limit if start is None or start == 0 else start + limit
                 self.object_list = filter_queryset_xform_meta_perms(
                     self.get_object(), self.request.user, self.object_list)
-                self.object_list = \
-                    self.object_list.order_by('pk')[start: limit]
+                self.object_list = self.object_list[start: limit]
```
pld: existing problem, but no space after the `:`

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -503,34 +502,38 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
...
+        Get a StreamingHttpResponse response object
+        """
+        def json_stream(data):
             yield u"["
 
-            for i, d in enumerate(data, start=1):
-                yield json.dumps(d.json if isinstance(d, Instance) else d)
-                yield "" if i == length else ","
+            data_iter = data.__iter__()
+            item = data_iter.next()
```
pld: are we guaranteed that calling `next` here will never throw an exception?

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -503,34 +502,38 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
...
+            while True:
+                try:
+                    next_item = data_iter.next()
+                    yield json.dumps(
+                        item.json if isinstance(item, Instance) else item)
+                    yield ","
+                    item = next_item
+                except StopIteration:
+                    yield json.dumps(
```
pld: this duplicates lines 524-525, but it's not clear if this is worth abstracting ¯\_(ツ)_/¯

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -503,34 +502,38 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
...
-        """Get a StreamingHttpResponse response object
-
-        @param length ensures a valid JSON is generated, avoid a trailing comma
+    def _get_streaming_response(self):
         """
-        def stream_json(data, length):
-            """Generator function to stream JSON data"""
+        Get a StreamingHttpResponse response object
+        """
+        def json_stream(data):
```
pld: why change this name but not `open_data_viewset` in `stream_json`

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -132,42 +132,48 @@ def data(self, request, **kwargs):
...
+        def stream_json(streaming_data):
             """Generator function to stream JSON data"""
             yield u"["
 
-            for i, d in enumerate(streaming_data, start=1):
-                yield json.dumps({
-                    re.sub(r"\W", r"_", a): b for a, b in d.items()
-                })
-                yield "" if i == length else ","
+            data_iter = streaming_data.__iter__()
```
pld: hmm, since lines 159-171 are repeating 519-531 (in one of the namespaces above), can we abstract this?

pld: can also abstract with 696-712 in the namespace below

onadata/apps/api/viewsets/data_viewset.py
pld: in combo w/the comments below, if we abstract this whole block into a function, this part can be a function passed into that function and that'll eliminate the duplication

onadata/apps/api/viewsets/data_viewset.py
pld: Thinking we only abstract the function:

```python
def yield_stream(yield_func):
    while True:
        try:
            next_item = data_iter.next()
                yield yield_func(item)
                yield ","
                item = next_item
        except StopIteration:
            yield yield_func(item)
```

then called as

```python
yield_stream(lambda item: json.dumps(item.json if isinstance(item, Instance) else item)))
```

```python
yield_stream(lambda item: json.dumps({re.sub(r"\W", r"_", a): b for a, b in item.items()})))
```

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -132,42 +132,48 @@ def data(self, request, **kwargs):
                 qs_kwargs.update({'id__gt': gt_id})
 
             instances = Instance.objects.filter(**qs_kwargs).order_by('pk')
-            length = self.total_count = instances.count()
+            self.total_count = instances.count()
```
pld: in this file, `total_count` is only referred to on line 138, can we inline this on line 138 and remove it here so that we only do this work when `count` is `True`?

ukanga: There is a mixin that relies on self.total_count hence it being set here is used elsewhere in the code.

pld: I see, let's remove that mixin, it looks like the only place X-Total is used is in the JS in this repo

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -503,34 +503,23 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
...
-            yield u"["
-
-            for i, d in enumerate(data, start=1):
-                yield json.dumps(d.json if isinstance(d, Instance) else d)
-                yield "" if i == length else ","
-
-            yield u"]"
+        Get a StreamingHttpResponse response object
+        """
+        def get_json_string(item):
```
pld: how about we call this `get_item_or_item_json`

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -681,31 +682,23 @@ def export_async(self, request, *args, **kwargs):
...
-            for xform in queryset:
-                yield json.dumps(XFormBaseSerializer(
-                    instance=xform,
-                    context={'request': self.request}
+        # use queryset_iterator.  Will need to change this to the Django
+        # native .iterator() method when we upgrade to django version 2
+        # because in Django 2 .iterator() has support for chunk size
+        queryset = queryset_iterator(self.object_list, chunksize=2000)
+
+        def get_json_string(item):
```
pld: how about `get_serialized_xform_instance`

onadata/libs/utils/common_tools.py
```diff
@@ -104,3 +104,28 @@ def get_response_content(response):
         contents = response.content
 
     return contents
+
+
+def json_stream(data, json_string):
+    """
+    Generator function to stream JSON data
```
pld: let's add a definition of the arguments, including that `data` must be an iterator

onadata/libs/utils/common_tools.py
```diff
@@ -104,3 +104,28 @@ def get_response_content(response):
...
+        while item:
+            try:
+                next_item = data.next()
+                yield json_string(item)
+                yield ","
+                item = next_item
+            except StopIteration:
+                yield json_string(item)
+                break
+    except AttributeError:
```
pld: let's change this to catch `StopIteration`, so we throw an exception if `data` is not an iterator

onadata/libs/utils/common_tools.py
```diff
@@ -104,3 +104,28 @@ def get_response_content(response):
...
+
+def json_stream(data, json_string):
+    """
+    Generator function to stream JSON data
+    """
+
+    yield u"["
+
+    try:
+        data = data.__iter__()
```
pld: we can move this above the `try` since we do not want to catch anything for it

pld: let's also change the name of this assigned variable from `data` to `iterator`

onadata/libs/utils/common_tools.py
```diff
@@ -104,3 +104,28 @@ def get_response_content(response):
         contents = response.content
 
     return contents
+
+
+def json_stream(data, json_string):
```
pld: for the second argument let's have the name indicate it's a function, e.g. `item_transform_fn`

onadata/libs/utils/common_tools.py
```diff
@@ -104,3 +104,28 @@ def get_response_content(response):
...
+    """
+    Generator function to stream JSON data
+    """
+
```
pld: since we're always going to be dumping the JSON, let's move that out of the passed in function and into here, a way we could do that without repeating the dump call, assuming we rename the argument to `item_transform_fn`, is:

```python
json_dumps_fn = lambda x: json.dumps(item_transform_fn(x))
```

and we replace `json_string` below with `json_dumps_fn`

PR #1290, opened 2018-03-14
1188 remove x total

Wambere:
Part of  #1188

### DIFF
--- a/onadata/apps/api/viewsets/open_data_viewset.py
+++ b/onadata/apps/api/viewsets/open_data_viewset.py
@@ -134,8 +134,7 @@ class OpenDataViewSet(ETagsMixin, CacheControlMixin,
             instances = Instance.objects.filter(**qs_kwargs).order_by('pk')
 
             if count:
-                self.total_count = instances.count()
-                return Response({'count': self.total_count})
+                return Response({'count': instances.count()})
 
             if should_paginate:
                 instances = self.paginate_queryset(instances)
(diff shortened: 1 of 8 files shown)


onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -132,42 +132,34 @@ def data(self, request, **kwargs):
                 qs_kwargs.update({'id__gt': gt_id})
 
             instances = Instance.objects.filter(**qs_kwargs).order_by('pk')
-            length = self.total_count = instances.count()
+            self.total_count = instances.count()
```
pld: this should be in-lined into the `count` conditinoal on line 138 so we don't calculate it by default

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -132,42 +132,34 @@ def data(self, request, **kwargs):
                 qs_kwargs.update({'id__gt': gt_id})
 
             instances = Instance.objects.filter(**qs_kwargs).order_by('pk')
-            length = self.total_count = instances.count()
 
             if count:
+                self.total_count = instances.count()
```
pld: do we even need to set a class variable here? Why not put `instances.count()` inline as the value below?

onadata/libs/utils/common_tools.py
```diff
@@ -104,3 +104,28 @@ def get_response_content(response):
         contents = response.content
 
     return contents
+
+
+def json_stream(data, json_string):
+    """
+    Generator function to stream JSON data
+    """
+
```
pld: remove blank line, other functions do not have a line between docstring and code

Issue #928, opened 2017-02-20
Unhandled IntegrityError at projects endpoint


ukanga:
```python
IntegrityError at /api/v1/projects
duplicate key value violates unique constraint "logger_project_name_74cebfdda007f55b_uniq"
DETAIL:  Key (name, organization_id)=(Monitoring Form, 675) already exists.


Request Method: POST
Request URL: https://api.ona.io/api/v1/projects
Django Version: 1.9.5
Python Executable: /..../.virtualenvs/onadata/bin/uwsgi
Python Version: 2.7.6
Python Path: ['/..../.virtualenvs/onadata/src/savreaderwriter/savReaderWriter', '/..../.virtualenvs/onadata/src/savreaderwriter/savReaderWriter/cWriterow', '.', '', '/..../.virtualenvs/onadata/src/django-nose', '/..../.virtualenvs/onadata/src/django-digest', '/..../.virtualenvs/onadata/src/j2xclient', '/..../.virtualenvs/onadata/src/cors-headers', '/..../.virtualenvs/onadata/src/pricing', '/..../.virtualenvs/onadata/src/savreaderwriter', '/..../.virtualenvs/onadata/src/google-export', '/..../.virtualenvs/onadata/lib/python2.7', '/..../.virtualenvs/onadata/lib/python2.7/plat-x86_64-linux-gnu', '/..../.virtualenvs/onadata/lib/python2.7/lib-tk', '/..../.virtualenvs/onadata/lib/python2.7/lib-old', '/..../.virtualenvs/onadata/lib/python2.7/lib-dynload', '/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/..../.virtualenvs/onadata/local/lib/python2.7/site-packages']
Server time: Mon, 20 Feb 2017 06:35:00 -0500
Installed Applications:
('google_export',
 'pricing',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize',
 'django.contrib.admin',
 'django.contrib.admindocs',
 'django.contrib.gis',
 'registration',
 'django_nose',
 'django_digest',
 'corsheaders',
 'oauth2_provider',
 'rest_framework',
 'rest_framework.authtoken',
 'taggit',
 'readonly',
 'onadata.apps.logger',
[...]

PR #1323, opened 2018-03-26
Test creating project with same name

Wambere:
Tests #928


onadata/apps/api/tests/viewsets/test_project_viewset.py
```diff
@@ -242,6 +242,37 @@ def test_projects_create(self):
             self.assertEqual(self.user, project.created_by)
             self.assertEqual(self.user, project.organization)
 
+    def test_create_duplicate_project(self):
+        """
+        Test creating a project with the same name
+        """
+        self._project_create()
```
moshthepitt: Please add comments to make it more clear that you have created two projects with the same name and that the second project creation fails with a validation error

PR #1330, opened 2018-03-26
Improve setup.py and dependency management

moshthepitt:
- Use `setup.py` to track package dependencies.
- Make `base.in` read from `setup.py`
- Remove `dependency_links` from `setup.py` completely.  This is more accurately handled in our requirement files.
- Exclude tests and docs from `find_packages` in `setup.py`
- Add `setup.py` module docstring
- Change package name to `onadata`
- Add correct version number in `setup.py`

Related reading:
- https://packaging.python.org/tutorials/distributing-packages/#configuring-your-project
- https://github.com/pypa/sampleproject/blob/master/setup.py


moshthepitt:
Tests are failing because we need to add `future` to requirements, which is done in https://github.com/onaio/onadata/pull/1307

moshthepitt:
I'll update this to pull install_requires from requirements/base.in

moshthepitt:
@ukanga after reading around, I am now completely convinced that we should not pull anything from the requirement files into setup.py

I therefore propose to move the dependencies that we have in `base.in` to `setup.py` and then change `base.in` and `dev.in` so that they can still be used to generate our respective `.pip` files.

Please see:
1. https://stackoverflow.com/a/19081268
2. https://caremad.io/posts/2013/07/setup-vs-requirement/
3. https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py#comment63662558_16624700
4. https://github.com/jazzband/pip-tools (`pip-tools` can work through setup.py)

pld:
I'm happy w/this @ukanga

Issue #1332, opened 2018-03-27
Add date and time a message was created


JohnMwashuma:
The messaging endpoint (`curl -X GET https://api.ona.io/api/v1/messaging?target_type=xform&target_id=[target_id]`) is missing two properties, the date the message was created and the time of creation. We also discussed about adding the creators name which is also missing though it is not part of the current deliverables.

Example response from the endpoint :
````
[
    {
        "id": 1,
        "message": "Extra points to those that finish today!",
        "target_id": 241260,
        "target_type": "xform"
    }
]

````

Issue #1331, opened 2018-03-27
Free Form Message Schema Updates


moshthepitt:
Change the free form message schema to include a metadata object with more information on the target as follows:

For user targets:

```json
{
    "payload": {
        "message": "I love oov",
        "context": {
            "type": "user",
            "metadata": {
                "id": 1337,
                "name": "John Doe",    
            }
        },
        "author": {
            "username": "Bob",
            "real_name": "Bob Smith"
        }
    },
    "id": 3,
    "time": "2018-03-27T08:14:51.136675+00:00"
}
```

For project targets:

```json
{
    "payload": {
        "message": "I love oov",
        "context": {
            "type": "project",
            "metadata": {
                "id": 1337,
                "name": "The Project Name", 
            }
        },
        "author": {
            "username": "Bob",
            "real_name": "Bob Smith"
        }
    },
    "id": 3,
    "time": "2018-03-27T08:14:51.136675+00:00"
}
```

For xform targets:

```json
{
    "payload": {
        "message": "I love oov",
        "context": {
            "type": "xform",
            "metadata": {
                "id": 1337,
                "name": "The Form Title",
                "form_id": "The_ID_String"    
            }
        },
        "author": {
            "username": "Bob",
            "real_name": "Bob Smith"
        }
    },
    "id": 3,
    "time": "2018-03-27T08:14:51.136675+00:00"
}
```

PR #1336, opened 2018-03-28
Improve messaging

moshthepitt:
- Enrich MQTT payload for the different target object types
- Add timestamp to MessagingSerializer

Fix: #1331
Fix: #1332


pld:
import these as `text`

PR #1382, opened 2018-04-25
Catch IOError when saving osm data

Wambere:
- Catch exception and raise Http404
- Add test
- Sort imports


onadata/libs/utils/osm.py
```diff
@@ -166,45 +168,51 @@ def save_osm_data(instance_id):
...
-                            osm_data.xml = osm_xml
-                            osm_data.osm_id = osmd['osm_id']
-                            osm_data.osm_type = osmd['osm_type']
-                            osm_data.tags = osmd['tags']
-                            osm_data.geom = GeometryCollection(osmd['geom'])
-                            osm_data.filename = filename
+            try:
+                osm_xml = osm.media_file.read()
+            except IOError as e:
+                raise Http404(str(e))
```
moshthepitt: Why Http404?

ukanga: Http404 will not make sense when this is run by celery as an async task. I will recommend failing silently, and simply skip to the next attachment on the list.

PR #1413, opened 2018-05-18
Add team permissions to org owner role

### LINKED ISSUE #1412: Grant team permissions to all users with org admin permissions
When a user is made admin of an organisation they should automatically get admin permissions to the team/members in the organisation as well, which will allow them to change member roles. 

If a user does not have admin permissions in an organisation, they should not have the option to make changes to the team/members

Wambere:
- Give team permissions to a user when they are given org owner role
- Add test for the above

Fix #1412


Wambere:
Team permissions are assigned in the ShareTeamProjectSerializer

PR #1415, opened 2018-05-21
Check value of remove

Wambere:
The boolean value for remove is in data as a string. This checks if the value is True


pld:
I wonder if there are any users of the API, that assume any value passed for remove means remove, like passing `1` or something. I wonder what Zebra does?

This sort of qualifies as a non-backwards compatible API change

pld:
ok cool​

PR #1423, opened 2018-05-25
New Monthly Submissions Endpoint

Wambere:
- Add new endpoint that returns the total number of submissions to a user's private and public forms
- Add monthly submissions serializer
- Add tests

### DIFF
--- a/onadata/apps/api/viewsets/user_profile_viewset.py
+++ b/onadata/apps/api/viewsets/user_profile_viewset.py
@@ -184,11 +184,9 @@ class UserProfileViewSet(AuthenticateHeaderMixin,  # pylint: disable=R0901
         year = year_param if year_param else str(now.year)
 
         instance_count = Instance.objects.filter(
-            xform__user=profile.user).filter(
-                xform__deleted_at__isnull=True).filter(
-                    date_created__year=year).filter(
-                        date_created__month=month).values(
-                            'xform__shared').annotate(
-                                num_instances=Count('id'))
+            xform__user=profile.user, xform__deleted_at__isnull=True,
+            date_created__year=year, date_created__month=month).values(
+                'xform__shared').annotate(num_instances=Count('id'))
+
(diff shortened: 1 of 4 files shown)


onadata/libs/serializers/monthly_submissions_serializer.py
```diff
@@ -0,0 +1,35 @@
...
+        result_dictionary = {}
+
+        def get_result():
+            for state in states:
+                for i in result:
+                    if i['xform__shared'] == state:
+                        label = 'public' if state else 'private'
+                        result_dictionary[label] = i['num_instances']
+
+        get_result()
```
ukanga: I see no reason for the `get_result()` function, the loop block could easily be not within a function and still work.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -145,3 +150,26 @@ def partial_update(self, request, *args, **kwargs):
...
+        year_param = self.request.query_params.get('year', None)
+
+        # Use query parameter values for month and year
+        # if none, use the current month and year
+        now = datetime.datetime.now()
+        month = month_param if month_param else str(now.month)
+        year = year_param if year_param else str(now.year)
+
+        instance_count = Instance.objects.filter(
+            xform__user=profile.user).filter(
```
pld: is chaining filter instead of a single filter the best practice for this?

onadata/libs/serializers/monthly_submissions_serializer.py
```diff
@@ -0,0 +1,35 @@
...
+class MonthlySubmissionsListSerializer(serializers.ListSerializer):
+
+    def to_representation(self, data):
+        result = super(MonthlySubmissionsListSerializer,
+                       self).to_representation(data)
+        states = [True, False]
+        result_dictionary = {}
+
+        def get_result():
+            for state in states:
```
pld: I think you should be able to do this without `states` and the extra outer loop around `states`

PR #1431, opened 2018-06-08
[DNM]: Update Dockerfile to use python3.6 an ubuntu 18.04

urbanslug:
Using an Ubuntu 18.04 base image
Using python 3.6.x
Removed tmux and other dev tools
Using java8 which is LTS and distributed with ubuntu 18.04

Also related problem that took long to figure out https://bitbucket.org/fomcl/savreaderwriter/issues/73/pip-install-fails-for-non-utf-8-encoding

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 ENV DEBIAN_FRONTEND noninteractive
 ENV PYTHONUNBUFFERED 1
@@ -10,9 +10,8 @@ RUN apt-get update \
     memcached \
     libmemcached-dev \
     build-essential \
-    python-pip \
-    python-virtualenv \
-    python-dev \
+    python3.6-dev \
+    python3.6-venv \
     git \
     libssl-dev \
     libpq-dev \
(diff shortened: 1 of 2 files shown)


urbanslug:
I think it would also be nice if the Dockerfile was more standalone and we had the docker-compose file kick off the setting up of postgres but I could be wrong on this one.

pld:
I agree w/that​

urbanslug:
Ran into an error (others have probably ran into them before) while uploading forms turns out openjdk-8-jre-headless has some issue that doesn't occur with openjdk-8-jre.
`Exception in thread "main" java.awt.AWTError: Assistive Technology not found: org.GNOME.Accessibility.AtkWrapper` so using openjdk-8-jre instead

 - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=798794
 - https://askubuntu.com/questions/695560/assistive-technology-not-found-error-while-building-aprof-plot

docker/docker-entrypoint.sh
```diff
@@ -6,11 +6,8 @@ psql -h db -U postgres -c "CREATE ROLE onadata WITH SUPERUSER LOGIN PASSWORD 'on
 psql -h db -U postgres -c "CREATE DATABASE onadata OWNER onadata;"
 psql -h db -U postgres onadata -c "CREATE EXTENSION postgis; CREATE EXTENSION postgis_topology;"
 
-virtualenv /srv/onadata/.virtualenv
-. /srv/onadata/.virtualenv/bin/activate
-
+source /srv/.virtualenv/bin/activate
 cd /srv/onadata
-pip install -r requirements/base.pip
```
ukanga: I think the pip install still needs to be present in the `docker-entrypoint.sh`, so as no to have to rebuild the image every time you make changes to the requirements file. I use the docker compose file when developing and change is something that is constantly happening.

docker/docker-entrypoint.sh
```diff
@@ -6,11 +6,8 @@ psql -h db -U postgres -c "CREATE ROLE onadata WITH SUPERUSER LOGIN PASSWORD 'on
 psql -h db -U postgres -c "CREATE DATABASE onadata OWNER onadata;"
 psql -h db -U postgres onadata -c "CREATE EXTENSION postgis; CREATE EXTENSION postgis_topology;"
 
-virtualenv /srv/onadata/.virtualenv
-. /srv/onadata/.virtualenv/bin/activate
```
ukanga: I would maintain the virtualenv within the source code directory, I wanted to share the same virtual environment that I could access when not using the docker container and that meant persisting it outside of the container. The alternative I see would be providing defining it as a volume, in either case, this may need to be initialized during startup of the container, hence it's place within the `docker-entrypoint.sh`.

Dockerfile
```diff
@@ -22,48 +21,30 @@ RUN apt-get update \
...
+  && /srv/.virtualenv/bin/pip install -r requirements/base.pip
 
 ENV DJANGO_SETTINGS_MODULE onadata.settings.docker
 
-USER root
-
-# for local development tmux is a nice to have
-RUN apt-get install -y tmux
-
-RUN rm -rf /var/lib/apt/lists/* \
```
ukanga: Reducing the size of the container is usually done by commands like this, probably it should be left behind.

Dockerfile
```diff
@@ -22,48 +21,30 @@ RUN apt-get update \
...
-  && find . -name '*.pyc' -type f -delete
-
-USER onadata
-
-# Run tmux to use bash shell.
-RUN echo "set-option -g default-shell /bin/bash" > ~/.tmux.conf
```
ukanga: Sad to see this go, but maybe I will find a way to run arbitrary commands in the `docker-entry.sh` file to setup a dev environment. As stated, I use this mainly for development purposes, which means, I run tests within the container via `docker exec -it onadata_web_1 bash`. Sometimes this maybe also running management commands to do migrations as well as the Django shell.

ukanga: As such, Tmux is a key part of my workflow.

PR #1439, opened 2018-07-10
Fix Image resize

### LINKED ISSUE #1440: Error resizing attachment image
The download url for a resized image ie `small_download_url`, `medium_download_url` and `large_download_url` gives the error 
```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "'utf8' codec can't decode byte 0xff in position 0: invalid start byte"
}
```

Wambere:
- Open image directly instead of using StringIO
- Add test to resize image

Fix #1440 
Fix https://github.com/onaio/zebra/issues/5405

### DIFF
--- a/onadata/libs/utils/image_tools.py
+++ b/onadata/libs/utils/image_tools.py
@@ -1,12 +1,9 @@
-from builtins import str as text
-from io import StringIO
 from tempfile import NamedTemporaryFile
 
 from django.conf import settings
 from django.core.files.base import ContentFile
 from django.core.files.storage import get_storage_class
 
-import requests
 from PIL import Image
 
 from onadata.libs.utils.viewer_tools import get_path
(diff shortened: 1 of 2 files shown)


onadata/libs/utils/image_tools.py
```diff
@@ -56,22 +53,19 @@ def _save_thumbnails(image, path, size, suffix):
...
-            for key in settings.THUMB_ORDER:
-                _save_thumbnails(
-                    image, filename,
-                    conf[key]['size'],
-                    conf[key]['suffix'])
-        except IOError:
-            raise Exception("The image file couldn't be identified")
+
+    try:
+        path = default_storage.open(filename)
```
ukanga: The variable `path` could be misleading when I `default_storage.open(filename)` returns a file object. Could this be addressed as
```python

with default_storage.open(filename) as image_file:
    image = Image.open(image_file)
    ...
```

PR #1442, opened 2018-07-24
1358 add email verification functionalty

ivermac:


onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +204,64 @@ def monthly_submissions(self, request, *args, **kwargs):
...
+                            return HttpResponseRedirect(redirect_url)
+
+                        response_message = "Email has been verified."
+
+                        return Response(response_message)
+
+        return HttpResponseBadRequest(response_message)
+
+    @action(methods=['POST'], detail=False)
+    def send_verification_email(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
ivermac: Not sure if I agree with the recommendation above. I can't move it outside the class because it's a class function, or a viewset action to be specific. If I'm not wrong, it's mainly complaining because i'm not using `self` object inside the function. I'd like to think the fix here would be retrieve the `request` object from `self` i.e. `self.request.user` but I don't see a plus to that since we already have the request object as part of the args. I'd like to assume that other actions in different classes use the request argument rather than retrieve it from self object, no?

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +204,64 @@ def monthly_submissions(self, request, *args, **kwargs):
 
         serializer = MonthlySubmissionsSerializer(instance_count, many=True)
         return Response(serializer.data[0])
+
+    @action(detail=False)
+    def verify_email(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
ivermac: Same as above

onadata/apps/api/permissions.py
```diff
@@ -157,6 +158,19 @@ def has_permission(self, request, view):
...
+        if view.action in ['send_verification_email', 'verify_email']:
+            enable_email_verification = getattr(
+                settings, 'ENABLE_EMAIL_VERIFICATION', False
+            )
+            if enable_email_verification is None or\
+                    not enable_email_verification:
+                return False
+
+            if view.action == 'send_verification_email':
+                if request.user.username == request.data.get('username'):
```
pld: the equality returns a boolean, why not return that instead of writing your own returns w/bools?

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +204,64 @@ def monthly_submissions(self, request, *args, **kwargs):
...
+        response_message = "Missing or invalid verification key"
+        if verification_key:
+            try:
+                rp = RegistrationProfile.objects.get(
+                    activation_key=verification_key
+                )
+            except RegistrationProfile.DoesNotExist:
+                pass
+            else:
+                response_message = "Email was NOT verified"
```
pld: text should use localization

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +204,64 @@ def monthly_submissions(self, request, *args, **kwargs):
...
+                        'verified': True
+                    }
+                    r = requests.post(webhook, data=payload)
+                    if r.status_code == 200:
+                        rp.activation_key = settings.VERIFIED_KEY_TEXT
+                        rp.save()
+                        if redirect_url:
+                            return HttpResponseRedirect(redirect_url)
+
+                        response_message = "Email has been verified."
```
pld: text should use localization

onadata/apps/api/permissions.py
ivermac: That makes sense. I'll change it.

onadata/libs/templates/registration/verification_email.txt
```diff
@@ -0,0 +1,10 @@
+{% load i18n %}
+{% blocktrans with expiration_days=expiration_days username=username expiration_days_p=expiration_days|pluralize %}
+Hi {{ username }},
+
+To verify your email please click on the link below:
```
ukanga: Can you start a language discussion on this with the product team internally, I think it may need some improvement unless this is exactly the same as the default done by the registration app.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +206,66 @@ def monthly_submissions(self, request, *args, **kwargs):
 
         serializer = MonthlySubmissionsSerializer(instance_count, many=True)
         return Response(serializer.data[0])
+
+    @action(detail=False)
+    def verify_email(self, request, *args, **kwargs):
+        webhook = settings.EMAIL_VERIFICATION_WEBHOOK
```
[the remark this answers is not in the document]
ivermac: I get your concern and I had considered it while working on this feature. I need to add documentation for the this endpoints and the settings variables that need to be set. So if you checkout the if-block that follows [this line](https://github.com/onaio/onadata/pull/1442/files#diff-186b711876dec6613b24f328e6e02ea4R161), you'll notice that if `ENABLE_EMAIL_VERIFICATION` is not set to `true`, the email verification endpoints won't be accessible and a user will get a 403. This means current installations won't be affected. If you set it to `true`, you have to set all the other email verification settings as well for the feature to work. In my opinion, if the app fails with a 500, the dev team would know that they need to set the settings variables that are required. Returning a `None` if the variable does not exist won't be helpful because all those variables are required.

onadata/libs/templates/registration/verification_email.txt
ivermac: I agree with you that this might need improvement. I'll start the discussion. The template was mainly used to check that the email functionality works.

onadata/apps/api/viewsets/user_profile_viewset.py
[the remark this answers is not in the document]
ukanga: I agree you should add API documentation for this endpoint.

However, on the 500 error, I would rather we define default behaviour if those values are not set and provide examples of how that can be defined as you have done on the docker settings file.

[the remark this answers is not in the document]
ivermac: Cool, I'll define a default behaviour.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +206,66 @@ def monthly_submissions(self, request, *args, **kwargs):
...
+        if verification_key:
+            try:
+                rp = RegistrationProfile.objects.get(
+                    activation_key=verification_key
+                )
+            except RegistrationProfile.DoesNotExist:
+                pass
+            else:
+                response_message = _("Email was NOT verified")
+                if webhook:
```
pld: Let's send a 204 if there's a no webhook, so the sender knows we've processed this, so add an `else` here

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +213,84 @@ def monthly_submissions(self, request, *args, **kwargs):
...
+        return HttpResponseBadRequest(response_message)
+
+    @action(methods=['POST'], detail=False)
+    def send_verification_email(self, request, *args, **kwargs):
+        verified_key_text = getattr(settings, "VERIFIED_KEY_TEXT", None)
+        if not verified_key_text:
+            return Response(status=status.HTTP_204_NO_CONTENT)
+
+        username = request.data.get('username')
+        redirect_url = request.data.get('redirect_url')
```
ukanga: The rest framework way I believe is we put this functionality in a serializer, not a big deal. Plenty of examples in this codebase.

onadata/libs/serializers/user_profile_serializer.py
```diff
@@ -218,7 +245,15 @@ def create(self, validated_data):
         new_user.last_name = params.get('last_name')
         new_user.save()
 
-        created_by = self.context['request'].user
+        enable_email_verification = getattr(
+            settings, 'ENABLE_EMAIL_VERIFICATION', False
+        )
+
+        if enable_email_verification:
```
ukanga: `enable_email_verification` is only used once, could simply be:

```python
if getattr(settings, 'ENABLE_EMAIL_VERIFICATION', False):
```

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +213,84 @@ def monthly_submissions(self, request, *args, **kwargs):
...
+        response_message = _("Missing or invalid verification key")
+        if verification_key:
+            try:
+                rp = RegistrationProfile.objects.get(
+                    activation_key=verification_key
+                )
+            except RegistrationProfile.DoesNotExist:
+                pass
+            else:
+                rp.activation_key = verified_key_text
```
ukanga: Why this assignment?

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +213,84 @@ def monthly_submissions(self, request, *args, **kwargs):
...
+                rp.save()
+
+                set_is_email_verified(rp.user.profile, True)
+
+                response_data = {
+                    'ona_verified_username': rp.user.username,
+                    'ona_verified_email_status': True
+                }
```
ukanga: I think  the above should just be:
```python
{'username': rp.user.profile, 'is_verified_email': True}
```

onadata/apps/api/tasks.py
```diff
@@ -75,3 +77,20 @@ def get_async_status(job_uuid):
         return {'JOB_STATUS': result}
 
     return result
+
+
+@task()
+def send_verification_email(data):
```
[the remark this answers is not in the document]
ivermac: I had >= 5 arguments before a refactor but I forgot to refactor this function. I'll refactor it to use 3 args as suggested.

onadata/apps/api/tasks.py
```diff
@@ -75,3 +77,20 @@ def get_async_status(job_uuid):
...
+    subject = data.get('subject')
+    message_txt = data.get('message_txt')
+    email = data.get('email')
```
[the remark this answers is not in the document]
ivermac: I'll add a check for this

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -200,3 +213,84 @@ def monthly_submissions(self, request, *args, **kwargs):
 
         serializer = MonthlySubmissionsSerializer(instance_count, many=True)
         return Response(serializer.data[0])
+
+    @action(detail=False)
+    def verify_email(self, request, *args, **kwargs):
+        verified_key_text = getattr(settings, "VERIFIED_KEY_TEXT", None)
```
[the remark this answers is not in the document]
ivermac: Contains the text that's assigned to `activation_key` field of `RegistrationProfile` instance object once the user has verified the email. When a user activates an email using django's email activation functionality,  the text`'ALREADY_ACTIVATED'` is assigned to `activation_key` field of `RegistrationProfile` instance which we are using when we register a user using the profile endpoint. I just borrowed the idea.

onadata/apps/api/viewsets/user_profile_viewset.py
ivermac: As explained above, it just assigns `ALREADY_ACTIVATED'` to `activation_key` field.

onadata/apps/api/viewsets/user_profile_viewset.py
ivermac: Cool, I'll change it.

docs/profiles.rst
```diff
@@ -269,12 +270,101 @@ Example
...
+
+    curl -X GET https://api.ona.io/api/v1/profiles/verify_email?verification_key=abc&redirect_url=https://red.ir.ect
+
+
+Successful Response
+^^^^^^^^^^^^^^^^^^^
+
+A succesful response or redirect includes 2 values:
+
+- ``ona_verified_username`` - the account username of the verified email
```
ukanga: This keys changed could you have this updated?

onadata/settings/travis_test.py
```diff
@@ -64,3 +64,7 @@
     'django.contrib.messages.middleware.MessageMiddleware',
     'onadata.libs.utils.middleware.HTTPResponseNotAllowedMiddleware',
 )
+
+EMAIL_VERIFICATION_WEBHOOK = "http://localhost:3000/email-verification"
```
ukanga: This is no longer being used, should we remove this option?

PR #1452, opened 2018-08-16
Add SubmissionReview Model

### LINKED ISSUE #1451: Add SubmissionReview Model
As Discussed on #1428. In order to Implement the Submission Review Feature a Model for which SubmissionReview Data will be held is needed.

The Model Should Contain:
- A relation to the Note Object
- Submission Status ( Defaults to Pending) 
- Link to Instance
- Author of the Review
- Date Created
- Date Modified

DavisRayM:
Fix #1451


onadata/apps/logger/models/submission_review.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    SubmissionReview Model Class
+    """
+
+    APPROVED = '1'
+    REJECTED = '2'
+    PENDING = '3'
+
+    STATUS_CHOICES = (
+        (APPROVED, 'Approved'),
```
moshthepitt: make this text translatable please

onadata/apps/logger/models/note.py
```diff
@@ -15,10 +22,17 @@ class Note(models.Model):
...
     class Meta:
+        """
+        Meta Options for Note Model
+        """
         app_label = 'logger'
         permissions = (('view_note', 'View note'), )
 
     def get_data(self):
+        """
+        Custom Method that returns a Dict Containing
```
ukanga: Could be "Returns Note data as a dictionary."

moshthepitt:
@ukanga only admins can delete a review: https://github.com/onaio/zebra/issues/5363#issuecomment-394343783

onadata/apps/logger/models/note.py
DavisRayM: Updated the Docstring on #1455

ukanga:
@moshthepitt  Should deletions be permanent.

moshthepitt:
@ukanga I would vote for Yes, they should be permanent.  I don't see the need to soft delete them.  But on the other hand, we are soft deleting all of the logger models so maybe we should do it to maintain consistency.

pld:
I'd advocate for soft delete because of consistency and if we want future
undo operation, or something analogous

PR #1457, opened 2018-08-21
Add Submission Review ViewSet

DavisRayM:


onadata/apps/logger/models/submission_review.py
```diff
@@ -51,6 +55,19 @@ class Meta:
         """
         app_label = 'logger'
 
+    @classmethod
+    def set_deleted_at(cls, review_id, deleted_at=timezone.now(), user=None):
```
moshthepitt: I think we should pick one of `self` or `cls` for use in the same module, for consistency

DavisRayM: Wouldn't `cls` be more proper to use class methods ?

moshthepitt: yeah, sorry I had missed that

onadata/apps/logger/models/submission_review.py
```diff
@@ -51,6 +55,19 @@ class Meta:
...
+        """
+        Retrieves the submission_review object and calls
+        set_deleted with it
+        """
+        try:
+            submission_review = cls.objects.get(id=review_id)
```
moshthepitt: This `cls.objects` and `cls.DoesNotExist` feels very weird to me.  I'd vote that you just use the model class name as that is how these statements are usually presented

onadata/apps/logger/models/submission_review.py
```diff
@@ -42,6 +43,9 @@ class SubmissionReview(models.Model):
         choices=STATUS_CHOICES,
         default=PENDING,
         db_index=True)
+    deleted_at = models.DateTimeField(null=True, default=None)
+    deleted_by = models.ForeignKey(
```
moshthepitt: lets set on_delete

moshthepitt: I mean the on_delete parameter for the foreign key field should be set.  One reason being that this will be required when we move to Django 2.0

DavisRayM: Added in latest commits

PR #1458, opened 2018-08-23
[WIP] Modify Instance JSON

DavisRayM:
- [ ] Add `_review_status` field to JSON serialization of an Instance
- [ ] Add Documentation on Submission Reviews
- [ ] Update API Documentation

### DIFF
--- a/onadata/apps/logger/models/data_view.py
+++ b/onadata/apps/logger/models/data_view.py
@@ -23,11 +23,13 @@ from onadata.libs.utils.cache_tools import (DATAVIEW_COUNT,
                                             safe_delete)
 from onadata.libs.utils.common_tags import (ATTACHMENTS, EDITED, GEOLOCATION,
                                             ID, LAST_EDITED, MONGO_STRFTIME,
-                                            NOTES, SUBMISSION_TIME)
+                                            NOTES, REVIEW_STATUS,
+                                            SUBMISSION_TIME)
 
 SUPPORTED_FILTERS = ['=', '>', '<', '>=', '<=', '<>', '!=']
 ATTACHMENT_TYPES = ['photo', 'audio', 'video']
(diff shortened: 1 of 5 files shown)


onadata/apps/api/tests/viewsets/test_xform_viewset.py
```diff
@@ -3123,22 +3123,23 @@ def test_csv_export_with_win_excel_utf8(self):
...
-                '\ufeff_media_all_received\n\ufeff#age,,,,,,,,,,,,,\n29,'
+                '\ufeff_total_media,\ufeff_media_count,\ufeff_review_status,'
+                '\ufeff_review_comment,\ufeff_media_all_received\n'
+                '\ufeff#age,,,,,,,,,,,,,,,\n29,'
                 '\ufeffLionel Messi,'
                 '\ufeffuuid:74ee8b73-48aa-4ced-9072-862f93d49c16,%s,'
                 '\ufeff74ee8b73-48aa-4ced-9072-862f93d49c16,'
                 '\ufeff2013-02-18T15:54:01,\ufeff,\ufeff,\ufeff201604121155,'
-                '\ufeff,\ufeffbob,0,0,True\n' % data_id)
+                '\ufeff,\ufeffbob,0,0,n/a,n/a,True\n' % data_id)
```
moshthepitt: why is it n/a?  I thought we had agreed on it being not there?  I think that would mean that here we would just have nothing i.e. something like `,,,` for the two new fields

ukanga:
I apologize I did not have the time to look at this before. I have some thoughts. Would it be a better approach to add `has_a_review` field with the default value as `False` to the `Instance` model? This value will only be changed to true when the a review is created, and the submission can be re-saved at that time to update it with the necessary information?

Issue #1428, opened 2018-06-04
Add Submission Reviews


moshthepitt:
# Submission Reviews

This is a new feature that will allow admins to approve and/or reject submissions made to any of their forms.

This document describes the structure and implementation of a minimal submission review feature and process for Ona.io.  That is, it represents the minimum that has to be done in order to support submission reviews for our most immediate use-case - the ILRI Kaznet project.

The general idea is that this is an extensible starting point for submission reviews in Ona - much more can be done to make it more feature-rich and useful for all our clients, but that will be done at a later time.

## Structure of a submission review

A submission review has two parts:

### Review status

The admin chooses from a number of review statuses:

* Approved - the submission is approved

* Rejected - the submission is rejected (if this status is chosen then the admin must write a review comment that will explain why the submission is rejected)

* Pending Review - the default status. It means that the admin has not yet started a submission review or is still in the review process.

### Review comment

This is an optional comment about the submission.  It is mandatory to provide a comment if a submission status is `Rejected`. You can also make a comment if Approved.

## Who can do a submission review?

Users who are admins (i.e. have the `owner role`) of a form can review submissions made to that form.

## Can a submission be reviewed more than once?

Yes.  A submission review history will be maintained.  The submission review status will be that of the very last submission review made.  e.g. if there are two reviews, one approving and one rejecting, and the one rejecting is the last one made then the status of the submission will be rejected.

## Can a submission reviews be edited?

Yes.  The user who created it can edit it. This may have the effect of changing the submission review status.

## Can submission reviews be deleted?

Yes.  Admins can delete submission reviews. This may have the effect of changing the submission review status.

## Changes to the Onadata API

In order to support the submission review feature, the Onadata API will need to change in the following ways:

### 1. Support for submission reviews

The API will need to accept and store submission reviews for each submission.  The submission review will have the following fields:

* the submission id
* the user who is doing the review
* the date and time of the review
* the review status
* the review comment

> Onadata already has various things that may be used to support the above.  This includes submission Notes and Tags.

### 2. Support for filtering submissions

The API will need to support filtering a form's submissions by the review status.

### 3. Support for filtering exports

The API will need to support generation of exports filtered by the review status.

## Notification to data collection agents

Once a submission has been reviewed, a notification should be sent back to the person who submitted the review.  This will be done through the [messaging application](https://github.com/onaio/onadata/tree/master/onadata/apps/messaging).

---

This is based on the discussion here: https://github.com/onaio/zebra/issues/5363

[...]

PR #1460, opened 2018-08-29
Add submission reviews

moshthepitt:
Fix: #1428

# Submission Reviews API endpoint - /api/v1/submissionreview

### Fields

- id: the id of the submission review (required when doing an update)
- instance: the id of the Instance object being reviewed
- note: the submission review comment
- status: the submission review status.  Should be one of:
  - '1': approved
  - '2': rejected
  - '3': pending

### Sample command

```console
curl -X POST -H "Content-type:application/json" -d '{"status":"1","instance":1337,"note":"This was approved because it is awesome!"}' https://example.com/api/v1/submissionreview/
```

## Updating Submission reviews

```console
curl -X PATCH -H "Content-Type:application/json" -d '{"id":"6","note":"Comment changed!"}' "https://example.com/api/v1/submissionreview/3/"
```

## Retrieving a Submission review

```console
curl -X GET -H "Content-Type:application/json" "https://example.com/api/v1/submissionreview/2/"
```

## Listing submission reviews

```console
curl -X GET -H "Content-Type:application/json" https://example.com/api/v1/submissionreview/
```

### Filtering by instance

```console
curl -X GET -H "Content-Type:application/json" https://example.com/api/v1/submissionreview/?instance=66
```

### Filtering by status

```console
curl -X GET -H "Content-Type:application/json" https://example.com/api/v1/submissionreview/?status=2
```

## Deleting submission reviews

```console
curl -X DELETE -H "Content-Type:application/json" "https://example.com/api/v1/submissionreview/12/"
```

## Changes to Instance JSON

Two new fields are added to the instance JSON returned by the API:

- `_review_status` - the review status of the Instance, taken from the very last submission review attached to that Instance
- `_review_comment` - the review comment of the Instance, taken from the very last submission review attached to that Instance


onadata/apps/api/tools.py
```diff
@@ -451,10 +457,27 @@ def get_user_profile_or_none(username):
     return profile
 
 
+def get_instance_xform_or_none(instance_id):
+    """
+    Returns an The XForm an Instance belongs to
```
pld: remove "an" lowercase "T"

onadata/apps/logger/models/submission_review.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        app_label = 'logger'
+
+    def get_note_text(self):
+        """
+        Custom Property returns associated note text
+        """
+        if self.note is not None:
+            return self.note.note  # pylint: disable=no-member
+        return None
```
pld: do we have a rule/standard on explicitly returning `None`? I noticed it somewhere else above, removing this line would result it the same functionality, with the cost of opacity

onadata/apps/logger/models/submission_review.py
```diff
@@ -0,0 +1,94 @@
...
+        blank=True,
+        default=None,
+        on_delete=models.CASCADE)
+    status = models.CharField(
+        'Status',
+        max_length=1,
+        choices=STATUS_CHOICES,
+        default=PENDING,
+        db_index=True)
+    deleted_at = models.DateTimeField(null=True, default=None)
```
pld: should probably index this, since we'll have plenty queries that ensure this is null

onadata/apps/viewer/models/parsed_instance.py
```diff
@@ -292,6 +293,12 @@ def to_dict_for_mongo(self):
         if isinstance(self.instance.deleted_at, datetime.datetime):
             data[DELETEDAT] = self.instance.deleted_at.strftime(MONGO_STRFTIME)
 
+        if self.get_review_status():
+            data[REVIEW_STATUS] = self.get_review_status()
```
pld: doesn't this (and below 299-300) create a race condition? After the 296 check, someone could remove the review status before it gets to line 297. We can but this into a var before hand, also remove performance deficit of calling this twice

onadata/libs/serializers/submission_review_serializer.py
```diff
@@ -0,0 +1,72 @@
...
+                  'note', 'date_modified')
+
+    def validate(self, attrs):
+        """
+        Custom Validate Method for SubmissionReviewSerializer
+        """
+        status = attrs.get('status')
+        note = attrs.get('note')
+
+        if status == SubmissionReview.REJECTED:
```
pld: why not use `and` in a single `if`?

onadata/libs/utils/common_tags.py
```diff
@@ -119,3 +121,7 @@
     'kml': 'vnd.google-earth.kml+xml',
     OSM: OSM
 }
+
+# Submission Review Tags
+COMMENT_REQUIRED = 'Cannot reject a submission without a comment.'
```
pld: needs to be translated?

ukanga:
Documentation looks good, run through a spell checker.

PR #1462, opened 2018-08-29
Add has has a review field on instance model

lincmba:
Updates upon SubmissionReview creation
Tests


onadata/apps/logger/migrations/0054_instance_has_a_review.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('logger', '0053_submissionreview'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='instance',
+            name='has_a_review',
+            field=models.BooleanField(default=False, verbose_name='Has Had A Review'),
```
moshthepitt: why not `Has Had A Review`

onadata/apps/logger/models/submission_review.py
```diff
@@ -16,6 +16,8 @@ def update_instance_json_on_save(sender, instance, **kwargs):
     Signal handler to update Instance Json with the submission review on save
     """
     submission_instance = instance.instance
+    if not submission_instance.has_a_review:
+        submission_instance.has_a_review = True
     submission_instance.save()
```
moshthepitt: does .save() need to be within the if block?

onadata/apps/logger/models/instance.py
```diff
@@ -335,10 +335,9 @@ def get_full_dict(self, load_existing=True):
             if isinstance(self.deleted_at, datetime):
                 doc[DELETEDAT] = self.deleted_at.strftime(MONGO_STRFTIME)
 
-            if self.get_review_status():
+            # pylint: disable=E1101
```
moshthepitt: Please write the English name of the rule you are disabling (for readability) is there is space on the line

onadata/apps/logger/models/submission_review.py
lincmba: Not really. It still needs to be called even when `has_a_review = True`, to update with the latest review incase of multiple reviews.

ukanga:
Nevermind I think I have traced it. 

:+1:

PR #1464, opened 2018-08-29
Add get_first_record() function.

ukanga:
Loop through instances instead of using queryset.first() method which
seems to be expensive because of the order_by id that it adds to the
query. This was observed on a table with millions of records on a
very fresh database, PostgreSQL selectect an IndexScan in the query
which was taking quite a long time execute and increased the number
of queries waiting to be processed.

### DIFF
--- a/onadata/libs/utils/logger_tools.py
+++ b/onadata/libs/utils/logger_tools.py
@@ -105,6 +105,17 @@ def dict2xform(jsform, form_id, root=None):
         root, form_id, dict2xml(jsform))
 
 
+def get_first_record(queryset):
+    """
+    Returns the first item in a queryset sorted by id.
+    """
+    records = sorted([record for record in queryset], key=lambda k: k.id)
+    if records:
+        return records[0]
+
+    return None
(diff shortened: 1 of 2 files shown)


onadata/libs/utils/logger_tools.py
```diff
@@ -105,6 +105,17 @@ def dict2xform(jsform, form_id, root=None):
         root, form_id, dict2xml(jsform))
 
 
+def get_first_record(queryset):
+    """
+    Returns the first item in a queryset sorted by id.
+    """
+    records = sorted([record for record in queryset], key=lambda k: k.id)
```
moshthepitt: is the queryset not already sorted by id at this point?

ukanga: Nope, that is being added by the `first()` method on the queryset and we do not have a default order on the `Instance` model which this function is mainly meant for.

moshthepitt: cool, I'm good with this

PR #1475, opened 2018-09-18
Use content_type to determin metadata content_object type

ukanga:
This prevents running a query against the xform object which will
include all fields and for forms whose XML is at least 1MB it can dump
more than 5GB of data into memory resulting in processess being killed,
especially where a form has more than 5 metadata objects.

### DIFF
--- a/onadata/libs/serializers/fields/instance_related_field.py
+++ b/onadata/libs/serializers/fields/instance_related_field.py
@@ -1,15 +1,32 @@
-from onadata.apps.logger.models import Instance
+from django.contrib.contenttypes.models import ContentType
+from django.core.cache import cache
+
 from rest_framework import serializers
 from rest_framework.fields import SkipField
+
+from onadata.apps.logger.models import Instance
 
 
 class InstanceRelatedField(serializers.RelatedField):
     """A custom field to represent the content_object generic relationship"""
 
     def get_attribute(self, instance):
-        # xform is not an attribute of the MetaData object
-        if instance and isinstance(instance.content_object, Instance):
-            return instance.content_object
+        content_type_id = cache.get("instance_content_type_id")
+        if not content_type_id:
+            try:
+                content_type_id = ContentType.objects.get(
+                    app_label="logger", model="instance"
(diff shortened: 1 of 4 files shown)


onadata/libs/serializers/fields/project_related_field.py
```diff
@@ -1,15 +1,32 @@
...
+
 
 class ProjectRelatedField(serializers.RelatedField):
     """A custom field to represent the content_object generic relationship"""
 
     def get_attribute(self, instance):
-        # xform is not an attribute of the MetaData object
-        if instance and isinstance(instance.content_object, Project):
-            return instance.content_object
+        content_type_id = cache.get("project_content_type_id")
```
pld: this exactly duplicates 14-29 above and below on lines 15-30, except with a different cache key, can we DRY this up by moving it to a helper function that takes the key as argument?

onadata/libs/serializers/fields/instance_related_field.py
```diff
@@ -1,15 +1,32 @@
...
+        if not content_type_id:
+            try:
+                content_type_id = ContentType.objects.get(
+                    app_label="logger", model="instance"
+                ).id
+            except ContentType.DoesNotExist:
+                pass
+            else:
+                cache.set("instance_content_type_id", content_type_id)
+        if not content_type_id:
```
moshthepitt: I find line 15 - 30 a bit strange, and think it would be really cool to have at least comments (in the case that you do not agree with my comments below) on what this code is doing.  Here's why:

- At line 15 if `content_type_id` is None it will go into the `try` block.
- Do we really need to `pass` on line 21?  To me it feels like the `if` statement on line 24 is unnecessary and the fact that it closely follows the identical `if` statement at line 15 also makes it confusing. 
- Furthermore, the stuff at line 24-26 naturally belongs in the `except ContentType.DoesNotExist` block
- Similarly, to me, line 27-29 naturally also belong to the `else` part of the try block above them.

I think in doing the above we make this code much easier to read and reason about.

Issue #1472, opened 2018-09-06
Update look and feel of classic login page


moshthepitt:
This is our Oauth2 login page and its design and content should be updated - perhaps even to match the Zebra login page.

Aha! Link: https://ona.aha.io/features/PROD-319

PR #1480, opened 2018-09-27
Add Login page tweaks

kahummer:
fix: #1472


onadata/apps/main/views.py
```diff
@@ -1445,3 +1446,25 @@ def get_context_data(self, **kwargs):
...
+            return ['registration/custom/login.html']
+        return super(CustomLoginView, self).get_template_names()
+
+    def get_context_data(self, **kwargs):
+        """
+        get context data i.e password_reset_url for the custom login view
+        """
+        context = super(CustomLoginView, self).get_context_data(**kwargs)
+        if settings.USE_CUSTOM_LOGIN_TEMPLATE:
+            context['PASSWORD_RESET_URL'] = settings.PASSWORD_RESET_URL
```
moshthepitt: This method needs to return `context`

onadata/libs/templates/registration/custom/login.html
```diff
@@ -0,0 +1,76 @@
...
+			</div>
+		</div>
+    </div>
+</div>
+<div class="wrapper cfix">
+<div id="main-menu" class="cfix login-menu"><div class="vw-menu pure-menu-horizontal container cfix">
+        <a id="logo-link" href="/" class="pure-menu-heading left">
+             <span><img id="ona-logo" src="{% static 'custom_login_assets/onadata-logo.png' %}" alt="ONA" /></span>
+        </a>
+        <ul id="user" class="right">
```
moshthepitt: Do we need to have this empty `<ul>`?

onadata/libs/templates/registration/custom/login.html
```diff
@@ -0,0 +1,76 @@
...
+        </p>
+    <input type="hidden" name="__anti-forgery-token" value="" /></form>
+  </div>
+</div> <!-- end content holder -->
+</div>
+<div id="wizard-view" style="display:none"></div>
+</div> <!-- end wrapper-->
+
+<div id="footer" class="footer-simple cfix"><div class="container footer-links">
+        <p><a href="#" id="homepage">Go to Ona homepage</a></p>
```
moshthepitt: This url doesnt lead anywhere.  It should lead to the "Ona homepage" which should come from settings.

moshthepitt: Similarly, the Privacy and Terms links below need to lead somewhere.

ukanga:
Do we really need to the custom view? Could we simply have the templates in place to achieve the same thing?
Could an edit on the existing [login.html](https://github.com/onaio/onadata/blob/master/onadata/libs/templates/registration/login.html) template achieve the same results without the need for a view?

kahummer:
Ona data being open source we want it to be configurable so one can choose which view to load

moshthepitt:
It would be cool to properly format your HTML

kahummer:
fix: #1472 
Update look and feel of classic login page

onadata/apps/main/views.py
kahummer: I have changed that

onadata/apps/main/views.py
```diff
@@ -1445,3 +1446,32 @@ def get_context_data(self, **kwargs):
...
+
+    def get_template_names(self, **kwargs):
+        """
+        get custom login template name if use_custom_login_template is true
+        """
+        context = super(CustomLoginView, self).get_context_data(**kwargs)
+        if settings.USE_CUSTOM_LOGIN_TEMPLATE:
+            context['CUSTOM_LOGIN_TEMPLATE_PAGE'] = \
+                settings.CUSTOM_LOGIN_TEMPLATE_PAGE
+            return context['CUSTOM_LOGIN_TEMPLATE_PAGE']
```
moshthepitt: why not do `return [settings.CUSTOM_LOGIN_TEMPLATE_PAGE]`?  You don't need to mess with context here, in this method.

onadata/apps/main/views_test/test_views_custom.py
```diff
@@ -0,0 +1,30 @@
...
+from django.conf import settings
+from onadata.apps.main.views import CustomLoginView
+
+if settings.USE_CUSTOM_LOGIN_TEMPLATE:
+    class TestCustomLoginView(unittest.TestCase):
+        def setUp(self):
+            super(TestCustomLoginView, self).setUp()
+            self.client = Client()
+            self.factory = RequestFactory()
+            request = (self.factory.get('registration:custom:login'))
```
moshthepitt: what is the purpose of lines 13 - 16?  Seem completely unnecessary to me..

onadata/apps/main/views.py
```diff
@@ -1445,3 +1446,29 @@ def get_context_data(self, **kwargs):
...
+    """
+    Custom LoginView subclass
+    """
+
+    def get_template_names(self, **kwargs):
+        """
+        get custom login template name if use_custom_login_template is true
+        """
+        if settings.USE_CUSTOM_LOGIN_TEMPLATE:
+            return settings.CUSTOM_LOGIN_TEMPLATE_PAGE
```
moshthepitt: This method needs to return a list see: https://docs.djangoproject.com/en/2.1/ref/class-based-views/mixins-simple/#django.views.generic.base.TemplateResponseMixin.get_template_names

onadata/settings/common.py
```diff
@@ -114,7 +114,13 @@
 
 # Login URLs
 LOGIN_URL = '/accounts/login/'
+ONA_LOGIN_LINK = '/login'
```
moshthepitt: I think the default values of these links should point to `https://ona.io`

PR #1488, opened 2018-10-11
Add submission fields to data exports

### LINKED ISSUE #1477: Add submission fields to data exports
Make sure that the submission status and submission comment for instances show up in the exports if the form has reviews.

kahummer:
Upon include_review 
Add submission review fields to data exports

**Milestone** 
- Having this fields should be optional depending on the form
- Addition of this fields should be handled at form level
Fixes #1477


onadata/libs/utils/export_builder.py
```diff
@@ -414,6 +415,10 @@ def get_choice_dict(xpath, label):
         return choices
 
     def set_survey(self, survey, xform=None):
+        if self.INCLUDE_REVIEWS:
```
moshthepitt: I would rather you do this in the `init`:

```python
self.extra_columns = (self.EXTRA_FIELDS + getattr(settings, 'EXTRA_COLUMNS', []))
if self.INCLUDE_REVEWS:
    self.extra_columns = self.extra_columns + [REVIEW_STATUS, REVIEW_COMMENT]
```

onadata/libs/utils/export_tools.py
```diff
@@ -158,6 +158,9 @@ def generate_export(export_type, xform, export_id=None, options=None):
         records = records.iterator()
 
     export_builder = ExportBuilder()
+    reviews = Instance.objects.filter(xform=xform, has_a_review=True).count()
```
ukanga: Should we have the option to export reviews something that is an export option rather than check the database for submissions with reviews?

ukanga:
Looks good but flat CSV exports have not been covered.

onadata/libs/utils/csv_builder.py
```diff
@@ -383,7 +383,14 @@ def __init__(self, username, id_string, filter_query=None,
                  include_images=False, include_hxl=False,
                  win_excel_utf8=False, total_records=None,
                  index_tags=DEFAULT_INDEX_TAGS, value_select_multiples=False,
-                 show_choice_labels=False, language=None):
+                 show_choice_labels=False, language=None,
+                 include_review=False):
+
+        if include_review:
+            if not all(
+                elem in self.ADDITIONAL_COLUMNS for elem in [REVIEW_STATUS,
```
ukanga: Are there situations where the review fields would have been included?

kahummer: I found this happening when I running the tests so I implemented the fix. I have realized the two fields are being added after every test that has the option include_review set to true. Am working on fixing this

docs/proposals/export-submission-review-fields.md
```diff
@@ -0,0 +1,20 @@
+# Export with submission review fields
```
ukanga: Could we also add the documentation in https://github.com/onaio/onadata/blob/master/docs/forms.rst? See https://api.ona.io/static/docs/forms.html#include-labels-in-csv-savzip-xls-and-zipped-csv-exports or https://github.com/onaio/onadata/blob/master/docs/forms.rst#include-labels-in-csv-savzip-xls-and-zipped-csv-exports of how we have done this before.

kahummer: done

onadata/libs/tests/utils/test_export_builder.py
```diff
@@ -52,6 +51,8 @@ class TestExportBuilder(TestBase):
             'name': 'Abe',
             'age': 35,
             'tel/telLg==office': '020123456',
+            '_review_status': 'Active',
```
moshthepitt: I know it is a test, but we dont have `Active` as a possible review status.  Should we avoid any possibly misleading even in tests by using the actual expected value(s)?

onadata/libs/tests/utils/test_export_builder.py
```diff
@@ -228,7 +229,9 @@ class TestExportBuilder(TestBase):
             'osm_building:building:levels': 4,
             'osm_building:ctr:lat': '23.707316084046038',
             'osm_building:ctr:lon': '90.40849938337506',
-            'osm_building:name': 'kol'
+            'osm_building:name': 'kol',
+            '_review_status': 'Active',
```
moshthepitt: I know it is a test, but we dont have `Active` as a possible review status.  Should we avoid any possibly misleading even in tests by using the actual expected value(s)?

PR #1492, opened 2018-10-26
Fix failing xform meta permissions

### LINKED ISSUE #1479: XForm meta permissions failing on org forms
If a form in project, created by an org, is assigned xform meta permissions, the submissions are not filtered based on the assigned permissions. To replicate:
- Create an org
- Create a project inside the org
- Create a form inside the project
- Make a couple of submissions to the form above as the owner
- Share the project with a different user and add the user as a member
- Create meta permissions using [this link](https://api.ona.io/static/docs/metadata.html#create-xform-meta-permissions). Set the `data-value` to `dataentry-minor|editor-minor`.
- Ask the other user to try accessing the data endpoint with the `formid` of the form above
- The other user should be able to view the submissions that were made by the form owner BUT they SHOULDN'T.

lincmba:
Update user roles according to xform meta permissions provided.
Fixes #1479

### DIFF
--- a/onadata/apps/api/tools.py
+++ b/onadata/apps/api/tools.py
@@ -681,26 +681,14 @@ def update_role_by_meta_xform_perms(xform):
     """
     # load meta xform perms
     metadata = MetaData.xform_meta_permission(xform)
-    editor_role_list = [EditorRole, EditorMinorRole]
-    editor_role = {role.name: role for role in editor_role_list}
-
-    dataentry_role_list = [
-        DataEntryMinorRole, DataEntryOnlyRole, DataEntryRole
-    ]
-    dataentry_role = {role.name: role for role in dataentry_role_list}
 
     if metadata:
         meta_perms = metadata.data_value.split('|')
-
-        # update roles
         users = get_xform_users(xform)
 
+        # update user roles with xform meta permissions
         for user in users:
-            role = users.get(user).get('role')
-            if role in editor_role:
-                role = ROLES.get(meta_perms[0])
-                role.add(user, xform)
-
-            if role in dataentry_role:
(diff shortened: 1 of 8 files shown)


moshthepitt:
there's a rogue `__import__('ipdb').set_trace()` somewhere in the code

onadata/libs/models/share_team_project.py
```diff
@@ -18,8 +20,28 @@ def save(self, **kwargs):
             role.add(self.team, self.project)
 
             for xform in self.project.xform_set.all():
+                # check if there is xform meta perms set
+                meta_perms = xform.metadata_set \
+                    .filter(data_type='xform_meta_perms')
```
ukanga: Should `xform_meta_perms` be `XFORM_META_PERMS`?
```python
from onadata.libs.utils.common_tags import XFORM_META_PERMS
...
```

onadata/apps/api/tools.py
```diff
@@ -675,6 +690,24 @@ def get_xform_users(xform):
...
+def get_team_members(org_username):
+    """Return members team if it exists else none.
+
+    :param org_username: organization name
+    :return: team
+    """
+    members = None
+    try:
+        team = Team.objects.get(
+            name="{}#{}".format(org_username, 'members'))
```
ukanga: Should members be a static variable in common_tags?

onadata/apps/api/tools.py
```diff
@@ -652,10 +651,14 @@ def get_xform_users(xform):
     :return:
     """
     data = {}
+    org_members = None
```
ukanga: Could this be an empty list that you simply iterate over it without the if check in line 659?

PR #1506, opened 2018-11-20
Fix error on deleting xform with long id_string or sms_id_string

### LINKED ISSUE #1430: DataError: value too long for type character varying(100)
When deleting a form with a long id_string, the appended text on deletion makes the field longer than the allowed length of the field.  The result is that the form is still not deleted.

```
DataError: value too long for type character varying(100)

(7 additional frame(s) were not displayed)
...
  File "onadata/apps/api/tasks.py", line 61, in delete_xform_async
    xform.soft_delete()
  File "onadata/apps/logger/models/xform.py", line 872, in soft_delete
    self.save(update_fields=update_fields)
  File "onadata/apps/logger/models/xform.py", line 846, in save
    super(XForm, self).save(*args, **kwargs)

DataError: value too long for type character varying(100)

```

lincmba:
Only take the first 100 characters of id_string and sms_id_string as per max_length set for field.

fixes #1430

### DIFF
--- a/onadata/apps/logger/models/xform.py
+++ b/onadata/apps/logger/models/xform.py
@@ -1,14 +1,12 @@
 import json
 import os
 import re
-from builtins import bytes as b, str as text
 from datetime import datetime
-from future.utils import iteritems
-from future.utils import listvalues
 from hashlib import md5
-from past.builtins import cmp
 from xml.dom import Node
 
+import pytz
+from builtins import bytes as b, str as text
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.fields import GenericRelation
@@ -21,9 +19,10 @@ from django.utils import timezone
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext as _
 from django.utils.translation import ugettext_lazy
-
-import pytz
+from future.utils import iteritems
+from future.utils import listvalues
 from guardian.models import GroupObjectPermissionBase, UserObjectPermissionBase
+from past.builtins import cmp
 from pyxform import SurveyElementBuilder, constants
 from pyxform.question import Question
 from pyxform.section import RepeatingSection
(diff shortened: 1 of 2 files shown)


moshthepitt:
@ukanga @lincmba why is the plan to do this by default for all forms?

onadata/apps/logger/models/xform.py
```diff
@@ -890,6 +889,11 @@ def soft_delete(self, user=None):
         self.id_string += deletion_suffix
```
moshthepitt: @ukanga @lincmba basically I am asking is there a reason why we do not want to rewrite lines `889` and `890` as something like:

```python
self.id_string = self.id_string[:self.MAX_ID_LENGTH - len(deletion_suffix)] + deletion_suffix
self.sms_id_string = self.sms_id_string[:self.MAX_ID_LENGTH - len(deletion_suffix)] + deletion_suffix
```

ukanga:
@moshthepitt The reason to rename `id_string` and `sms_id_string` is such that it can allow a user to reuse the same `id_string` on a new form. Appending the date on the id string did allow us to change it and if we had to restore the record, it is easy to identify and remove the additional text.
I think having the limit of 100 characters is more than sane - it probably should be shorter, but that has led me to https://docs.opendatakit.org/openrosa-metadata/#form-identity which does say we should support up to 249 characters or more. The change should affect all forms which exceed this length the same not sure what special case should be excluded. What had you in mind?

onadata/apps/logger/models/xform.py
ukanga: Sounds sane, the question I have, is having the `deletion_suffix` in full important? We do have the `deleted_at` field that has the timestamp of the deletion event.

moshthepitt:
@ukanga Having read your comments and thought about this some more, I think it is fine as it is.

However I do agree with you that we do not really have a reason to append the full deleted_at string to the id_strings.  It seems for some forms we might not be able to recover the original id_strings.

PR #1526, opened 2018-12-21
Exposes submissions url to enketo

WinnyTroy:
This PR contains the functionality to expose Enketo's single submission url.

### DIFF
--- a/onadata/libs/tests/utils/test_viewer_tools.py
+++ b/onadata/libs/tests/utils/test_viewer_tools.py
@@ -6,10 +6,13 @@ import os
 from mock import patch
 
 from django.core.files.base import File
+import requests_mock
+from django.conf import settings
 from django.http import Http404
 from django.test.client import RequestFactory
 from django.test.utils import override_settings
 from django.utils import timezone
+from onadata.libs.exceptions import EnketoError
 
 from onadata.apps.logger.models import XForm, Instance, Attachment
 from onadata.apps.main.tests.test_base import TestBase
@@ -80,7 +83,7 @@ class TestViewerTools(TestBase):
         """
         Test generate_enketo_form_defaults() with multiple params
         """
-        # create xform
+        # create xformtest_create_attachments_zipfile_file_too_big
         self._publish_transportation_form()
         # create kwargs with existing xform variable
         transportation_types = \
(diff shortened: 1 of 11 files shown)


onadata/libs/tests/utils/test_viewer_tools.py
```diff
@@ -164,3 +165,13 @@ def test_get_form_url(self):
...
         url = get_form_url(request, username='bob', xform_pk=1)
         self.assertEqual(url, 'https://ona.io/bob/1')
+
+    @override_settings(TESTING_MODE=False)
+    def test_get_submissions_url(self):
+        """Test get_submissions_url()."""
+        request = RequestFactory().get('/')
+        url = get_submission_url(
+            request, username="milly", id_string="tag_team")
+        __import__('ipdb').set_trace()
```
lincmba: There's a wild ipdb

onadata/libs/utils/viewer_tools.py
```diff
@@ -367,3 +369,31 @@ def get_enketo_preview_url(request, username, id_string, xform_pk=None):
...
+    enketo_url = "https://enketo-stage.ona.io/api/v2/survey/single/once"
+    form_id = id_string
+    server_url = get_form_url(
+        request, username, settings.ENKETO_PROTOCOL, True, xform_pk=xform_pk)
+
+    url = '{}?server_url={}&form_id={}'.format(
+        enketo_url, server_url, form_id)
+
+    response = requests.get(url, auth=(settings.ENKETO_API_TOKEN, ''))
+    __import__('ipdb').set_trace()
```
lincmba: There's a wild ipdb

onadata/libs/utils/viewer_tools.py
```diff
@@ -367,3 +369,31 @@ def get_enketo_preview_url(request, username, id_string, xform_pk=None):
...
+
+    if response.status_code == 200:
+        data = response.json()
+        submission_url = data['single_url']
+        return submission_url
+    elif response.status_code == 400:
+        logger.info(
```
lincmba: Wouldn't it be better to use the error messages as returned by Enketo

onadata/libs/utils/viewer_tools.py
```diff
@@ -367,3 +369,31 @@ def get_enketo_preview_url(request, username, id_string, xform_pk=None):
...
+            "Malformed request. \
+            Kindly check the parameters you have passed in.")
+    elif response.status_code == 401:
+        logger.info(
```
lincmba: Being an api tool, it will be better to return instead of logging the error.

onadata/libs/tests/utils/test_viewer_tools.py
```diff
@@ -164,3 +165,13 @@ def test_get_form_url(self):
...
+        self.assertEqual(
```
lincmba: Wouldn't it be better to test the exact content of the `submission_url` instead of None?

onadata/libs/tests/utils/test_viewer_tools.py
```diff
@@ -164,3 +165,13 @@ def test_get_form_url(self):
...
+            url, None)
```
lincmba: Will be nicer to test the other segments of the function....like the errors section

onadata/libs/tests/utils/test_viewer_tools.py
WinnyTroy: noted. Sounds better. Lemme change that

onadata/libs/utils/viewer_tools.py
```diff
@@ -367,3 +366,27 @@ def get_enketo_preview_url(request, username, id_string, xform_pk=None):
...
+    url = '{}?server_url={}&form_id={}'.format(
+        enketo_url, server_url, form_id)
+
+    response = requests.get(url, auth=(settings.ENKETO_API_TOKEN, ''))
+
+    if response.status_code == 200:
+        data = response.json()
+        submission_url = data['single_url']
+        return submission_url
+    elif response.status_code == 400:
```
ukanga: Since we are raising the same exception with the same error message we probably should only have only the else clause. Check `enketo_url()` function, could be something that might need to be refactored into a function that can be used by both.

requirements/base.pip
```diff
@@ -110,6 +110,7 @@ pyxform==0.12.2
 raven==6.9.0
 recaptcha-client==1.0.6
 requests==2.20.1          # via datapackage, django-oauth-toolkit, httmock, sphinx, tableschema, tabulator
+requests-mock==1.5.2
```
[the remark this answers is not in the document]
WinnyTroy: noted. will include the packages there as well.

Issue #1498, opened 2018-11-07
Don't recreate the same dataview


moshthepitt:
We should prevent the creation of exactly the same dataview for the same form.

When receiving a request to create a new dataview, we should check if the same exact dataview already exists, and if it does, we should not recreate it.

Related: https://github.com/onaio/onadata/issues/1497

ukanga:
Could there be a situation where the data views might be exposed in different projects with different permissions? Wondering whether the limitation should apply to a single project or it is across all systems.

PR #1532, opened 2019-01-04
Fix duplication of the same Dataview  for the same Xform

bmarika:
Fixes https://github.com/onaio/onadata/issues/1498

### DIFF
--- a/onadata/libs/tests/serializers/test_dataview_serializer.py
+++ b/onadata/libs/tests/serializers/test_dataview_serializer.py
@@ -9,16 +9,15 @@ from onadata.apps.logger.models import DataView
 class TestDataViewSerializer(TestAbstractViewSet):
 
     def setUp(self):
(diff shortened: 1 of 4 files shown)


onadata/libs/tests/serializers/test_dataview_serializer.py
```diff
@@ -50,3 +49,41 @@ def get_has_hxl_support_value(data_view_name):
         self.assertFalse(get_has_hxl_support_value('name_only'))
         self.assertTrue(get_has_hxl_support_value('age_only'))
         self.assertTrue(get_has_hxl_support_value('age_and_name'))
+
+    def test_name_and_xforn_are_unique(self):
```
lincmba: typo?

bmarika: fixed

Issue #1494, opened 2018-11-02
Make sure that when a project is deleted all forms are deleted


moshthepitt:
It is possible to have a project that is deleted whose forms are not deleted.

This probably happens when the async tasks to delete forms while deleting a project fail.

We need to make the whole action of deleting projects and its forms atomic.

PR #1534, opened 2019-01-08
WRAP PROJECT AND FORMS SOFT_DELETE IN TRANSACTION.ATOMIC

bmarika:
Ensure that no project forms are left undeleted when the project is deleted. Fixes https://github.com/onaio/onadata/issues/1494

### DIFF
--- a/onadata/apps/logger/models/project.py
+++ b/onadata/apps/logger/models/project.py
@@ -6,7 +6,7 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.postgres.fields import JSONField
 from django.core.exceptions import ValidationError
-from django.db import models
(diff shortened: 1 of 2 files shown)


moshthepitt:
I think it would have been cool to add a test for this @ukanga @bmarika

bmarika:
Had confirmed that tests exist for the Xform soft_delete but will make sure to add more tests especially for projects soft_delete.

moshthepitt:
@bmarika sorry, I meant tests that assure us that the whole transaction(s) is now atomic.  i.e. what if in the future someone mistakenly remove this line of code `@transaction.atomic()`?  A test would catch this regression, no?

bmarika:
Makes sense. Will let you know once I have added the tests.

Issue #1535, opened 2019-01-09
DataError: invalid input syntax for type json


ukanga:
https://sentry.onalabs.org/ona/onadata/issues/568/

```
DataError: invalid input syntax for type json
LINE 1: ...media_count", "checksum", "has_a_review") VALUES ('{"Identif...
                                                             ^
DETAIL:  Token "Infinity" is invalid.
CONTEXT:  JSON data, line 1: ...71432292", "fish_products_prices/bonga": Infinity...

(28 additional frame(s) were not displayed)
...
  File "onadata/libs/utils/logger_tools.py", line 371, in safe_create_instance
    username, xml_file, media_files, uuid=uuid, request=request)
  File "onadata/libs/utils/logger_tools.py", line 339, in create_instance
    date_created_override, checksum)
  File "onadata/libs/utils/logger_tools.py", line 248, in save_submission
    checksum)
  File "onadata/libs/utils/logger_tools.py", line 97, in _get_instance
    checksum=checksum)
  File "onadata/apps/logger/models/instance.py", line 571, in save
    super(Instance, self).save(*args, **kwargs)

DataError: invalid input syntax for type json
LINE 1: ...media_count", "checksum", "has_a_review") VALUES ('{"Identif...
                                                             ^
DETAIL:  Token "Infinity" is invalid.
CONTEXT:  JSON data, line 1: ...71432292", "fish_products_prices/bonga": Infinity...

```

bmarika:
Looks like these two issues are related  https://github.com/onaio/onadata/issues/949

ukanga:
Its a duplicate of #949.

Issue #949, opened 2017-03-13
DataError exception not handled at /submission endpoint


ukanga:
```python
File "/.../.virtualenvs/onadata/local/lib/python2.7/site-packages/django/db/backends/utils.py" in execute
  64.                 return self.cursor.execute(sql, params)

Exception Type: DataError at /submission
Exception Value: invalid input syntax for type json
LINE 1: ...ed", "status", "uuid", "version", "geom") VALUES ('{"grpC6_1...
                                                             ^
DETAIL:  Token "Infinity" is invalid.
CONTEXT:  JSON data, line 1: ...013": "2", "grpC6_4cal/calculate_C6_2E": Infinity...

```

PR #1537, opened 2019-01-11
Handle DataError during XForms submission.

bmarika:
Fixes https://github.com/onaio/onadata/issues/1535 and  https://github.com/onaio/onadata/issues/949


onadata/libs/utils/logger_tools.py
```diff
@@ -403,14 +404,16 @@ def safe_create_instance(username, xml_file, media_files, uuid, request):
         error = OpenRosaResponseBadRequest(
             _(u"Unable to submit because there are multiple forms with"
               u" this formID."))
+    except DataError as e:
+        error = OpenRosaResponseBadRequest(
+            'Data Incompatible with the database provided')
```
ukanga: According to the exceptions, it seems we had a calculate field that returned an Infinity result. I am wondering whether rejecting such submissions is the right way to go about it. Can we be clear in our message of the issue so that an enumerator can know what to check and change on their end? Or should we accept the result as is and replace the Infinity with something that can be stored in a JSON field?
It would be ideal to see if we can actually replicate the issue to determine if the issue is in the data collection tools Enketo or ODK Collect and what could be potentially be solved on that end.

onadata/libs/utils/logger_tools.py
```diff
@@ -353,7 +353,8 @@ def create_instance(username,
             instance.save()
 
         instance = DuplicateInstance()
-
+    except DataError as error:
```
ukanga: This `except` block here is irrelevant, we are not raising a specific exception. I will recommend removing it.

onadata/libs/utils/logger_tools.py
```diff
@@ -403,14 +404,17 @@ def safe_create_instance(username, xml_file, media_files, uuid, request):
         error = OpenRosaResponseBadRequest(
             _(u"Unable to submit because there are multiple forms with"
               u" this formID."))
+    except DataError as e:
+        error = OpenRosaResponseBadRequest(
```
ukanga: Did this have to be 3 lines could, `OpenRosaResponseBadRequest(str(e))` not fit within the 79/80 line limit?

onadata/libs/utils/logger_tools.py
bmarika: Done

onadata/libs/utils/logger_tools.py
bmarika: Fixed

Issue #1525, opened 2018-12-19
ProgrammingError: syntax error at or near "căm"


ukanga:
https://sentry.onalabs.org/ona/onadata/issues/300/

```
ProgrammingError: syntax error at or near "căm"
LINE 1: ... AND deleted_at IS NULL  AND json->>'xa' = 'Chư R'căm' GROUP...
                                                             ^

(12 additional frame(s) were not displayed)
...
  File "onadata/apps/api/viewsets/dataview_viewset.py", line 210, in charts
    data_view=dataview
  File "onadata/libs/utils/chart_tools.py", line 472, in get_chart_data_for_field
    data_view=data_view)
  File "onadata/libs/utils/chart_tools.py", line 281, in build_chart_data_for_field
    field_name, data_view)
  File "onadata/libs/data/query.py", line 211, in get_form_submissions_grouped_by_field
    return _execute_query(_postgres_count_group(field, name, xform, data_view))
  File "onadata/libs/data/query.py", line 20, in _execute_query
    cursor.execute(query)

ProgrammingError: syntax error at or near "căm"
LINE 1: ... AND deleted_at IS NULL  AND json->>'xa' = 'Chư R'căm' GROUP...
                                                             ^

```

PR #1542, opened 2019-01-21
Escape apostrophes in SQL queries

bmarika:
Fixes https://github.com/onaio/onadata/issues/1525


ukanga:
https://github.com/onaio/onadata/issues/1546 should address this when we have more time to refactor the query.py module.

PR #1543, opened 2019-01-21
Optional manifest url

### LINKED ISSUE #1519: Error in ODK Briefcase when pulling data in versions 11 to 13
The following error message is displayed in ODK Briefcase whenever versions 11 to 13 are used to pull data:

`Fetching form manifest
Error fetching form definition: Fetch of manifest failed. Detailed reason: Invalid url. Failed with error: no protocol: . Did you forget to prefix the address with 'http://' or 'https://' ?`

lincmba:
Optional fields in formList such as manifestUrl should not be included in formList incase none exists.
fixes #1519

### DIFF
--- a/onadata/libs/renderers/renderers.py
+++ b/onadata/libs/renderers/renderers.py
@@ -7,13 +7,11 @@ import json
 import math
 from io import BytesIO
 
-from future.utils import iteritems
-
+import pytz
 from django.utils.dateparse import parse_datetime
 from django.utils.encoding import smart_text
 from django.utils.xmlutils import SimplerXMLGenerator
-
-import pytz
+from future.utils import iteritems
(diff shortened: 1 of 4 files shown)


onadata/libs/renderers/renderers.py
```diff
@@ -244,9 +250,12 @@ def _to_xml(self, xml, data):
 
         elif isinstance(data, dict):
             for (key, value) in iteritems(data):
-                xml.startElement(key, {})
-                self._to_xml(xml, value)
-                xml.endElement(key)
+                if key not in FORMLIST_MANDATORY_FIELDS and value is None:
+                    pass
```
ukanga: Could we use the `continue` keyword and avoid the else block? I think it seems cleaner that way.

PR #1564, opened 2019-02-13
Fix Upload Dataset

### LINKED ISSUE #1561: Data Upload Failing
When parameter `overwrite =true` is passed to the csv import endpoint, only one record is updated 

**Expected Behaviour**
Given this data 
screen shot 2019-02-11 at 16 51 15
All three records should be uploaded.

**Actual Behaviour**
Only one record is uploaded.

lincmba:
only prepend 'uuid:' if there's a `_uuid`. Avoid `uuid:None` being considered as uuid. 
Fixes #1561


ukanga:
How does this code change address issue #1561? I had thought that adding the test would have made it clear to me but it does not.

ukanga:
Nevermind, I think I get it now.

Issue #1491, opened 2018-10-25
Add command to transfer projects in one user account to another user account


ukanga:
Command: **transfer_user_projects**.

Transfer projects to a different user/organisation account.
- [x] Reassign created_by and organisation to the new owner/organisation.
- [x] Reassign all XForms under the user or project to the new
      user/organisation. Updating the fields `user` and `created_by`.
 - [x] Reassign all MergedXForms under the user or project to the new
      user/organisation. 
- [x] Reassign all Dataviews under the user or project to the new
      user/organisation. 
- [x] Apply project and form permissions to the new user.
- [x] Remove project and form permissions from the old user?
- [ ] Update Enketo references accordingly in issue #845 ?

ukanga:
@msschroeder @ivermac 

- Anything I may have missed to add in this description relevant to transferring of entire projects from one user account to another?
- Does transferring the project also mean existing users with access to the project will no longer have access to the transferred project?
- Should this functionality also be exposed via API - for a user to transfer all their projects to another user?

ivermac:
**_Anything I may have missed to add in this description relevant to transferring of entire projects from one user account to another?_**
> Looks good to me

**_Does transferring the project also mean existing users with access to the project will no longer have access to the transferred project?_**
> It makes sense to me to revoke any permission or access the existing users had to the project that has been transferred. I'd like to think it's up to the new user/org to share/re-share the project.

**_Should this functionality also be exposed via API - for a user to transfer all their projects to another user?_**
> If we ever need a UI for this functionality then i'd like to think exposing the functionality via the API would make sense

Issue #845, opened 2016-11-21
Form ownership transfer does not result to an update in enketo


ivermac:
@cagulas and myself just discovered a bug - I don't think we ever thought about it initially when forms transfer was introduced. When a user transfers form ownership to another user/org, enketo is not updated to reflect the new owner or the new formList. This is a problem especially if the original url is already being used. I guess this is also a chance for someone else, other than me, to checkout and explore if this possible via the enekto API and how onadata will be able to do this

DavisRayM:
This doesn't seem like an issue any more. Closing it for now, please reopen if it's need once again.

PR #1574, opened 2019-02-25
Command to move projects amongst users

bmarika:
Fixes https://github.com/onaio/onadata/issues/1491

Carried Over from #1491 
Transfer projects to a different user/organisation account.
- [x] Reassign created_by and organisation to the new owner/organisation.
- [x] Reassign all XForms under the user or project to the new
      user/organisation. Updating the fields `user` and `created_by`.
 - [x] Reassign all MergedXForms under the user or project to the new
      user/organisation. 
- [x] Reassign all Dataviews under the user or project to the new
      user/organisation. 
- [x] Apply project and form permissions to the new user.
- [x] Remove project and form permissions from the old user?


onadata/apps/logger/management/commands/transfer_project.py
```diff
@@ -0,0 +1,97 @@
...
+        xforms = XForm.objects.filter(project=project)
+        for form in xforms:
+            form.user = user
+            form.created_by = user
+            form.save()
+            self.update_data_viewws(form)
+            set_project_perms_to_xform(form, project)
+
+    @staticmethod
+    def update_data_viewws(form):
```
ukanga: Is the spelling a typo or intentional? `update_data_viewws`? I would suggest we stick to `update_data_views`.

onadata/apps/logger/management/commands/transfer_project.py
```diff
@@ -0,0 +1,97 @@
...
+        except user_model.DoesNotExist:
+            self.errors.append("User {0} does not exist \n".format(username))
+        return user
+
+    def update_xform_with_new_user(self, project, user):
+        """
+        Update XForm user update the DataViews and also set the permissions
+        for the xform and the project.
+        """
+        xforms = XForm.objects.filter(project=project)
```
ukanga: We may need to address the issue for inactive forms i.e `downloadable=False`, there is usually an exception that is raised for inactive forms. I am not sure if it applies to the other models. Also, for the situation where an object has been soft deleted.

onadata/apps/logger/management/commands/transfer_project.py
bmarika: It is a typo. Will fix it.

onadata/apps/logger/management/commands/transfer_project.py
bmarika: I have made changes to address both scenarios.

onadata/apps/logger/management/commands/transfer_project.py
```diff
@@ -0,0 +1,125 @@
...
+    def update_merged_xform(project, user):
+        merged_xforms = MergedXForm.objects.filter(
+            project=project, deleted_at__isnull=True)
+        for form in merged_xforms:
+            form.user = user
+            form.created_by = user
+            form.save()
+            set_project_perms_to_xform(form, project)
+
+    def update_enketo_urls(self, form):
```
ukanga: I think this does correctly create Enketo metadata records for a new Enketo URI from the move but maintains the old URI. I don't think this sufficiently addresses the issue in #845. A fix for that could be having an endpoint that does not include user/org identifying information in the URI.

onadata/apps/logger/management/commands/transfer_project.py
```diff
@@ -0,0 +1,93 @@
...
+from django.db import transaction
+
+from onadata.apps.logger.models import Project, XForm, DataView, MergedXForm
+from onadata.apps.logger.models.project import set_object_permissions \
+    as set_project_permissions
+from onadata.libs.utils.project_utils import set_project_perms_to_xform
+
+
+class Command(BaseCommand):  # pylint: disable=C0111
+    help = 'A command to reassign a project form one user to the other.'
```
ukanga: This command transfers all projects from one user to another user, the description seems to indicate we are only transferring one project.

Could we have two options:
1. `--project` -  the project id of the project to transfer
2. `--all` - transfer all projects

onadata/apps/logger/management/commands/transfer_project.py
```diff
@@ -0,0 +1,93 @@
...
+    errors = []
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--current_owner',
+            help='Username of the current owner of of the projects',
+        )
+        parser.add_argument(
+            '--new_owner',
+            help='TUsername of new owner of the projects',
```
ukanga: I think this could be a typo `TUsername`.

PR #1581, opened 2019-03-01
Convert excel date format to csv format

### LINKED ISSUE #1577: Date data imported in a format that's not recognized by ODK Briefcase
Currently while importing data,  date data is imported in this format dd/mm/yyyy while ODK Briefcase recognizes this format dd-mm-yyyy. As a result, an error message is thrown while exporting data as CSV using the ODK Briefcase "Unexpected error. See logs".

- Upload and submit data to this form: https://docs.google.com/spreadsheets/d/1tE-KO5rK-n8Ifag77CfFiNtzyleim6u-H_c6-MXxm1g/edit?usp=sharing
- Export the data as a CSV or Excel file.
- Open the exported file then, save the file.
- Import the data back to ona.io using the [Import CSV feature](https://help.ona.io/knowledge-base/how-do-i-import-data-to-a-form-using-import-csv/).
- Pull the form submissions using [ODK Briefcase](https://help.ona.io/knowledge-base/how-do-i-use-odk-briefcase/#how-to-pull-forms-with-submissions-from-ona) then, [export ](https://docs.opendatakit.org/briefcase-using/#export-forms-to-csv)the pulled submissions as CSV.
- "Unexpected error. See logs" error message will be returned.

lincmba:
Converts date from `d/m/yyyy` to `yyyy-mm-dd`

CSV date is in the format `yyyy-mm-dd`. When this file is opened by excel, the date format changes  to `d/m/yyyy`. This format is maintained after saving. 
fixes #1577

### DIFF
--- a/onadata/libs/utils/common_tags.py
+++ b/onadata/libs/utils/common_tags.py
@@ -133,3 +133,4 @@ SUBMISSION_REVIEW_INSTANCE_FIELD = '_review_status'
 EXCEL_TRUE = 1
 EXCEL_FALSE = 0
 MEMBERS = 'members'
+XLS_DATE_FIELDS = ['date', 'today']
--- a/onadata/libs/utils/csv_import.py
+++ b/onadata/libs/utils/csv_import.py
@@ -254,10 +254,11 @@ def submit_csv(username, xform, csv_file, overwrite=False):
             for key in xl_date_columns:
                 try:
                     date = datetime.strptime(row.get(key, ''), '%m/%d/%Y')
-                    str_date = datetime.strftime(date, '%Y-%m-%d')
-                    row.update({key: str_date})
                 except ValueError:
                     pass
+                else:
+                    str_date = datetime.strftime(date, '%Y-%m-%d')
+                    row.update({key: str_date})
 
             # remove the additional columns
             for index in addition_col:
--- a/onadata/libs/utils/csv_import.py
+++ b/onadata/libs/utils/csv_import.py
@@ -241,23 +241,16 @@ def submit_csv(username, xform, csv_file, overwrite=False):
     ona_uuid = {'formhub': {'uuid': xform.uuid}}
     error = None
     additions = duplicates = inserts = 0
-    xl_date_columns = []
+
+    x_json = json.loads(xform.json)
+    xl_date_columns = [
+        dt.get('name') for dt in x_json.get('children')
+        if dt.get('type') == 'date']
 
     try:
-        for indx, row in enumerate(csv_reader):
-            #  check for any dates in first row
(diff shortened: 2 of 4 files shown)


onadata/libs/utils/csv_import.py
```diff
@@ -241,8 +241,25 @@ def submit_csv(username, xform, csv_file, overwrite=False):
...
     try:
-        for row in csv_reader:
+        for indx, row in enumerate(csv_reader):
+            #  check for any dates in first row
+            if indx == 0:
+                for key in list(row):
+                    if not key.startswith('_'):
+                        try:
+                            date = datetime.strptime(
+                                row.get(key, ''), '%d/%m/%Y')
```
ukanga: Is it possible to gate which fields are date fields from the `XForm` object?
How do we handle `datetime` fields?

lincmba: The `datetime` fields remain unchanged with `-` as it should in excel version 16.0 that I have tested with. So far, only `date` fields are affected. 
And it's possible to get the date fields from the `XForm`, and I think it could be more efficient so.

onadata/libs/utils/common_tags.py
```diff
@@ -133,3 +133,4 @@
 EXCEL_TRUE = 1
 EXCEL_FALSE = 0
 MEMBERS = 'members'
+XLS_DATE_FIELDS = ['date', 'today']
```
ukanga: I think `datetime` is not being considered, does it not affect that type of field? The other fields to look at are `start` and `end` fields.

ukanga: Seems I missed https://github.com/onaio/onadata/pull/1581#discussion_r262221961.

onadata/libs/utils/csv_import.py
```diff
@@ -241,8 +242,23 @@ def submit_csv(username, xform, csv_file, overwrite=False):
...
     try:
         for row in csv_reader:
+            # convert some excel dates, replace / with -
+            for key in xl_date_columns:
+                try:
+                    date = datetime.strptime(row.get(key, ''), '%m/%d/%Y')
+                    str_date = datetime.strftime(date, '%Y-%m-%d')
+                    row.update({key: str_date})
+                except ValueError:
+                    pass
```
ukanga: I think the section above will be cleaner if we re-wrote it to:
```python
             for key in xl_date_columns:
                 try:
                     date = datetime.strptime(row.get(key, ''), '%m/%d/%Y')
                 except ValueError:
                     pass
                else:
                     row.update({key: datetime.strftime(date, '%Y-%m-%d')})

```

PR #1582, opened 2019-03-05
Only load current Image being viewed on `forms/{username}/photos`

### LINKED ISSUE #1560: Limit number of images fetched when `/forms/{ form_id_string }/photos` is hit.
**Actual behaviour**

When this endpoint is hit `/username/forms/{ form_id_string }/ photos`,  links to as many photos as there are in all submissions to that form are fetched. 
This subsequently leads to as many calls to  `/attachments/small`. 

There needs to be a rearchitecting of how this is done. 

**Suggested implementations**

1.  Limit number of subsequent calls to `/attachment/small` to one at a time.
2. Paginate the response (limit to a given number say 10).
3. Do away with this feature in classic

lincmba:
This addresses the implementation mentioned here https://github.com/onaio/onadata/issues/1560#issuecomment-469560305
 
fixes #1560

### DIFF
--- a/onadata/apps/main/views.py
+++ b/onadata/apps/main/views.py
@@ -1086,7 +1086,7 @@ def form_photos(request, username, id_string):
     image_urls = json.dumps(image_urls)
 
     data['images'] = image_urls
-    data['profilei'], created = UserProfile.objects.get_or_create(user=owner)
+    data['profile'], created = UserProfile.objects.get_or_create(user=owner)
 
     return render(request, 'form_photos.html', data)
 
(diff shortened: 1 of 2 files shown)


onadata/apps/main/views.py
```diff
@@ -1076,13 +1076,15 @@ def form_photos(request, username, id_string):
...
                 url = reverse(attachment_url, kwargs={'size': i})
                 url = '%s?media_file=%s' % (url, attachment.media_file.name)
                 data[i] = url
 
             image_urls.append(data)
 
+    image_urls = json.dumps(image_urls)
+
     data['images'] = image_urls
     data['profilei'], created = UserProfile.objects.get_or_create(user=owner)
```
ukanga: Can you fix this typo on this PR, please?

Issue #1601, opened 2019-05-01
DataError: value too long for type character varying(100)


ukanga:
```
DataError: value too long for type character varying(100)

(15 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 480, in dispatch
    response = handler(request, *args, **kwargs)
  File "onadata/apps/api/viewsets/project_viewset.py", line 164, in destroy
    project.soft_delete(user)
  File "onadata/apps/logger/models/project.py", line 133, in soft_delete
    form.soft_delete(user=user)
  File "onadata/apps/logger/models/xform.py", line 899, in soft_delete
    self.save(update_fields=update_fields)
  File "onadata/apps/logger/models/xform.py", line 873, in save
    super(XForm, self).save(*args, **kwargs)

DataError: value too long for type character varying(100)

```

PR #1604, opened 2019-05-21
Better Error message in the case of a long id_string

WinnyTroy:


WinnyTroy:
Addresses #1601

onadata/apps/logger/models/xform.py
```diff
@@ -870,6 +870,15 @@ def save(self, *args, **kwargs):
         if 'skip_xls_read' in kwargs:
             del kwargs['skip_xls_read']
 
+        if (self.id_string and len(
+            self.id_string) > self.MAX_ID_LENGTH) or \
+            (self.sms_id_string and len(
+                self.sms_id_string) > self.MAX_ID_LENGTH):
+            raise XLSFormError(
+                    _(u'The XForm id_string provided exceeds'
+                        ' the stipulated 100 characters.'
```
ukanga: I see we have a value for `MAX_ID_LENGTH` could we use it in the error message? I will propose:
```python
_(u'The XForm id_string provided exceeds %s characters. '
     'Please change the 'id_string' or "form_id" values in settings sheet or '
      'reduce the file name if you do not have a settings sheets.' % self.MAX_ID_LENGTH)
```

WinnyTroy: Changed this to include the `MAX_ID_LENGTH` attribute

PR #1617, opened 2019-06-04
Pin temptoken authentication to master.

### LINKED ISSUE #1616: GDPR pop up on zebra on subsequent logins
Production

### Expected behavior
When you log in the first time(using Zebra), you should get the GDPR pop up. Subsequent logins should not lead to this prompt.

### Actual behavior
Subsequent logins intermittently show the GDPR pop-ups.

Logout and log in repeatedly, a number of times on Zebra, the GDPR pop up will appear in one of the logins.

lincmba:
When temptoken authentication is preceded by basic/digest authentication that generates the temptoken, the temptoken authentication fails.
  This is because the changes in master db have not been reflected on the replica db yet.
  Here we pin the the temptoken authentication to master db.
fixes #1616 

### DIFF
--- a/onadata/libs/authentication.py
+++ b/onadata/libs/authentication.py
@@ -7,16 +7,17 @@ import jwt
 from django.conf import settings
 from django.core.signing import BadSignature
 from django.db import DataError
+from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from django.utils.translation import ugettext as _
 from django_digest import HttpDigestAuthenticator
-from django.shortcuts import get_object_or_404
+from multidb.pinning import use_master
 from rest_framework import exceptions
-from rest_framework.authentication import get_authorization_header
 from rest_framework.authentication import BaseAuthentication
 from rest_framework.authentication import TokenAuthentication
-from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.authentication import get_authorization_header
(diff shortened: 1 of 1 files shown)


onadata/libs/authentication.py
```diff
@@ -113,7 +114,8 @@ def authenticate_credentials(self, key):
         try:
             if isinstance(key, bytes):
                 key = key.decode('utf-8')
-            token = self.model.objects.get(key=key)
+            with use_master:
+                token = self.model.objects.get(key=key)
```
ukanga: This means that we will always hit the master database for this request. It would be ideal to only read the master database only in the situation where the file does not exist in the read replica.

```python
try:
    token = self.model.objects.get(key=key)
except self.model.DoesNotExist:
    invalid_token = True
    if getattr(settings, 'SLAVE_DATABASES', []):
        try:
            with use_master:
                token = self.model.objects.get(key=key)
        except self.model.DoesNotExist:
            invalid_token = True
        else:
            invalid_token = False
    if invalid_token:
        raise exceptions.AuthenticationFailed(_(u'Invalid token'))
```

ukanga:
@lincmba can we also include a test for this?

onadata/libs/authentication.py
lincmba: Cool, this makes sense and looks better. I'll implement this

PR #1622, opened 2019-06-11
Limit login attempts

lincmba:
Track the login attempts using MemCache.
Lockout user if the limit is reached.
Send email upon lockout and end of the lockout.
Addresses [this](https://github.com/onaio/playbooks/issues/1298).

### DIFF
--- a/onadata/libs/authentication.py
+++ b/onadata/libs/authentication.py
@@ -86,9 +86,16 @@ class DigestAuthentication(BaseAuthentication):
             if self.authenticator.authenticate(request):
                 return request.user, None
             else:
-                login_attempts(request)
+                attempts = login_attempts(request)
+                remaining_attempts = \
+                    getattr(settings, 'MAX_LOGIN_ATTEMPTS', 5) - attempts
                 raise AuthenticationFailed(
-                    _('Invalid username/password'))
+                    _("Invalid username/password. "
+                      "For security reasons, after {} more failed "
+                      "login attempts you'll have to wait {} minutes "
+                      "before trying again.".format(
+                        remaining_attempts, getattr(
+                            settings, 'LOCKOUT_TIME', 1800) // 60)))
         except (AttributeError, ValueError, DataError) as e:
             raise AuthenticationFailed(e)
 
@@ -216,8 +223,8 @@ def check_lockout(request):
                 (getattr(settings, 'LOCKOUT_TIME', 1800) -
                  time_locked_out.seconds) / 60)
             raise AuthenticationFailed(
-                _('Locked out. Too many password attempts. '
-                  'Try again in {} minutes'.format(remaining_time)))
+                _('Locked out. Too many wrong username/password attempts. '
+                  'Try again in {} minutes.'.format(remaining_time)))
         return username
 
 
@@ -232,13 +239,16 @@ def login_attempts(request):
         attempts = cache.get('{}{}'.format(LOGIN_ATTEMPTS, username))
         if attempts >= getattr(settings, 'MAX_LOGIN_ATTEMPTS', 5):
             send_lockout_email(username)
-            return cache.set(
+            cache.set(
                 '{}{}'.format(LOCKOUT_USER, username),
                 datetime.now().strftime('%Y-%m-%dT%H:%M:%S'),
                 getattr(settings, 'LOCKOUT_TIME', 1800))
-        return
+            check_lockout(request)
+            return attempts
+        return attempts
 
-    return cache.set('{}{}'.format(LOGIN_ATTEMPTS, username), 1)
+    cache.set('{}{}'.format(LOGIN_ATTEMPTS, username), 1)
+    return cache.get('{}{}'.format(LOGIN_ATTEMPTS, username))
 
 
 def send_lockout_email(username):
--- a/onadata/libs/authentication.py
+++ b/onadata/libs/authentication.py
@@ -3,8 +3,12 @@
 """
 from __future__ import unicode_literals
 
+from datetime import datetime
+
 import jwt
 from django.conf import settings
+from django.contrib.auth.models import User
+from django.core.cache import cache
 from django.core.signing import BadSignature
 from django.db import DataError
 from django.shortcuts import get_object_or_404
@@ -20,7 +24,10 @@ from rest_framework.authtoken.models import Token
 from rest_framework.exceptions import AuthenticationFailed
 
 from onadata.apps.api.models.temp_token import TempToken
+from onadata.apps.api.tasks import send_account_lockout_email
+from onadata.libs.utils.cache_tools import LOCKOUT_USER, LOGIN_ATTEMPTS
 from onadata.libs.utils.common_tags import API_TOKEN
+from onadata.libs.utils.email import get_account_lockout_email_data
 
 ENKETO_AUTH_COOKIE = getattr(settings, 'ENKETO_AUTH_COOKIE', '__enketo')
 JWT_SECRET_KEY = getattr(settings, 'JWT_SECRET_KEY', 'jwt')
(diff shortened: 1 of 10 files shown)


onadata/libs/templates/account_lockout/lockout_start.txt
```diff
@@ -0,0 +1,12 @@
...
+{% blocktrans with lockout_time=lockout_time username=username %}
+Hi {{ username }},
+
+You have been locked out from your Ona account because of too many wrong passwords.
+Please wait for {{ lockout_time }} minutes before trying to access your account again.
+If this was not you please reach out to our support team for help.
+Thank you for choosing Ona!
+
+Thanks,
+The Team at Ona
```
ukanga: I think there should be line spacing in between the lines, i.e:
```TEXT
Hello {{ username}}

You have been locked out from your Ona account because of too many wrong passwords.

Please wait for {{ lockout_time }} minutes before trying to access your account again.

If this was not you please reach out to our support team for help.

Thank you for choosing Ona!
```

The same should also apply to the other email.

onadata/libs/utils/email.py
```diff
@@ -43,3 +42,22 @@ def get_verification_email_data(email, username, verification_url, request):
...
+
+def get_account_lockout_email_data(username, end=False):
+    """Generates both the email upon start and end of account lockout"""
+    message_path = 'account_lockout/lockout_start.txt'
+    subject_path = 'account_lockout/lockout_email_subject.txt'
+    if end:
+        message_path = 'account_lockout/lockout_end.txt'
+    ctx_dict = {
+        'username': username,
+        'lockout_time': settings.LOCKOUT_TIME / 60
```
ukanga: Access to a setting that does not have a default value in our default settings. Could we use `getattr()` instead?
```python

'lockout_time': getattr(settings, 'LOCKOUT_TIME', 3600)  / 60)
```

onadata/libs/authentication.py
```diff
@@ -185,3 +194,67 @@ def authenticate(self, request):
...
+            username, end=True)
+
+        send_account_lockout_email.apply_async(
+            args=[user.email, email_data.get('subject'),
+                  email_data.get('message_txt')])
+        # send end of lockout email 1 minute after lockout time
+        send_account_lockout_email.apply_async(
+            args=[user.email, end_email_data.get('subject'),
+                  end_email_data.get('message_txt')],
+            countdown=settings.LOCKOUT_TIME + 60)
```
ukanga: Access to a setting that does not have a default value in our default settings. We could perhaps have a constant on this module that would use `getattr()` instead?
```python

LOCKOUT_DURATION = getattr(settings, 'LOCKOUT_TIME', 3600)
```

It could also apply to `MAX_LOGIN_ATTEMPTS`.

onadata/libs/templates/account_lockout/lockout_start.txt
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+{% blocktrans with lockout_time=lockout_time username=username %}
+Hi {{ username }},
+
+You have been locked out from your Ona account because of too many wrong passwords.
+Please wait for {{ lockout_time }} minutes before trying to access your account again.
+If this was not you please reach out to our support team for help.
```
msschroeder: This should be, "If this was not you, please reach out to our support team for help"

Also, do we provide them with how to access our support team - support@ona.io - do we need to provide the email or can we just use the in-app chat?

onadata/libs/templates/account_lockout/lockout_start.txt
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+{% blocktrans with lockout_time=lockout_time username=username %}
+Hi {{ username }},
+
+You have been locked out from your Ona account because of too many wrong passwords.
+Please wait for {{ lockout_time }} minutes before trying to access your account again.
```
msschroeder: How long is the lock_out time? I thought it was a matter of hours/day not minutes.

onadata/libs/templates/account_lockout/lockout_start.txt
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+{% blocktrans with lockout_time=lockout_time username=username %}
+Hi {{ username }},
+
+You have been locked out from your Ona account because of too many wrong passwords.
```
msschroeder: I would go more with "You have entered too many wrong passwords" @rowo What are your thoughts?

onadata/libs/templates/account_lockout/lockout_end.txt
```diff
@@ -0,0 +1,11 @@
+{% load i18n %}
+{% blocktrans with username=username %}
+Hi {{ username }},
+
+You are no longer locked out of Ona, you can now log in to your account.
+Thank you for choosing Ona!
+Please contact us with any questions, we're always happy to help.
```
msschroeder: How will they contact us - are we providing a contact address or directions to how to reach us? (similar comment below) Otherwise, I am fine with this email

onadata/libs/templates/account_lockout/lockout_start.txt
rowo: Is this an email in addition to messaging in the UI where the attempts happened?

rowo: When I've seen this it's not clear that it's only a password problem — it's a login attempt (user name + password) problem. See "Warn Users Before Locking Their Account" section on https://uxplanet.org/most-common-log-in-problems-and-solutions-163a6209c0eb which uses the language, "For security reasons, after 9 more failed login attempts you'll have to wait 30 minutes before trying again." The specific number of failed attempts is useful.

onadata/libs/templates/account_lockout/lockout_start.txt
lincmba: I think 30 minutes is a reasonable amount of time after 5 invalid password attempts.

onadata/libs/templates/account_lockout/lockout_start.txt
lincmba: >Is this an email in addition to messaging in the UI where the attempts happened?
Yes it is.
>When I've seen this it's not clear that it's only a password problem — it's a login attempt (user name + password) problem. See "Warn Users Before Locking Their Account" section on https://uxplanet.org/most-common-log-in-problems-and-solutions-163a6209c0eb which uses the language, "For security reasons, after 9 more failed login attempts you'll have to wait 30 minutes before trying again." The specific number of failed attempts is useful.
Just updated the error messages on the UI. The warning starts upon first invalid login, and the number of remaining attempts in the warning decrease with every subsequent login attempt.

onadata/libs/templates/account_lockout/lockout_end.txt
lincmba: I have included `support@ona.io` in the email. Have a look and suggest any improvements on the email.

onadata/libs/templates/account_lockout/lockout_start.txt
lincmba: I think 30 minutes is a reasonable amount of time after 5 invalid login attempts.

onadata/libs/templates/account_lockout/lockout_start.txt
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+{% blocktrans with lockout_time=lockout_time username=username %}
+Hi {{ username }},
+
+You have been locked out from your Ona account because you have entered too many wrong passwords.
+
+Please wait for {{ lockout_time }} minutes before trying to access your account again.
+
+If this was not you, you can reach out to our support team for help, support@ona.io.
```
ukanga: Could we make the email configurable? Someone else other than us is likely to have a different support email than ours.

onadata/libs/templates/account_lockout/lockout_start.txt
msschroeder: That sounds good to me too

PR #1635, opened 2019-06-17
Include check for label value during sav exports

### LINKED ISSUE #1629: KeyError: 'label'
https://sentry.onalabs.org/ona/onadata/issues/70958/

```
KeyError: 'label'
(1 additional frame(s) were not displayed)
...
  File "onadata/libs/utils/common_tools.py", line 176, in function_retry
    result = func(self, *args, **kwargs)
  File "onadata/libs/utils/export_tools.py", line 212, in generate_export
    total_records=total_records
  File "onadata/libs/utils/export_builder.py", line 1185, in to_zipped_sav
    sav_options = self._get_sav_options(section['elements'])
  File "onadata/libs/utils/export_builder.py", line 1097, in _get_sav_options
    all_value_labels = self._get_sav_value_labels(xpath_var_names)
  File "onadata/libs/utils/export_builder.py", line 1017, in _get_sav_value_labels
    label = self.get_choice_label_from_dict(choice['label'])

KeyError: 'label'
```

WinnyTroy:
Fixes #1629

### DIFF
--- a/onadata/libs/utils/export_builder.py
+++ b/onadata/libs/utils/export_builder.py
@@ -1015,7 +1015,8 @@ class ExportBuilder(object):
                                 if (float(name) > int(name)) else int(name)
                         except ValueError:
                             pass
-                    label = self.get_choice_label_from_dict(choice['label'])
+                    label = self.get_choice_label_from_dict(
+                        choice.get("label", ""))
                     _value_labels[name] = label.strip()
             sav_value_labels[var_name or q['name']] = _value_labels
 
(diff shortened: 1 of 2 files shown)


onadata/libs/utils/export_builder.py
```diff
@@ -1015,8 +1015,13 @@ def _get_sav_value_labels(self, xpath_var_names=None):
                                 if (float(name) > int(name)) else int(name)
                         except ValueError:
                             pass
-                    label = self.get_choice_label_from_dict(choice['label'])
-                    _value_labels[name] = label.strip()
+                    try:
+                        label = self.get_choice_label_from_dict(
```
lincmba: Could this be a less verbose fix? 
`choice.get("label", "")`

WinnyTroy: While this would look neater, I believe that we would be making an assumption that the form's choices would always be a dictionary with the following syntax:

`{'name': 'red', 'label': 'red'}`

This does not cover instances where the form supports multiple languages.
After investigating, I realized the `get_choice_label_from_dict` function unpacks the label attribute in an instance where the choices obtained would yield a dict of this syntax:

`{'name': 'orange', 'label': {'Swahili': 'Chungwa', ' English': 'Orange'}}`

PR #1646, opened 2019-07-09
Capture users who delete submissions

### LINKED ISSUE #1640: Track users who delete submissions
When submissions are deleted, both the timestamp and user who deleted it should be tracked.
Currently, only the timestamp is logged and not the users who did the deletion.

WinnyTroy:
Fixes #1640

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_data_viewset.py
@@ -1353,10 +1353,14 @@ class TestDataViewSet(TestBase):
         self._make_submissions()
         formid = self.xform.pk
         dataid = self.xform.instances.all().order_by('id')[0].pk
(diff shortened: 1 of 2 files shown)


onadata/apps/api/tests/viewsets/test_data_viewset.py
```diff
@@ -1528,6 +1528,32 @@ def test_delete_submission_by_owner(self):
         response = view(request, pk=formid)
         self.assertEqual(len(response.data), 3)
 
+    def test_deleted_xform_instance_captures_user(self):
```
ukanga: Can we add the `deleted_by` check in the existing `test_delete_submission` function to minimize duplication of tests? https://github.com/onaio/onadata/blob/b4a8070af20ed0b797d7ee7556be36f43bf846e9/onadata/apps/api/tests/viewsets/test_data_viewset.py#L1352

PR #1647, opened 2019-07-10
Handle operational errors

WinnyTroy:


onadata/libs/utils/middleware.py
```diff
@@ -64,3 +75,20 @@ def __call__(self, request):
...
+    def process_exception(self, request, exception):
+        # Filter out OperationalError Exceptions
+        if isinstance(exception, OperationalError):
+            already_raised = getattr(settings, "ALREADY_RAISED", False)
+            if not already_raised:
+                settings.ALREADY_RAISED = True
+                with use_master:
+                    response = self.get_response(request)
+                    return response
+            settings.ALREADY_RAISED = None
```
lincmba: I'd advise keeping this value as a bool and hence have either `True` or`False` as its value and not `None`.

onadata/libs/utils/middleware.py
```diff
@@ -64,3 +75,20 @@ def __call__(self, request):
...
+
+class OperationalErrorMiddleware(BaseMiddleware):
+    """
+    Captures requests returning 500 status code.
+    Then retry it against master database.
+    """
+    def process_exception(self, request, exception):
+        # Filter out OperationalError Exceptions
+        if isinstance(exception, OperationalError):
+            already_raised = getattr(settings, "ALREADY_RAISED", False)
```
lincmba: I believe keeping this value in the settings file means all subsequent errors after the first will be ignored as this is not tracking independent requests.

ukanga: I'm not sure if the setting traverses across multiple threads, an attempt to use `threading.local()` did not work. Using settings is what seemed to work.

onadata/libs/utils/middleware.py
ukanga: I am okay with `False`.

onadata/libs/utils/middleware.py
ukanga: I am not sure we should add the setting `ALREADY_RAISED` to the setting files, it should just have life in this middleware in the event it is used. Using `getattr()` with a default value is a valid use case with no required configuration.

lincmba: I agree. `ALREADY_RAISED` variable should be defined and only used in this middleware.

ukanga: I think we need to limit the middleware to very specific `OperationaError` exceptions. Specifically the ones that contain "canceling statement due to conflict with recovery".

PR #1648, opened 2019-07-10
return integer representation of negative values

### LINKED ISSUE #1641: Negative Integer Values Returned as Decimal Value
Negative values entered in fields of Integer type through the forms are returned as decimal values when hitting the data endpoint: https://api.ona.io/api/v1/data/{formID}. 

Positive values are accurately represented as integers.

Negative values are represented accurately both in the table view of ona.io and in the xml form. The issue only arises when retrieving the json representation of the data

**Observed Behaviour**
Value entered in INT field in forms `-2`
Value returned from [data endpoint](https://api.ona.io/api/v1/data/{formID}) `-2.0`

WinnyTroy:
Fixes #1641

### DIFF
--- a/onadata/apps/logger/models/instance.py
+++ b/onadata/apps/logger/models/instance.py
@@ -103,19 +103,17 @@ class FormIsMergedDatasetError(Exception):
 
 
 def numeric_checker(string_value):
-    if string_value.isdigit():
+    try:
         return int(string_value)
-    elif string_value.startswith('-') and string_value[1:].isdigit():
-        return int(string_value)
-    else:
+    except ValueError:
         try:
             value = float(string_value)
-            if math.isnan(value):
-                value = 0
-            return value
         except ValueError:
             pass
+        else:
+            return 0 if math.isnan(value) else value
 
+    return string_value
 
 # need to establish id_string of the xform before we run get_dict since
 # we now rely on data dictionary to parse the xml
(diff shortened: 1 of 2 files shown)


onadata/apps/logger/tests/models/test_instance.py
```diff
@@ -263,3 +264,14 @@ def test_instance_json_updated_on_review(self):
...
+    def test_numeric_checker_with_negative_integer_values(self):
+        # Evaluate negative integer values
+        string_value = "-16"
+        result = numeric_checker(string_value)
+        self.assertEqual(result, -16)
+
+        # Evaluate un-signed integer values
+        string_value = "16"
+        result = numeric_checker(string_value)
+        self.assertEqual(result, 16)
```
ukanga: Can we also add checks for a `float` value, `nan` value and a non-numeric string for the `numeric_checker` function? It does not seem like we test them anywhere else.

onadata/apps/logger/models/instance.py
```diff
@@ -105,6 +105,8 @@ def __str__(self):
 def numeric_checker(string_value):
     if string_value.isdigit():
         return int(string_value)
+    elif string_value.startswith('-') and string_value[1:].isdigit():
```
ukanga: I think a more pythonic approach would be to refactor with `try .. except`.
```python
def numeric_checker(string_value):

    try:
        return int(string_value)
    except ValueError:
        try:
            value = float(string_value)
        except ValueError:
            pass
        else:
            return 0 if math.isnan(value) else value

    return None

PR #1656, opened 2019-07-17
Limit change password attempts

RayceeM:
Check if the user is locked out when changing password.
Tracks the number of attempts made to a maximum limit of 10.
Locks out user when maximum limit is reached.

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
@@ -5,6 +5,7 @@ from builtins import str
 from future.moves.urllib.parse import urlparse, parse_qs
 
 from django.contrib.auth.models import User
+from django.core.cache import cache
 from django.db.models import signals
 from django.test.utils import override_settings
 from django.utils.dateparse import parse_datetime
@@ -734,6 +735,9 @@ class TestUserProfileViewSet(TestAbstractViewSet):
         response = view(request, user='bob')
         user = User.objects.get(username__iexact=self.user.username)
         self.assertEqual(response.status_code, 400)
+        self.assertEqual(
+            response.data,
+            "Invalid password. You have 9 attempts left.")
         self.assertFalse(user.check_password(new_password))
 
     def test_profile_create_with_name(self):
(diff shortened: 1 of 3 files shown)


onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -95,6 +98,46 @@ def set_is_email_verified(profile, is_email_verified):
...
+            'Too many password reset attempts, Try again in {} minutes'.format(
+                remaining_time)}
+        return response
+
+
+def change_password_attempts(request):
+    """Track number of login attempts made by user within a specified amount
+     of time"""
+    username = request.user.username
+    attempts = cache.get('{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username))
```
ukanga: The expression `'{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username)` is repeated several times in this function, can we have it assigned to a variable and then used instead? It will be simpler to make modifications in one place than to look for all the places it appears. 

```python
your_variable = '{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username)
...
cache.incr(your_variable)
```

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -95,6 +98,46 @@ def set_is_email_verified(profile, is_email_verified):
...
+        cache.incr('{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username))
+        attempts = cache.get('{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username))
+        if attempts >= getattr(settings, 'MAX_CHANGE_PASSWORD_ATTEMPTS', 10):
+            cache.set(
+                '{}{}'.format(LOCKOUT_CHANGE_PASSWORD_USER, username),
+                datetime.now().strftime('%Y-%m-%dT%H:%M:%S'),
+                getattr(settings, 'LOCKOUT_TIME', 1800))
+            if check_user_lockout(request):
+                return check_user_lockout(request)
+            return attempts
```
ukanga: I think we can do without this return statement, there is one on 134 already does the same thing.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -155,19 +198,32 @@ def change_password(self, request, *args, **kwargs):  # noqa
...
+                if user_profile.user.check_password(current_password):
+                    metadata = user_profile.metadata or {}
+                    metadata['last_password_edit'] = timezone.now().isoformat()
+                    user_profile.user.set_password(new_password)
+                    user_profile.metadata = metadata
+                    user_profile.user.save()
+                    user_profile.save()
+
+                    return Response(status=status.HTTP_204_NO_CONTENT)
+                else:
```
ukanga: We could also do without the else section since we have a return in the above if block.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -95,6 +98,46 @@ def set_is_email_verified(profile, is_email_verified):
...
+    attempts = cache.get('{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username))
+
+    if attempts:
+        cache.incr('{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username))
+        attempts = cache.get('{}{}'.format(CHANGE_PASSWORD_ATTEMPTS, username))
+        if attempts >= getattr(settings, 'MAX_CHANGE_PASSWORD_ATTEMPTS', 10):
+            cache.set(
+                '{}{}'.format(LOCKOUT_CHANGE_PASSWORD_USER, username),
+                datetime.now().strftime('%Y-%m-%dT%H:%M:%S'),
+                getattr(settings, 'LOCKOUT_TIME', 1800))
```
ukanga: Considering `getattr(settings, 'LOCKOUT_TIME', 1800)` does not rely on any variable in the function, could we have it as a static variable on top of the page? We could use the same variable in all places this appears on this file, i.e.
```python
LOCKOUT_TIME = getattr(settings, 'LOCKOUT_TIME', 1800)
..
```
The same applies to `getattr(settings, 'MAX_CHANGE_PASSWORD_ATTEMPTS', 10)`.

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -1151,3 +1155,61 @@ def test_monthly_submissions_with_year_param(self):
...
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data,
+                         "Invalid password. You have 8 attempts left.")
+        self.assertEqual(cache.get('change_password_attempts-bob'), 2)
+
+        # check user is locked out
+        current_password = "wrong_pass"
+        new_password = "bobbob1"
+        post_data = {'current_password': current_password,
+                     'new_password': new_password}
```
ukanga: Lines 1195 to 1198 seem exactly the same as lines 1183 to 1186, I think this is duplicate code that is not necessary.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -45,6 +48,9 @@
     UserProfileSerializer
 
 BaseViewset = get_baseviewset_class()  # pylint: disable=invalid-name
+Lockout_time = getattr(settings, 'LOCKOUT_TIME', 1800)
+Max_change_password_attempts = getattr(
```
ukanga: The naming conventions for this would be all in caps. i.e. `LOCKOUT_TIME = getattr(settings, 'LOCKOUT_TIME', 1800)`

PR #1658, opened 2019-07-18
Bulk creation of submission reviews

### LINKED ISSUE #1623: Submission Review fails when payload is used
The feature works well when data is sent as Form Data but fails when a json Payload is used.

lincmba:
Bulk creation of reviews was only implemented and tested at the serializer but not at the Viewset.
 This PR addresses this with a custom create function to handle bulk creation of submission reviews.
 fixes #1623

Signed-off-by: lincmba <lincolncmba@gmail.com>

### DIFF
--- a/docs/submission_review.rst
+++ b/docs/submission_review.rst
@@ -1,0 +1,214 @@
+SubmissionReview
+****************
+
+This endpoint supports List, Retrieve, Update, Create Submission Reviews
+
+Where:
+
+- ``id`` - the id of the submission review (required when doing an update)
+- ``instance`` - the id of the Instance object being reviewed
+- ``note`` - the submission review comment
+- ``status`` - the submission review status. Should be one of:
+
+    - ``'1'`` - approved
+    - ``'2'`` - rejected
+    - ``'3'`` - pending
+
+Note must be provided incase the status in '2' (rejected)
+
+Make a Submission Review
+------------------------
+.. raw:: html
+
+    <pre class="prettyprint">
+    <b>POST</b> /api/v1/submissionreview.json</pre>
+
+Example
+^^^^^^^
+::
+
+    curl -X POST -H "Content-type:application/json" -d '{"status":"1","instance":1337,"note":"This was approved because it is awesome!"}' https://example.com/api/v1/submissionreview.json
+
+Response
+^^^^^^^^^
+::
+
+    {
+        "id": 4,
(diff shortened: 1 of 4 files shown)


onadata/apps/api/tests/viewsets/test_submission_review_viewset.py
```diff
@@ -351,3 +351,23 @@ def test_submission_review_created_by_filter(self):
...
+        """
+        Test that we can bulk create approved submission reviews without a note
+        """
+        instances = self.xform.instances.all()
+        submission_data = [
+            {
+                'instance': _.id,
+                'status': SubmissionReview.APPROVED
+            } for _ in instances
+        ]
```
ukanga: If I recall correctly, a user would select a group of submissions and either approve them or reject all of them. The note made will apply to all selected submissions. Would this mean that we need a list of submission IDs and a single status that applies to all? Or are we expecting each submission ID will have a status that may not be the same as the rest?

ivermac: I think it means the selected submission IDs would all have the same single status. Unless I'm mistaken @moshthepitt

lincmba: I think it would be best to allow for bulk creation of the SubmissionReviews regardless of consistency in the status. All data in the list submitted to be processed individually. Hence allowing bulk creation with different statuses.

Having a list of submission IDs and a single status for all will be a little restrictive from our end I think.

lincmba:
It appears the whole feature needs documentation.

docs/submission_review.rst
```diff
@@ -0,0 +1,214 @@
...
+This endpoint supports List, Retrieve, Update, Create Submission Reviews
+
+Where:
+
+- ``id`` - the id of the submission review (required when doing an update)
+- ``instance`` - the id of the Instance object being reviewed
+- ``note`` - the submission review comment
+- ``status`` - the submission review status. Should be one of:
+
+    - ``'1'`` - approved
```
ukanga: Does this mean it has to be a string? Can it be a number as well? Or should we better yet use the string `"approved" "rejected" "pending"` which are far much clearer?

ivermac: I would say will leave it as it is because the documentation explains what each digit means.

Issue #1664, opened 2019-07-26
Add total number of attachments in headers


RayceeM:
Add in the header the total number of attachments in the form

PR #1665, opened 2019-07-26
Add count parameter to media endpoint

RayceeM:
Fixes issue #1664


onadata/apps/api/viewsets/attachment_viewset.py
```diff
@@ -83,6 +84,16 @@ def retrieve(self, request, *args, **kwargs):
 
         return Response(serializer.data)
 
+    @action(methods=['GET'], detail=False)
+    def count(self, request, *args, **kwargs):
+        response = {}
+        xform = request.query_params.get('xform')
+        total_count = Attachment.objects.filter(
```
ukanga: I think it is best we rely on using `self.filter_queryset(self.get_queryset())` so that we maintain the queryset and filters to applied to it by the viewset. Hence
```python
data = {
  "count":  self.filter_queryset(self.get_queryset()).count()
}

return Response(data=data)
```

PR #1670, opened 2019-08-09
Override django inbuilt password reset token generation

WinnyTroy:
Fixes playbooks#1410

Introduces the user email during token generation.
This would invalidate the token when the user updates their email.

### DIFF
--- a/onadata/libs/serializers/password_reset_serializer.py
+++ b/onadata/libs/serializers/password_reset_serializer.py
@@ -2,8 +2,7 @@ from builtins import bytes as b
 from future.moves.urllib.parse import urlparse
 
 from django.contrib.auth.models import User
-from django.contrib.auth.tokens import default_token_generator,\
-    PasswordResetTokenGenerator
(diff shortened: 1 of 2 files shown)


onadata/libs/serializers/password_reset_serializer.py
```diff
@@ -12,10 +12,25 @@
...
+        # Include user email alongside user password to the generated token
+        # as the user state object that might change after a password reset
+        # to produce a token that invalidated.
+        login_timestamp = '' if user.last_login is None\
+            else user.last_login.replace(microsecond=0, tzinfo=None)
+        return str(user.pk) + user.password + user.email +\
+            str(login_timestamp) + str(timestamp)
+
+
+custom_default_token_generator = CustomPasswordResetTokenGenerator()
```
ukanga: Let's just make it default_token_generator. i.e.:
```python
default_token_generator = CustomPasswordResetTokenGenerator()
```

WinnyTroy: This is done.

PR #1674, opened 2019-08-20
Clean elements of bind attributes when creating merged datasets

WinnyTroy:
Fixes onaio/zebra#5683

Newly created merged-datasets have bind attributes that contain relevance fields references to fields that were removed when generating the merged datasets. 

This PR seeks to drop/ignore the relevant column coming from the bind attributes of child elements in the newly created merged-datasets.

### DIFF
--- a/onadata/libs/tests/serializers/test_merged_xform_serializer.py
+++ b/onadata/libs/tests/serializers/test_merged_xform_serializer.py
@@ -67,18 +67,39 @@ GROUP_A_MD = """
 
 GROUP_B_MD = """
 | survey |
-|        | type              | name   | label |
-|        | text              | name   | Name  |
-|        | begin group       | info   | Info  |
-|        | integer           | age    | Age   |
-|        | select one gender | gender | Sex   |
-|        | end group         |        |       |
-|        | begin group       | other  | Other |
-|        | begin group       | person | Person |
-|        | integer           | age    | Age   |
-|        | select one gender | gender | Sex   |
-|        | end group         |        |       |
-|        | end group         |        |       |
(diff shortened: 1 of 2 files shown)


onadata/libs/tests/serializers/test_merged_xform_serializer.py
```diff
@@ -424,3 +445,20 @@ def test_matching_fields_by_type(self):
...
+        Test get_merged_xform_survey(): should not contain bind elements.
+        """
+        self.project = get_user_default_project(self.user)
+        xform1 = self._publish_markdown(GROUP_A_MD, self.user, id_string='a')
+        xform2 = self._publish_markdown(GROUP_B_MD, self.user, id_string='b')
+        xform3 = self._publish_markdown(GROUP_C_MD, self.user, id_string='c')
+        survey = get_merged_xform_survey([xform1, xform2, xform3])
+
+        result = survey.to_json_dict()
+        count = [child for child in result["children"] if 'bind' in child]
```
ukanga: Should this be a number? i.e.:
``python
count = len([child for child in result["children"] if 'bind' in child])
... 
self.assertEqual(count, 0)
````

ukanga:
I think we should in future add this support back, we would need to confirm that any references in the bind attributes are in both forms being merged and remove only those attributes where references might refer to elements not in both forms.

onadata/libs/tests/serializers/test_merged_xform_serializer.py
WinnyTroy: Made this change.

PR #1680, opened 2019-09-02
Validate auth user username before creating Registration Profile

### LINKED ISSUE #936: IntegrityError at /api/v1/orgs when creating an org or a user account on /api/v1/profiles
```python
Internal Server Error: /api/v1/orgs

IntegrityError at /api/v1/orgs
duplicate key value violates unique constraint "auth_user_username_key"
DETAIL:  Key (username)=(distmanagement) already exists.
```

- [x] Handle on `/api/v1/orgs` API endpoint
- [x] Handle on `/api/v1/profiles` API endpoint

WinnyTroy:
During registration of a new account profile, a new instance of the user model, `Django’s django.contrib.auth.models.User` is created to represent the new account, with the is_active field set to False. 

This occurs in the Registration Profile model.

Onadata validates the username, email and twitter field provided during the creation of a user registration profile against the auth User model.
This PR seeks to catch the IntegrityError exception and return a proper message to the user if the exception is triggered during creation of a user profile. Resolves #936

### DIFF
--- a/onadata/libs/serializers/user_profile_serializer.py
+++ b/onadata/libs/serializers/user_profile_serializer.py
@@ -245,7 +245,8 @@ class UserProfileSerializer(serializers.HyperlinkedModelSerializer):
                 send_email=settings.SEND_EMAIL_ACTIVATION_API)
         except IntegrityError:
             raise serializers.ValidationError(_(
-                u"User account already exists"
+                u"User account {} already exists".format(
+                    params.get('username'))
             ))
         new_user.is_active = True
         new_user.first_name = params.get('first_name')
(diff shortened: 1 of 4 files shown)


onadata/libs/serializers/user_profile_serializer.py
```diff
@@ -236,12 +236,17 @@ def create(self, validated_data):
...
+        try:
+            new_user = RegistrationProfile.objects.create_inactive_user(
+                username=params.get('username'),
+                password=params.get('password1'),
+                email=params.get('email'),
+                site=site,
+                send_email=settings.SEND_EMAIL_ACTIVATION_API)
+        except IntegrityError:
+            raise serializers.ValidationError(_(
+                u"User account already exists"
```
ukanga: We should include the username in this message and add checks in tests.

PR #1686, opened 2019-09-10
open-data schema endpoint fixes

WinnyTroy:


onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -175,22 +175,28 @@ def schema(self, request, **kwargs):
...
             xform_json = json.loads(xform.json)
             self.flatten_xform_columns(
                 json_of_columns_fields=xform_json.get('children'))
 
-            tableau_column_headers = self.get_tableau_column_headers()
+            # Get data fields
+            instances = Instance.objects.filter(xform=xform.pk)
+
+            if instances:
+                submission_data_fields = [k for k in instances[0].json]
```
ukanga: We are making a huge assumption that the first submission will have all fields in the form. This may not always be the case especially if we have skip logic in the form. For example, the first record on the JSON below does not have the field fruit, but it is present in the second record. We are likely to send only one column when the two columns should be part of the data being sent.
```JSON
[
  {
    "_id": 38741,
    "your_choice": "yes",
  },
  {
    "_id": 38742,
    "fruit": "orange",
    "your_choice": "no",
  }
]
```

WinnyTroy: Tableau has a refresh data source feature that is able to pull in the new fields that come in with subsequent new submissions. So, if the first submission does not have responses for some fields and the second comes in with those responses filled, the dashboard structure gets refreshed with the new fields and the slots for the first submission that lacked responses for those fields are filled with a null value on Tableau's table view.

This has been tested by @bkimondiu  and @samwata with v2018.2, v2019.1, v2019.2 and v2019.3

PR #1691, opened 2019-10-01
Fix form level permission restrictions on search

### LINKED ISSUE #1692: Form Permissions are not applied when searching submissions
The form permission restriction that restricts users with can submit permission to view their own submitted data is not being applied when searching submissions. One is able to view data submitted by other user's.

# Steps to reproduce:
1. Create form (form_a) in project (project_a) and make submissions
2. Share project_a with another user (user_b) with can submit permissions.
3. Login with user_b account and make submissions.
4. Go to table tab, can only view user_b submissions
5. Start searching, see all submissions

lincmba:
Form level permission restrictions were failing upon searching the data.
This PR addresses this issue.

fixes #1692

Signed-off-by: lincmba <lincolncmba@gmail.com>

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_data_viewset.py
@@ -3,12 +3,12 @@ from __future__ import unicode_literals
 import datetime
 import json
 import os
+from builtins import open
 from datetime import timedelta
 from tempfile import NamedTemporaryFile
 
 import geojson
 import requests
-from builtins import open
 from django.test import RequestFactory
 from django.test.utils import override_settings
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


ukanga:
@lincmba Is it that the `query` parameter is a list instead of a string? And we only handle the case when it is a string and not a list? From the test, it is not clear that is what you check.

lincmba:
@ukanga 
> Is it that the query parameter is a list instead of a string? And we only handle the case when it is a string and not a list? From the test, it is not clear that is what you check.

Yes this is exactly what was happening. We only handled the case when its either a dictionary ie `query = {"_submitted_by": "bob"}` or when its a string i.e `query="ambulance"` but not when its a list of both `query=[{"_submitted_by": "bob"}, "ambulance"]`. 

I have included more tests for clarity.

PR #1695, opened 2019-10-02
Modify Flow Results Response Endpoint Formatting

### LINKED ISSUE #1690: Return flow results in correct flow-results-spec format
Flow results responses returned from the flow `/api/v1/flow-results/packages/[form_uuid]/responses` is of the format:
```
{
  “data”: {
    “responses”: [
 
    ]
  }
}
```

The correct flow results responses [here](https://floip.gitbooks.io/flow-results-specification/content/api-specification.html#publish-responses-to-a-package) requires it to be in the format:
```
{
  "data": {
    "attributes": {
      "responses": [
        
      ]
    }
  }
}
```

DavisRayM:
This PR modifies the `/api/v1/flow-results/packages/[form_uuid]/responses` endpoints response in order to adhere to the [flow results specification](https://floip.gitbooks.io/flow-results-specification/content/api-specification.html#publish-responses-to-a-package)

Fixes #1690

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_floip_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_floip_viewset.py
@@ -17,8 +17,7 @@ class TestFloipViewSet(TestAbstractViewSet):
     Test FloipViewSet class.
     """
 
-    def _publish_floip(self,
-                       path='flow-results-example-2-api.json',
+    def _publish_floip(self, path='flow-results-example-2-api.json',
                        test=True):
         view = FloipViewSet.as_view({'post': 'create'})
         path = os.path.join(os.path.dirname(__file__), "../", "fixtures", path)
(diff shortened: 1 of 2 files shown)


DavisRayM:
> Do we have any test for this? Could you add one, please?

There currently isn't a test for this. I'll add a test to cover this particular case

onadata/apps/api/tests/fixtures/flow-results-responses-list-format.json
```diff
@@ -0,0 +1,9 @@
+{
+    "data": {
+        "type": "flow-results-data",
+        "id": null,
+        "attributes": {
+            "responses": []
+        }
+    }
+}
```
ukanga: I don't think we should have this as an external file - can we make it inline below in the test?
```python
sample_response = {
    "data": {
        "type": "flow-results-data",
        "id": null,
        "attributes": {
            "responses": []
        }
    }
}
...
..
.
self.assertEqual(response.data, sample_response['data'])
```

DavisRayM: Yes we can. Made the change in the latest commits

PR #1698, opened 2019-10-09
Add deleted_at to attachments

### LINKED ISSUE #1696: Submission attachments list is not updated correctly after replacing an old image/file with a new one.
All

### Expected behaviour

When an image/file in a submission is replaced, the old image/file should be replaced with the new one and deleted from the attachments list.

### Actual behaviour

The old image/file is replaced successfully but the old image/file is still retained in the attachments list.

Use Enketo to edit a submission that has an attachment. Check the attachments for that submission and you will notice the previous attachment is still available.

---

### Implementation plan

- 
- 
-

WinnyTroy:
Add deleted_at Field to Attachments. 
Create soft_delete action that will populate deleted_at field for attachments. 
Resolves #1696


onadata/apps/logger/models/attachment.py
```diff
@@ -86,3 +87,14 @@ def file_hash(self):
...
         if self.media_file:
             return os.path.basename(self.media_file.name)
+
+    def soft_delete(self, user=None):
+        """
+        Soft deletes an attachment by adding a deleted_at timestamp.
+        """
+
+        soft_deletion_time = timezone.now()
+        self.deleted_at = soft_deletion_time
```
pld: I don't see the value in a separate variable here, why not put `timezone.now()` right here instead of assigning it to the var first?

onadata/libs/serializers/attachment_serializer.py
```diff
@@ -54,7 +54,7 @@ class AttachmentSerializer(serializers.HyperlinkedModelSerializer):
     class Meta:
         fields = ('url', 'filename', 'mimetype', 'field_xpath', 'id', 'xform',
                   'instance', 'download_url', 'small_download_url',
-                  'medium_download_url')
+                  'medium_download_url', 'deleted_at')
```
pld: is this exposing `deleted_at` to users? This should be handled opaquely, ie I don't see how/why a user needs to see this ever, in an edge-case an dev might need to see this but I'd rather wait to understand why we'd expose this to a user before doing so

WinnyTroy:
> is there a test that `deleted_at` is set to the user that deleted it correctly? I didn't see it, maybe I missed it

I dont believe so. `deleted_at` is set to a datetime object.

pld:
> > is there a test that `deleted_at` is set to the user that deleted it correctly? I didn't see it, maybe I missed it
> 
> I dont believe so. `deleted_at` is set to a datetime object.

my mistake, I meant to write `delete_by`

can we please add this test if it doesn't exist

WinnyTroy:
> > > is there a test that `deleted_at` is set to the user that deleted it correctly? I didn't see it, maybe I missed it
> > 
> > 
> > I dont believe so. `deleted_at` is set to a datetime object.
> 
> my mistake, I meant to write `delete_by`
> 
> can we please add this test if it doesn't exist

Sure, let me confirm, and include one if none exists

onadata/apps/api/tests/viewsets/test_attachment_viewset.py
```diff
@@ -229,8 +229,10 @@ def test_list_view_surfaces_current_attachment(self):
         self.assertEqual(len(response.data), 5)
 
         # test when the attachment is soft deleted
-        self.attachment.soft_delete()
+        self.attachment.soft_delete(user=self.user)
         self.attachment.save()
+        # Test that deleted_by field captures the right user
+        self.assertTrue(self.attachment.deleted_by, self.user)
```
pld: Can we move this to a separate test please?

So not assign a user here, add a new independent test that assigns user and tests they are assigned to delete object

Having each test only test for one thing is ideal

WinnyTroy: sure. Let me move that.

onadata/apps/api/viewsets/attachment_viewset.py
```diff
@@ -100,7 +100,9 @@ def list(self, request, *args, **kwargs):
                 if not xform.shared_data:
                     raise Http404(_("Not Found"))
 
-        self.object_list = self.filter_queryset(self.get_queryset())
+        request_data = self.filter_queryset(self.get_queryset())
+        self.object_list = self.object_list = [
```
DavisRayM: Would filtering this while declaring the `queryset` work ? Like so `Attachment.objects.filter(instance__deleted_at__isnull=True, deleted_at__isnull=True)` ?

onadata/apps/logger/models/attachment.py
```diff
@@ -86,3 +87,12 @@ def file_hash(self):
     def filename(self):
         if self.media_file:
             return os.path.basename(self.media_file.name)
+
+    def soft_delete(self, user=None):
+        """
+        Soft deletes an attachment by adding a deleted_at timestamp.
+        """
+        self.deleted_at = timezone.now()
+        if user is not None:
```
DavisRayM: Would this work if we kept `if user:` here ?

onadata/apps/api/tests/viewsets/test_attachment_viewset.py
```diff
@@ -188,6 +188,72 @@ def test_list_view(self):
...
+    def test_soft_delete_action_returns_correct_user(self):
+        self._submit_transport_instance_w_attachment()
+
+        request = self.factory.get('/', **self.extra)
+        response = self.list_view(request)
+        self.assertEqual(len(response.data), 1)
+
+        # test when the attachment is soft deleted
+        self.attachment.soft_delete(user=self.user)
+        self.attachment.save()
```
DavisRayM: Why do we call `save()` here ?

onadata/apps/logger/models/attachment.py
WinnyTroy: Sure. I'll change that.

onadata/apps/api/viewsets/attachment_viewset.py
WinnyTroy: Aaaaaaah...totally makes sense. I'll change that.

onadata/apps/api/tests/viewsets/test_attachment_viewset.py
WinnyTroy: I see how this may be repetitive. Removing that.

WNjihia:
@WinnyTroy I uploaded a submission with an image then edited the submission and uploaded a different image. It seems like both images are still being fetched. Also kindly check on the photo tab view for the form - I noticed that the deleted image is still being fetched here too and in an instance where I edited the submission and deleted the image without uploading a new one, the deleted image still appears on the photo tab view
cc: @DavisRayM  @pld

WinnyTroy:
> @WinnyTroy I uploaded a submission with an image then edited the submission and uploaded a different image. It seems like both images are still being fetched. Also kindly check on the photo tab view for the form - I noticed that the deleted image is still being fetched here too and in an instance where I edited the submission and deleted the image without uploading a new one, the deleted image still appears on the photo tab view
> cc: @DavisRayM @pld

this seems to be a deployment issue. I'll re-deploy the changes and request for QA once more. Appologies

DavisRayM:
Added a _Blocked_ label as this is being blocked by the discussion ongoing [here](https://github.com/onaio/onadata/issues/1696#issuecomment-575650501).

WinnyTroy:
> Added a _Blocked_ label as this is being blocked by the discussion ongoing [here](https://github.com/onaio/onadata/issues/1696#issuecomment-575650501).

Removing this label

onadata/libs/utils/logger_tools.py
```diff
@@ -240,6 +240,13 @@ def save_attachments(xform, instance, media_files):
                 mimetype=content_type,
                 name=filename,
                 extension=extension)
+        replaced_attachments = (
+            a for a in Attachment.objects.filter(instance=instance)
+            if a.media_file.name.split('/')[-1] not in media_files)
+
+        if replaced_attachments:
+            for a in replaced_attachments:
```
pld: a single command should be able to soft_delete these all at once without a loop, e.g. setting the `deleted_at` value on all records matching the IDs in `replaced_attachments`

onadata/libs/utils/logger_tools.py
```diff
@@ -240,6 +240,13 @@ def save_attachments(xform, instance, media_files):
                 mimetype=content_type,
                 name=filename,
                 extension=extension)
+        replaced_attachments = (
```
pld: I'm unclear on why the added code is inside the `f in media_files` loop, why is that necessary?

onadata/libs/utils/logger_tools.py
```diff
@@ -240,6 +240,13 @@ def save_attachments(xform, instance, media_files):
...
+            a for a in Attachment.objects.filter(instance=instance)
+            if a.media_file.name.split('/')[-1] not in media_files)
```
pld: I believe there's a way to push this all into a single query, but it may require a raw SQL query (it may not, so worth investigating that first) and it may not be worthwhile, since I'm not sure about the performance of the SQL operators used, with those warnings in place..

It is generally faster to use a single query instead of a loop. You can put the string search into a where clause searching a joined `media_file` table with something like

```sql
REVERSE(SPLIT_PART(REVERSE(a.media_file.name, '/', 1))) NOT LIKE %{media_files.join('|')}
```

that's pseudocode, but hopefully the idea is clear, the inner reverse to find the last occurs of the delimiter, the outer reverse so it's in the correct other for the LIKE clause that joins all the media file text.

onadata/libs/utils/logger_tools.py
WinnyTroy: not really. Removed that.

onadata/libs/utils/logger_tools.py
WinnyTroy: After some research, I am thinking of working with django's [F expression](https://docs.djangoproject.com/en/3.0/ref/models/expressions/#f-expressions) for this. I mean, we could still  use raw sql, but i have a feeling the queryset api could adequately handle this, for this use case.

Using F expressions allows you to perform manipulations on model fields without loading them into memory.
So a clean way to filter and update this attachments could be:

`Attachment.objects.filter(instance=instance).filter(~Q(name_in=filenames).update(deleted_at=F('deleted_at') timezone.now())`

Here we could filter out the queryset for:
1. Attachments for that specific submission
2. Attachment that are not present in the media files recieved

and consequently update the deleted field for those attachments, in a way 'soft deleting' those attachments.

But I guess this could work too:
`Attachment.objects.filter(instance=instance).filter(~Q(name_in=filenames).update(deleted_at=timezone.now())`

without adding another layer of complexity to the query.
Any take on this?

pld: This looks great, I'm pretty sure Django will lazy-evaluate this query, which means you can split this into 2 steps

1) define the queryset
2) pass that queryset to a soft_delete function

so in this code you'd define the queryset and pass it:

```python
to_delete_queryset = Attachment.objects.filter(instance=instance).filter(~Q(name_in=filenames)
Attachment.soft_delete()
```

or actually, in a single call is probably fine:

```python
Attachment.soft_delete(Attachment.objects.filter(instance=instance).filter(~Q(name_in=filenames))
```



then in `models/attachment.py` add the class function

```python
@classmethod
def safe_delete(cls, queryset)
    queryset.update(deleted_at=timezone.now())
```

I think you can overload class/non-class methods like that, not sure if it's best practice.

The reason I prefer splitting it up, is so that if we can modify the logic for soft deletion in one place in the future, instead of having to comb through the code and find place where we set `deleted_at`, e.g. suppose we change the name of the `deleted_at` column, or more likely suppose we want to explicitly log or call a metric service when we soft delete something.

onadata/apps/logger/models/instance.py
```diff
@@ -79,6 +79,9 @@ def _get_attachments_from_instance(instance):
         attachment['id'] = a.id
         attachments.append(attachment)
 
+        if isinstance(a.deleted_at, datetime):
```
DavisRayM: Since `deleted_at` defaults to null, couldn't we just change this to `if a.deleted_at` ?

onadata/apps/logger/models/attachment.py
```diff
@@ -86,3 +90,12 @@ def file_hash(self):
...
             return os.path.basename(self.media_file.name)
+
+    @classmethod
+    def soft_delete(cls, queryset, user=None):
+        """
+        Soft deletes an attachment by adding a deleted_at timestamp.
+        """
+        queryset.update(deleted_at=timezone.now())
+        if user:
+            queryset.update(deleted_by=user)
```
ivermac: Can the above function be edited to take an instance instead of a queryset? Then we can use the `cls` object?
```python
def soft_delete(cls, instance, user=None):
    ...
    queryset = cls.objects.filter(name__in=instance.get_expected_media())
    kwargs = {'deleted_at': timezone.now()}
    if user:
        kwargs.update({'deleted_by': user})
    queryset.update(**kwargs)
```

This would then the classmethod would be called as:
```python
...
Attachment.soft_delete(instance)
```

onadata/libs/utils/logger_tools.py
```diff
@@ -241,6 +241,9 @@ def save_attachments(xform, instance, media_files):
                 mimetype=content_type,
                 name=filename,
                 extension=extension)
+    Attachment.soft_delete(Attachment.objects.filter(instance=instance).filter(
+        ~Q(name__in=instance.get_expected_media())))
```
ivermac: Any reason why we need to filter by instance then filter by an instance method? If we filtered by the instance method, would it give us the desired result? Like the following:
```python
Attachment.soft_delete(Attachment.objects.filter(name__in=instance.get_expected_media()))
```
And if the above is possible, can we filter it inside the classmethod as suggested above?

onadata/libs/utils/logger_tools.py
```diff
@@ -241,6 +241,9 @@ def save_attachments(xform, instance, media_files):
...
+
```
ukanga: Why is it relevant to delete attachments every time we receive a submission?

WinnyTroy: This allows us to apply soft deletion action against the replaced attachments, if any.
At the moment, if a user edits a submission with 4 attachments and replaces 1 of the attachments with a new media file, the replaced attachment is retained as a valid attachment for that instance and not really deleted, rather a new attachment is created for that submission

Having this filter will allow us know and delete attachments no longer prevalent for that instance.

ukanga: I would recommend we add an optional field such that we only delete on edits. We can apply true to this field at https://github.com/onaio/onadata/blob/df3d077d6467e07b1d69003de0efc04a442aee74/onadata/libs/utils/logger_tools.py#L320.

```python
def save_attachments(xform, instance, media_files, remove_deleted_media=False):
...
        
    if remove_deleted_media:
       Attachment.soft_delete(instance.attachments.filter(~Q(name__in=instance.get_expected_media())))
```

onadata/libs/utils/logger_tools.py
ukanga: We do need to only affect the current submission and not all attachments in the database. Perhaps you meant:

```python
Attachment.soft_delete(instance.attachments.filter(~Q(name__in=instance.get_expected_media())))
```

onadata/libs/utils/logger_tools.py
WinnyTroy: Sure. I have included this optional field.

onadata/apps/logger/models/attachment.py
pld: well actually, I didn't realize this was relative to an instance when I suggested adding a class method, so that's my mistake.

If this is always relative to an instance, I'd think it's better as an instance method, and we'd call with `instance.soft_delete()`, have `queryset = self.__class__.objects.filter(name__in=self.get_expexted_media())`

Is there a problem w/that?

onadata/apps/logger/models/attachment.py
```diff
@@ -1,8 +1,11 @@
 import mimetypes
 import os
 from hashlib import md5
+from django.db.models import Q
```
pld: Codacy Issue found: [Unused Q imported from django.db.models](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4290671)

onadata/apps/logger/models/attachment.py
```diff
@@ -86,3 +90,16 @@ def file_hash(self):
...
         if self.media_file:
             return os.path.basename(self.media_file.name)
+
+    @classmethod
+    def soft_delete(self, instance, user=None):
+        """
+        Soft deletes an attachment by adding a deleted_at timestamp.
+        """
+        if instance:
+            queryset = self.__class__.objects.filter(
```
DavisRayM: Did you mean to make this `cls.objects.filter(name__in=instance.get_expected_media())` ?

onadata/apps/logger/models/attachment.py
```diff
@@ -86,3 +90,16 @@ def file_hash(self):
     def filename(self):
         if self.media_file:
             return os.path.basename(self.media_file.name)
+
+    @classmethod
+    def soft_delete(self, instance, user=None):
```
DavisRayM: I personally think the first variable here should be `cls` but that up for discussion.

onadata/libs/utils/logger_tools.py
```diff
@@ -241,6 +241,11 @@ def save_attachments(xform, instance, media_files):
                 mimetype=content_type,
                 name=filename,
                 extension=extension)
+    if remove_deleted_media:
+        Attachment.soft_delete(
```
DavisRayM: Why are we passing the queryset still ? Seems like the `soft_delete` function signature changed to `(instance, user: Optional)` ?

onadata/apps/logger/models/attachment.py
pld: As is makes sense if remove `classmethod` and make this an instance method, which is better if this always requires an instance to be passed, instead operate off of the instance

onadata/apps/logger/models/attachment.py
pld: This shouldn't be a classmethod if it always requires an instance, instead it should be called off the instance w/no instance passed

DavisRayM: But in this case we would like access to multiple instances of `Attachment` right ? Or should this function be introduced into the `Instance` model as something like `soft_delete_attachments` ?

pld: ah good point! I didn't realize we're dealing w/the attachment's model, yes moving this to the `Instance` model is exactly what I was thinking

onadata/apps/logger/models/attachment.py
```diff
@@ -3,6 +3,8 @@
 from hashlib import md5
 
 from django.db import models
+from django.utils import timezone
```
pld: Codacy Issue found: [Unused timezone imported from django.utils](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4290671)

onadata/apps/logger/models/instance.py
```diff
@@ -79,6 +80,9 @@ def _get_attachments_from_instance(instance):
         attachment['id'] = a.id
         attachments.append(attachment)
```
DavisRayM: Does the `attachment` value within the `attachments` list update if we set `attachment['deleted_at']` below ?

onadata/libs/tests/utils/test_logger_tools.py
```diff
@@ -236,6 +237,87 @@ def test_attachment_tracking_for_nested_repeats(self):
...
+            BytesIO(xml_string.strip().encode('utf-8')),
+            media_files=[media_file])
+        self.assertTrue(instance.json[MEDIA_ALL_RECEIVED])
+        self.assertEquals(instance.json[TOTAL_MEDIA], 1)
+        self.assertEquals(instance.json[MEDIA_COUNT], 1)
+        self.assertEquals(instance.json[TOTAL_MEDIA], instance.total_media)
+        self.assertEquals(instance.json[MEDIA_COUNT], instance.media_count)
+        self.assertEquals(instance.json[MEDIA_ALL_RECEIVED],
+                          instance.media_all_received)
+        patch_value = 'onadata.apps.logger.models.Instance.get_expected_media'
```
DavisRayM: Why do we have to patch `get_expected_media` ?

onadata/apps/logger/models/instance.py
WinnyTroy: Sure does!
I mean, the submission does still have all the attachments it received, but at this point, only surfaces those which havent been 'deleted'.

DavisRayM: Cool 👍

onadata/libs/tests/utils/test_logger_tools.py
WinnyTroy: In an attempt to mock replacing an attachment.

When editing an instance, we receive an updated xml representaion of the instance that we then use to determine attachments for the same submission.

In this test case, since the instance already recognizes the first attachment, the updated xml would be creating a new attachment, instead of replacing the first attachment.
Patching the return value of this function allows us to use the updated xml to determine the current attachments for a specific instance.

DavisRayM: Okay, seems from the test we're passing an updated xml. Not quite sure what's different between how you're testing it and how a user would normally interact with the API that warrants the patch on the `get_expected_media` ?

_I might be missing something here_

faith-mutua:
@WinnyTroy The fix has passed QA. Deleted images are no longer being displayed on the Photo's tab nor are their links in the Table tab. Also, both audio and video files are being replaced successfully.

PR #1705, opened 2019-10-17
Add ODKToken Authentication Method

DavisRayM:
This PR adds a new Digest Authentication method name "ODKTokenAuthentication" which accepts an email and ODKToken generated on the onadata platform.

## Changes Implemented in this PR

- Add new ODK Token Model
- Add new settings `ODK_KEY_LIFETIME` and `ODK_TOKEN_LENGTH`
- Added new endpoint `/api/v1/user/odk_token` that accepts `GET` and `POST` requests
- Added new authentication class `ODKTokenAuthentication`


onadata/apps/api/viewsets/connect_viewset.py
```diff
@@ -114,6 +122,42 @@ def regenerate_auth_token(self, request, *args, **kwargs):
...
+        user = request.user
+
+        if request.method == 'GET':
+            data = {}
+            try:
+                token = ODKToken.objects.get(user=user)
+                expiry_date = token.created + timedelta(
+                    days=ODK_TOKEN_LIFETIME)
+                data.update({'odk_token': token.key})
+                data.update({'expiry_date': expiry_date})
```
ukanga: Could we have it:
```python
data = {
    'odk_token': token.key,
    'expiry_date': expiry_date,
}
```
No need to call any function but have a simple direct assignment.

onadata/apps/api/viewsets/connect_viewset.py
```diff
@@ -114,6 +122,42 @@ def regenerate_auth_token(self, request, *args, **kwargs):
...
+                return Response(data=data, status=status.HTTP_200_OK)
+            except ODKToken.DoesNotExist:
+                data.update(
```
ukanga: See my comment above

onadata/apps/api/models/odk_token.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class ODKToken(models.Model):
+    """
+    ODK Token class
+    """
+    key = models.CharField(max_length=40, primary_key=True)
+    user = models.OneToOneField(
+        AUTH_USER_MODEL, related_name='odk_token', on_delete=models.CASCADE)
+    created = models.DateTimeField(auto_now_add=True)
```
ukanga: I think we may need to have a `status` column and maintain one and only one active token per user. I am hoping it is useful for troubleshooting and auditing purposes. That way, we never really delete a token but make it inactive.

onadata/apps/api/viewsets/connect_viewset.py
DavisRayM: Changed this in the latest commits.

onadata/apps/api/viewsets/connect_viewset.py
DavisRayM: Changed in the latest commit

onadata/apps/api/models/odk_token.py
DavisRayM: @ukanga In this case will regenerating tokens not be supported ? Or will the user be able to do both ?

onadata/apps/api/storage.py
```diff
@@ -0,0 +1,64 @@
...
+def _check_odk_token_expiry(created):
+    """
+    Checks to see whether an ODK Token has expired
+    """
+    time_diff = (timezone.now() - created).total_seconds
+
+    return time_diff > ODK_KEY_LIFETIME_IN_SEC
+
+
+class DigestAccountStorage(AccountStorage):
```
ukanga: I would probably propose a more aligned naming  `ODKTokenAccountStorage`. `Digest` is too generic and is not as specific to the class referred here. Or perhaps the functionality should be adapted to handle both cases as we had discussed.

onadata/apps/api/models/odk_token.py
ukanga: Should we update the last time the user used the token? like a `last_login` field or should we use the `user.last_login` on successful authentication?

onadata/apps/api/viewsets/connect_viewset.py
```diff
@@ -114,6 +122,80 @@ def regenerate_auth_token(self, request, *args, **kwargs):
...
+        acceptable_statuses = [ODKToken.ACTIVE, ODKToken.INACTIVE]
+
+        if request.method == 'GET':
+            try:
+                token = ODKToken.objects.get(user=user)
+                expiry_date = token.created + timedelta(
+                    days=ODK_TOKEN_LIFETIME)
+                return Response(data={
+                    'enc_odk_token': token.key,
+                    'active_till': expiry_date,
```
ukanga: `active_till`  to `expires`?

onadata/apps/api/viewsets/connect_viewset.py
```diff
@@ -114,6 +122,80 @@ def regenerate_auth_token(self, request, *args, **kwargs):
...
+                    'status': token.status,
+                    'status_display': token.get_status_display()
```
ukanga: Stick to `'status': 'ACTIVE | INACTIVE'`. Or simply do not return it, there is no meaningful use of the status since we should only return an active token.

onadata/apps/api/models/odk_token.py
DavisRayM: Utilizing the `user.last_login` seems like the most ideal way in my opinion. I've implemented the change on the `DigestAuthentication` class to update the `user.last_login` on successful authentication.

onadata/apps/api/storage.py
DavisRayM: Changed in latest commits

onadata/apps/api/viewsets/connect_viewset.py
DavisRayM: Changed to `expires` in latest commits.

onadata/apps/api/viewsets/connect_viewset.py
DavisRayM: Chose to simply not return the status information. The changes have been implemented in the latest commits.

PR #1706, opened 2019-10-22
Add Open ID Connect Authentication

DavisRayM:
Add the ability to authenticate users using an Open ID Connect Provider. Currently supports WFP CIAM and Microsoft


onadata/apps/api/viewsets/open_id_connect_viewset.py
```diff
@@ -0,0 +1,213 @@
...
+            # Cache the nonce for verification on callback
+
+            if open_id_provider == 'msft':
+                scope = 'openid profile'
+                response_type = 'id_token'
+                response_mode = 'form_post'
+                cache.set(nonce, 'msft')
+            elif open_id_provider == 'wfp':
+                cache.set(nonce, 'wfp')
+
```
ukanga: Can all this come from a configuration?

```python
# local_settings.py

OPEN_ID_PROVIDER = {
   'msft': { 
        'scope': 'openid profile',
        'nonce': 'msft',
        'response_type': 'id_token',
        'response_mode': 'form_post',
    },
}
```
Then you access it as:

```python
from django.conf import settings

...

provider = getattr(settings, 'OPEN_ID_PROVIDER', {}).get('open_id_connect_provider', {})

scope = provider.get('scope', 'scope')
response_type = provider.get('response_type', 'code')
...

```

DavisRayM: Yes, i'll make this change

DavisRayM: Changes made in latest commit

PR #1707, opened 2019-10-25
Fix Multi ODK Token return on regeneration

DavisRayM:
Fix an issue where multiple ODKTokens would be returned when more that one token existed at the time of regeneration.

### DIFF
--- a/onadata/apps/api/viewsets/connect_viewset.py
+++ b/onadata/apps/api/viewsets/connect_viewset.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import MultipleObjectsReturned
 from django.utils import timezone
 from django.utils.decorators import classonlymethod
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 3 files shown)


onadata/apps/api/viewsets/connect_viewset.py
```diff
@@ -135,7 +135,8 @@ def odk_token(self, request, *args, **kwargs):
         elif request.method == 'POST':
             # Regenerates the ODK Token if one is already existant
             try:
-                old_token = ODKToken.objects.get(user=user)
+                old_token = ODKToken.objects.get(
+                    user=user, status=ODKToken.ACTIVE)
```
ukanga: We do not expect it but there are chances where we may have multiple active tokens. In this scenario, we should make any active token inactive, i.e.:
```python
ODTOken.objects.filter(user=user, status=ODKToken.ACTIVE).update(status=ODKToken.INACTIVE)
```

DavisRayM: Added this in the latest commits. We know deactivate all tokens when two or more tokens are active then generate and return a new token.

PR #1708, opened 2019-10-27
Add ability to encrypt forms after creation

DavisRayM:
This PR implements the ability to encrypt forms after they have being created through a `PATCH` request with a payload like:

```
{
    "public_key": "MII...."
}
```

- [x] Add ability to encrypt forms on `PATCH`
- [x] Write tests

## Sample Payload

```
{
    "public_key": "some_rsa_key"
}
```


onadata/apps/logger/models/xform.py
```diff
@@ -822,6 +825,17 @@ def _set_encrypted_field(self):
             else:
                 self.encrypted = False
 
+    def _set_public_key_field(self):
+        if self.json and self.json != '':
+            if self.submission_count() == 0:
+                json_dict = json.loads(self.json)
+                json_dict['public_key'] = self.public_key
+                json_dict['submission_url'] = ODK_SUBMISSION_URL
```
ukanga: Submission URL is no longer a requirement for encrypted forms, I don't think we should enforce this here. We should remove it.

onadata/settings/travis_test.py
```diff
@@ -67,4 +67,5 @@
 VERIFIED_KEY_TEXT = 'ALREADY_ACTIVATED'
 
 ODK_TOKEN_FERNET_KEY = 'ROsB4T8s1rCJskAdgpTQEKfH2x2K_EX_YBi3UFyoYng='
+ODK_SUBMISSION_URL = ''
```
ukanga: This is not necessary, let us remove it.

DavisRayM: Removed the `ODK_SUBMISSION_URL` setting in latest commits

onadata/apps/logger/models/xform.py
DavisRayM: Removed the `ODK_SUBMISSION_URL` in latest commits

PR #1733, opened 2019-11-29
Add a way to deactivate organizations by default on create 

DavisRayM:
Add optional `ORG_ON_CREATE_IS_ACTIVE` setting

### DIFF
--- a/onadata/apps/api/tools.py
+++ b/onadata/apps/api/tools.py
@@ -121,7 +121,10 @@ def create_organization_object(org_name, creator, attrs=None):
         first_name=first_name,
         last_name=last_name,
         email=email,
-        is_active=True)
+        is_active=getattr(
+            settings,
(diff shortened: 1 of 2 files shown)


onadata/apps/api/tools.py
```diff
@@ -121,7 +121,10 @@ def create_organization_object(org_name, creator, attrs=None):
         first_name=first_name,
         last_name=last_name,
         email=email,
-        is_active=True)
+        is_active=getattr(
+            settings,
+            'ORGANIZATION_ACTIVE_STATUS_ON_CREATE',
```
ukanga: I would recommend `ORG_ON_CREATE_IS_ACTIVE`.

DavisRayM: Alright let me change the naming.

DavisRayM: Done

PR #1734, opened 2019-12-02
Return owner team members on the Organization endpoint

DavisRayM:
## Changes

- Return members of the 'Owners' team on the `/orgs` endpoint under the `users` key.

### DIFF
--- a/onadata/libs/serializers/organization_serializer.py
+++ b/onadata/libs/serializers/organization_serializer.py
@@ -13,7 +13,8 @@ from rest_framework import serializers
 from onadata.apps.api import tools
 from onadata.apps.api.models import OrganizationProfile
 from onadata.apps.api.tools import (_get_first_last_names,
-                                    get_organization_members)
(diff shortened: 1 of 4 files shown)


onadata/libs/serializers/organization_serializer.py
```diff
@@ -113,12 +114,24 @@ def get_users(self, obj):  # pylint: disable=no-self-use
...
+            } for u in user_list]
+
         members = get_organization_members(obj) if obj else []
+        owners = get_organization_owners(obj) if obj else []
+
+        repeated_users = set(owners) & set(members)
+
+        for user in repeated_users:
+            members.exclude(username=user.username)
+
```
ukanga: I think this can be rewritten to 
```python
members.exclude(username__in=[user.username for user in repeated_users])
```

DavisRayM: Changed in latest commits

Issue #1732, opened 2019-11-29
Do not return projects tied to inactive organizations


DavisRayM:
We currently return `projects` tied to an inactive `user` this shouldn't be the case.

PR #1755, opened 2020-01-03
Change queryset in project viewset in order to return only projects associated to active organizations.

FrankApiyo:

Added a filter to the queryset in ProjectViewset to ensure that we do not return projects associated with inactive organizations

Ran onadata locally and ensured that only projects associated with active organizations are gotten from the projects endpoint

Fixes: #1732

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_project_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_project_viewset.py
@@ -142,7 +142,12 @@ class TestProjectViewSet(TestAbstractViewSet):
         request.user = self.user
         response = self.view(request)
         self.assertEqual(len(response.data), 2)
-        self.assertEqual(response.data[1].get('projectid'), shared_project.id)
+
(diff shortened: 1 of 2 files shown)


onadata/apps/api/tests/viewsets/test_project_viewset.py
```diff
@@ -106,6 +106,46 @@ def test_projects_list(self):
...
+        alice_data = {'username': 'alice', 'email': 'alice@localhost.com'}
+        alice_profile = self._create_user_profile(alice_data)
+        alice_user = alice_profile.user
+        shared_project = Project(name='demo2',
+                                 shared=True,
+                                 metadata=json.dumps({'description': ''}),
+                                 created_by=alice_user,
+                                 organization=alice_user)
+        alice_profile.save()
+        alice_user.save()
```
ivermac: Why are we calling `.save()` on `alice_profile` and `alice_user` objects?

PR #1757, opened 2020-01-07
Assign Owner permission to creator for organization account UserProfile

### LINKED ISSUE #1754: Assign OwnerRole explicitly to and org creator
### Environmental Information

- Operating System: all
- Onadata version: all
- Python version: 3

### Problem description

Not all OwnerRole permissions are assigned to the creator of an organization.

### Expected behaviour

Permissions like add a form to an organization should be assigned to org owners who create the organization.

Create an organization and then try publishing a form using ODK Briefcase. You should get a 403 error message on the`/ formUpload` endpoint.

DavisRayM:

- Add `OwnerRole` permissions to creator for `OrganizationalProfile.userprofile_ptr`
- Remove `Role` permissions from a user when they are removed from an organization

### Pending work on PR

- [x] Make sure users added into the organization have the correct permissions
- [x] Make sure users permissions are removed when the user is removed from the organization

- Added a test to verify that the change works

- In order for these changes to take effect for all existing organizations the `set_api_permissions` management command needs to be run

### Additional Information

Issue #1754 is being caused because the `User` object returned through `User.objects.get` and `get_object_or_404` links to a `UserProfile` and not a `OrganizationProfile` when `user.profile` is called. The user / creator in this case normally doesn't have the OwnerRole / `can_add_xform` attached to the `UserProfile` / `org.userprofile_ptr`.

The `UserProfile` is present due to how django handles model inheritances(more info on that [here](https://docs.djangoproject.com/en/2.2/topics/db/models/#multi-table-inheritance)), as such it seems for some reason the `User` object defaults to linking to `org_profile.userprofile_ptr`

Closes #1754

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
@@ -3,7 +3,6 @@ from mock import patch
 from builtins import str as text
 
 from django.contrib.auth.models import User
-from django.shortcuts import get_object_or_404
 
 from onadata.apps.api.tests.viewsets.test_abstract_viewset import\
     TestAbstractViewSet
@@ -1030,7 +1029,7 @@ class TestOrganizationProfileViewSet(TestAbstractViewSet):
         }
         self.assertIn(expected_user, response.data[0]['users'])
 
-    def test_creator_and_user_permissions(self):
+    def test_creator_permissions(self):
         """
         Test that the creator of the organization has the necessary
         permissions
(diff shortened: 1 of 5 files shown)


DavisRayM:
This PR (as of commit [60affe6](https://github.com/onaio/onadata/pull/1757/commits/60affe64d42c1aa6184a76d9b250ae9779e25a39)) contains one approach on how to solve the Issue.

The approach: Assign Permissions to users for the `org.userprofile_ptr`

onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
```diff
@@ -1028,3 +1028,58 @@ def test_creator_in_users(self):
...
+        """
+        self._org_create()
+        request = self.factory.get('/', **self.extra)
+        response = self.view(request)
+        self.assertNotEqual(response.get('Cache-Control'), None)
+        self.assertEqual(response.status_code, 200)
+
+        orgs = OrganizationProfile.objects.filter(creator=self.user)
+        self.assertEqual(orgs.count(), 1)
+        org = OrganizationProfile.objects.filter(creator=self.user).first()
```
ivermac: Would the following work since `orgs` is already defined in line 1043?
```python
...
org = orgs.first()
```

DavisRayM: Yes it would, changed this.

PR #1760, opened 2020-01-08
use_master when fetching external export metadata information

### LINKED ISSUE #1758: module 'onadata.apps.main.models.meta_data' has no attribute 'DoesNotExist'
``` code
[2020-01-07 06:38:15,791: ERROR/ForkPoolWorker-1] Task onadata.apps.viewer.tasks.create_external_export[2e3d52cf-02e1-418f-a501-bd35d2e07524] raised unexpected: DoesNotExist('MetaData matching query does not exist.',)
Traceback (most recent call last):
  File "/home/onadata/.virtualenvs/onadata/lib/python3.6/site-packages/celery/app/trace.py", line 382, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/onadata/.virtualenvs/onadata/lib/python3.6/site-packages/celery/app/trace.py", line 641, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/onadata/app-versioned/1575642939/onadata/apps/viewer/tasks.py", line 364, in create_external_export
    xform=export.xform)
  File "/home/onadata/app-versioned/1575642939/onadata/libs/utils/export_tools.py", line 747, in generate_external_export
    server, name = _get_server_from_metadata(xform, meta, token)
  File "/home/onadata/app-versioned/1575642939/onadata/libs/utils/export_tools.py", line 704, in _get_server_from_metadata
    result = report_templates.get(pk=meta)
  File "/home/onadata/.virtualenvs/onadata/lib/python3.6/site-packages/django/db/models/query.py", line 399, in get
    self.model._meta.object_name
onadata.apps.main.models.meta_data.MetaData.DoesNotExist: MetaData matching query does not exist
```

WinnyTroy:
Metadata details for extenal exports are being fetched from the replica database. This may fail if the record may still not be available in read replica.

Fixes #1758

### DIFF
--- a/onadata/apps/viewer/tests/test_exports.py
+++ b/onadata/apps/viewer/tests/test_exports.py
@@ -1285,12 +1285,6 @@ class TestExports(TestBase):
         self.assertEqual(response.status_code, 302)
         self.assertEqual(Export.objects.count(), num_exports + 1)
 
-    def test_external_export_metadata_retrieved_from_master_db(self):
-        self._publish_transportation_form()
-        self._submit_transport_instance()
-        meta = MetaData.external_export(self.xform)
-        self.assertEqual(meta.db, "default")
-
     @patch('onadata.apps.viewer.tasks.get_object_or_404')
     def test_create_external_export_url_with_non_existing_export_id(
             self, mock_404):
--- a/onadata/apps/viewer/tests/test_exports.py
+++ b/onadata/apps/viewer/tests/test_exports.py
@@ -1285,6 +1285,12 @@ class TestExports(TestBase):
         self.assertEqual(response.status_code, 302)
         self.assertEqual(Export.objects.count(), num_exports + 1)
 
+    def test_external_export_metadata_retrieved_from_master_db(self):
+        self._publish_transportation_form()
+        self._submit_transport_instance()
+        meta = MetaData.external_export(self.xform)
+        self.assertEqual(meta.db, "default")
+
     @patch('onadata.apps.viewer.tasks.get_object_or_404')
     def test_create_external_export_url_with_non_existing_export_id(
             self, mock_404):
--- a/onadata/libs/utils/export_tools.py
+++ b/onadata/libs/utils/export_tools.py
@@ -692,7 +692,10 @@ def clean_keys_of_slashes(record):
 
 
 def _get_server_from_metadata(xform, meta, token):
-    report_templates = MetaData.external_export(xform)
+    from multidb.pinning import use_master
+    # Fetch metadata details from master directly
+    with use_master:
+        report_templates = MetaData.external_export(xform)
 
     if meta:
         try:
--- a/onadata/libs/utils/export_tools.py
+++ b/onadata/libs/utils/export_tools.py
@@ -694,8 +694,11 @@ def clean_keys_of_slashes(record):
 def _get_server_from_metadata(xform, meta, token):
     from multidb.pinning import use_master
     # Fetch metadata details from master directly
-    with use_master:
+    try:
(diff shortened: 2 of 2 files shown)


onadata/libs/utils/export_tools.py
```diff
@@ -692,7 +692,10 @@ def clean_keys_of_slashes(record):
 
 
 def _get_server_from_metadata(xform, meta, token):
-    report_templates = MetaData.external_export(xform)
+    from multidb.pinning import use_master
+    # Fetch metadata details from master directly
+    with use_master:
+        report_templates = MetaData.external_export(xform)
```
ukanga: In other parts of the codebase, we have taken the approach of using master only when the initial request has failed. I would see something like:
```python
try:
    report_templates = MetaData.external_export(xform)
except MetaData.DoesNotExist:
   with use_master:
        report_templates = MetaData.external_export(xform)
```

ivermac:
Found the following as well but I forgot to post
https://stackoverflow.com/questions/13358939/python-mock-side-effect-or-return-value-dependent-on-call-count

onadata/apps/viewer/tests/test_exports.py
```diff
@@ -1285,6 +1285,37 @@ def test_create_external_export_url(self):
...
+            with self.assertRaises(MetaData.DoesNotExist):
+                mock.side_effect = _side_effect()
+                mock(self.xform, data_value)
+
+            def __second_side_effect(*args):
+                """
+                Change the side effect so that on the second round  we do not
+                raise a DoesNotExist exception
+                """
+                return None
```
DavisRayM, line 1314: I'm curious why did we have to do this ? To test out it retries ? Is there an arg we can check or variable to make sure this is actually the master db it uses on the second time ?

onadata/libs/utils/export_tools.py
```diff
@@ -692,7 +692,13 @@ def clean_keys_of_slashes(record):
 
 
 def _get_server_from_metadata(xform, meta, token):
-    report_templates = MetaData.external_export(xform)
+    from multidb.pinning import use_master
```
DavisRayM, line 695:
```suggestion
```

onadata/libs/utils/export_tools.py
```diff
@@ -692,7 +692,13 @@ def clean_keys_of_slashes(record):
...
+    # Fetch metadata details from master directly
+    try:
+        report_templates = MetaData.external_export(xform)
+    except MetaData.DoesNotExist:
+        with use_master:
```
DavisRayM, line 700:
```suggestion
        from multidb.pinning import use_master
        with use_master:
```

onadata/apps/viewer/tests/test_exports.py
WinnyTroy: By default, since this is a test case, it would be using master?

onadata/libs/utils/export_tools.py
WinnyTroy: Sure, I'll move this in here

onadata/apps/viewer/tests/test_exports.py
```diff
@@ -1285,6 +1285,37 @@ def test_create_external_export_url(self):
...
+
+            mock.side_effect = __second_side_effect
+            mock(self.xform, data_value)
```
DavisRayM, line 1317: When will this fail ?

DavisRayM: I don't think this should be included. I don't see how this would be tested unless we can assert the context the function is run in the second time.

PR #1765, opened 2020-01-10
Handle error thrown by urllib

### LINKED ISSUE #1759: Check submitted file attributes before processing
### Environmental Information

- Operating System: ubuntu, macos
- Onadata version: v2.1.0
- Python version: 3.6

### Problem description
There are times publishing a form via a url/file triggers unwanted/invalid responses

### Expected behavior
Before processing a submitted link or file, ensure the following checks are successful:

- [ ]  if the file cannot be read, or its Unix file type is undetermined, the file program will indicate that the file was processed but its type was undetermined.
- [ ]  file must be able to determine the types directory, FIFO, socket, block special file, and character special file
- [ ] zero-length files are identified as invalid
- [ ] an initial part of file is considered and file is to use position-sensitive tests
- [ ] the entire file is considered and file is to use context-sensitive tests
- [ ] the file is identified as a data file

Try publishing a form using an invalid url 

### Additional Information

DavisRayM:

- Handle `BadStatusLine` error raised when _"a server responds with a HTTP status code that we don’t understand."_

- [x] Tested on a local server of onadata
- [x] Added a test `test_error_raised_xform_url_upload_urllib_error`

*N/A*

Closes #1759

### DIFF
--- a/onadata/libs/utils/logger_tools.py
+++ b/onadata/libs/utils/logger_tools.py
@@ -24,6 +24,7 @@ from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from django.utils.encoding import DjangoUnicodeDecodeError
 from django.utils.translation import ugettext as _
+from http.client import BadStatusLine
 from modilabs.utils.subprocess_timeout import ProcessTimedOut
 from multidb.pinning import use_master
 from pyxform.errors import PyXFormError
@@ -495,7 +496,7 @@ def publish_form(callback):
             'type': 'alert-error',
             'text': _(u'Form validation timeout, please try again.'),
         }
-    except (MemoryError, OSError):
+    except (MemoryError, OSError, BadStatusLine):
         return {
             'type': 'alert-error',
             'text': _((u'An error occurred while publishing the form. '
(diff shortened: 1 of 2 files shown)


DavisRayM:
> Looks good. Can we add a test for it by mocking the function that triggers `BadStatusLine`? We have an example of how to mock an exception by using a `side_effect` as seen below:
> 
> https://github.com/onaio/onadata/blob/fdd91d7024c9b15ee1eee4e074e2e0ef21939b8b/onadata/apps/api/tests/viewsets/test_charts_viewset.py#L104-L113

Added.

onadata/libs/utils/logger_tools.py
```diff
@@ -501,6 +502,12 @@ def publish_form(callback):
             'text': _((u'An error occurred while publishing the form. '
                        'Please try again.')),
         }
+    except BadStatusLine:
+        # Catch urllib error
+        return {
+            'type': 'alert-error',
+            'text': _(u'An error occured while trying to retrieve the form.')
+        }
```
ukanga: What would the action that would be required by the user to take if they encounter this error?
Would the error message in https://github.com/onaio/onadata/pull/1765/files#diff-018f1a90f318bf1748393be886fee2c0R502-R503 be sufficient in this case? If it would be sufficient could we add `BadStatusLine` to `except (BadStatusLine, MemoryError, OSError)`?

DavisRayM: The error message would be perfect for this scenario. I'll update my PR.

WNjihia:
Testing checks out @DavisRayM

PR #1766, opened 2020-01-13
Add X-Frame-Options Middleware class.

### LINKED ISSUE #1767: Add X-Frame-Options Middleware class
### Environmental Information

- Operating System: All
- Onadata version:
- Python version: 3.6

### Problem description
Responses are missing `x-frame-options` header which prevent the page from being loaded on a server whose domain is different to the one where onadata instance is being hosted.

### Expected behavior
The response object should have `x-frame-options` header

Try opening a page on an iframe on a locally running app whose address is like `http://localhost:8007`

### Additional Information

WinnyTroy:
Set the x-frame-options header. Guide from django's documentation [here](https://docs.djangoproject.com/en/3.0/ref/clickjacking/#setting-x-frame-options-for-all-responses)
Fixes #1767

### DIFF
--- a/onadata/settings/common.py
+++ b/onadata/settings/common.py
@@ -169,7 +169,10 @@ MIDDLEWARE = (
     'django.contrib.messages.middleware.MessageMiddleware',
     'onadata.libs.utils.middleware.HTTPResponseNotAllowedMiddleware',
     'onadata.libs.utils.middleware.OperationalErrorMiddleware',
+    'django.middleware.clickjacking.XFrameOptionsMiddleware',
 )
+
+X_FRAME_OPTIONS = 'SAMEORIGIN'
 
 LOCALE_PATHS = (os.path.join(PROJECT_ROOT, 'onadata.apps.main', 'locale'), )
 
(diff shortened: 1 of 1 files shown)


onadata/settings/common.py
```diff
@@ -169,8 +169,11 @@
     'django.contrib.messages.middleware.MessageMiddleware',
     'onadata.libs.utils.middleware.HTTPResponseNotAllowedMiddleware',
     'onadata.libs.utils.middleware.OperationalErrorMiddleware',
+    'django.middleware.clickjacking.XFrameOptionsMiddleware',
 )
 
+X_FRAME_OPTIONS = 'SAMEORIGIN'
```
moshthepitt: I think that we should make the default "DENY" as recommended in Django 3.0 - see https://docs.djangoproject.com/en/3.0/ref/clickjacking/

ivermac: Fair point

WNjihia:
Tested this and it checks out. <onadata-api-domain>/accounts/login doesn't load on the iframe
cc: @ivermac @WinnyTroy

PR #1770, opened 2020-01-21
Upgrade to Django 2.2

### LINKED ISSUE #1769: Upgrade Django version
### Environmental Information

- Operating System: All
- Onadata version: Master
- Python version: All

### Problem description

Onadata is currently running on Django 2.1 which is no longer being supported as far as I know. In order to keep up to date with all the security and maintenance patches we should update our version of Django to the latest LTS version

### Expected behavior

- N/A

- N/A

### Additional Information

- N/A

DavisRayM:

- Upgraded the Django version to 2.2

- Tested locally
- Tested on travis

- Full QA may need to be done. In case something bypassed our tests.

Closes #1769

### DIFF
--- a/onadata/apps/api/tests/views/test_user_permissions.py
+++ b/onadata/apps/api/tests/views/test_user_permissions.py
@@ -83,6 +83,11 @@ class TestUserPermissions(TestAbstractViewSet):
 
         self.assertFalse(self.xform.shared)
 
+        # Remove key:value pairs where the value is None.
+        # More info: https://code.djangoproject.com/ticket/30024
+        data.pop('enketo_preview_url')
+        data.pop('last_submission_time')
+
         request = self.factory.put('/', data=data, **self.extra)
         response = view(request, pk=self.xform.id)
         self.assertEqual(response.status_code, 404)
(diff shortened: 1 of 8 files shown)


onadata/apps/api/tests/views/test_user_permissions.py
```diff
@@ -83,6 +83,9 @@ def test_manager_can_update_xform(self):
 
         self.assertFalse(self.xform.shared)
 
+        data.pop('enketo_preview_url')
```
DavisRayM: This change is related to this issue [here](https://code.djangoproject.com/ticket/30024)

onadata/libs/serializers/password_reset_serializer.py
```diff
@@ -35,14 +35,13 @@ def get_password_reset_email(user, reset_url,
     """Creates the subject and email body for password reset email."""
     result = urlparse(reset_url)
     site_name = domain = result.hostname
-    encoded_username = urlsafe_base64_encode(
-        b(user.username.encode('utf-8'))).decode('utf-8')
+    encoded_username = urlsafe_base64_encode(b(user.username.encode('utf-8')))
```
DavisRayM: Change this because `urlsafe_base64_encode` now returns a string instead of a bytestring.

onadata/apps/api/tests/views/test_user_permissions.py
ivermac: This makes sense. Do we add the link - https://code.djangoproject.com/ticket/30024 - as a comment just above the `<obj>.pop('<key>')` so that in future, we can easily know why we had to do this?

DavisRayM: Yes, I'll add a link later on in case someone works on the code later on.

DavisRayM:
Using the `QA-` label to signify this PR needs to go through QA.

PR #1773, opened 2020-01-29
Return flow data results for merged datasets

### LINKED ISSUE #1643: Flow data results in a merged dataset is blank
The flow data results endpoint returns blank for a merged dataset form- https://api.ona.io/api/v1/flow-results/packages/{UUID}/responses whereas the data endpoint - https://api.ona.io/api/v1/data/{pk}.json - would return data.

There should be data returned from the flow data results endpoint.

DavisRayM:

- Set uuid for `MergedXForm` on save
- Return flow data results for merged datasets
- Add test `test_retrieve_responses_merged_dataset`
- Enable one to retrieve a FLOIP Package or response using the complete `uuid4` string or its hex version
- Add migration to set default uuid for XForms that are missing the field.
- Return complete uuid4 strings on the floip viewset
- Bump pyfloip version

- Added test to ensure data is returned
- Modified tests to ensure UUID is returned from the Merged XForm Endpoint

## QA Checklist

- [x] Able to retrieve a MergedXForm and it's data from the FLOIP Endpoint.
> Step 1: Retrieve MergedXForm `UUID` from the XForm endpoint `/api/v1/forms/<form_id>`
> Step 2: Access the FLOIP Package `http://localhost:8000/api/v1/flow-results/packages/<uuid>`. _Should return package information for the MergedXForm_
> Step 3: Access the FLOIP Responses `http://localhost:8000/api/v1/flow-results/packages/<uuid>/responses`. _Should return all the answers to the form questions as JSON Objects_
> Things to consider: Forms with media, Forms with nested groups

- [x] `uuid4` strings are used to identify packages(_All ids returned by the endpoint should be uuid4_). _A uuid4 string can be identified by its 32 hex characters and 4 dashes_

Closes #1643


DavisRayM:
### Side effects of implementing the PR as is

- Need to find a way to set `uuid` for all pre-existing MergedXForms.

> The pre-existing `MergedXForms` defaulted to an empty string(Inherits the field from [XForm](https://github.com/onaio/onadata/blob/master/onadata/apps/logger/models/xform.py#L721)) as the `uuid` we may need to set the `uuid` fields for these xforms somehow.

Screenshot 2020-01-29 at 13 10 20

onadata/apps/api/tests/viewsets/test_floip_viewset.py
```diff
@@ -196,6 +207,50 @@ def test_publish_number_question_names(self):  # pylint: disable=C0103
...
+            content_type='application/vnd.api+json', **self.extra)
+        response = view(request, uuid=dataset_uuid)
+        self.assertEqual(response.status_code, 200)
+
+        # Convert the returned generator object into a list
+        response.data['attributes']['responses'] = list(
+            response.data['attributes']['responses'])
+        # The transportation form(self.xform) contains 11 responses
+        # Assert that the responses are returned
+        self.assertEqual(len(response.data['attributes']['responses']), 11)
```
DavisRayM: Chose to use the number of responses returned(11) for this test due to erratic time values returned in the responses.

DavisRayM:
> ### Side effects of implementing the PR as is
> 
>     * Need to find a way to set `uuid` for all pre-existing MergedXForms. _Non breaking issue_
> 
> 
> > The pre-existing `MergedXForms` defaulted to an empty string(Inherits the field from [XForm](https://github.com/onaio/onadata/blob/master/onadata/apps/logger/models/xform.py#L721)) as the `uuid` we may need to set the `uuid` fields for these xforms somehow.
> 
> Screenshot 2020-01-29 at 13 10 20

Using a data migration to solve the missing `uuid` field value may be a bit painful for production servers since the XForm model it inherits from is usually quite bulky.

Another option we have is to set the values lazily either:

- When a merged dataset is retrieved by the forms endpoint
- When `.save()` is called on a dataset

CC: @pld @ivermac

onadata/libs/serializers/merged_xform_serializer.py
```diff
@@ -198,6 +199,7 @@ def create(self, validated_data):
         validated_data['instances_with_geopoints'] = any([
             __.instances_with_geopoints for __ in validated_data.get('xforms')
         ])
+        validated_data['uuid'] = uuid.uuid4().hex
```
ukanga: Could we use the same uuid4 used in https://github.com/onaio/onadata/pull/1773/files#diff-ab8d7875ff4cf8338ac5076d56ff3d47R183? Only that we now use the full hex values rather than the first 6 values?

DavisRayM: The link seems broken ? Not able to pinpoint the exact line you're referencing

ukanga:
> Using a data migration to solve the missing `uuid` field value may be a bit painful for production servers since the XForm model it inherits from is usually quite bulky.
> 
> Other options we have is to set the values lazily either:
> 
> * When a merged dataset is retrieved by the forms endpoint
> * When `.save()` is called on a dataset
> 
> Opinions on a favorable way to proceed ?

I will recommend the data migration as well as the default value for the UUID field. This will be the data migration first and a second migration that sets the default value.

onadata/libs/serializers/merged_xform_serializer.py
DavisRayM: Never mind I think the line linked was this ? I'll make the change

```
https://github.com/onaio/onadata/blob/200aefe4a0c8a13f09d6c8e62b79ccb489b55602/onadata/libs/serializers/merged_xform_serializer.py#L182-L183
```

onadata/apps/logger/tests/models/test_xform.py
```diff
@@ -167,6 +168,26 @@ def test_check_xform_uuid(self):
...
+        Objects
+        """
+        self._publish_transportation_form()
+        uuid = self.xform.uuid
+        # The _publish_transportation_form uses the DataDictionary
+        # Model which has some processing steps that ensure XForms
+        # Always have uuids.
+        # Assert that set_uuid is called with an object
+        # that already has a uuid.
+        self.assertEqual(set_uuid_mock.call_count, 1)
```
DavisRayM: Seems like the `DataDictionary` model always sets uuid for XForms. Sadly, since the `MergedXForm` model inherits from the `XForm` model directly it doesn't have the uuid set...

_I wonder if we should just modify the `save` method of the MergedXForm model to set the uuid ?_

DavisRayM:
> can you please add a test for the generation of default IDs

I've added a test to test the generation of the default uuids in the latest commits. Please have a look.

onadata/apps/logger/tests/models/test_xform.py
```diff
@@ -167,6 +168,26 @@ def test_check_xform_uuid(self):
...
+        self.assertEqual(set_uuid_mock.call_args[0][0].uuid, uuid)
+        # Assert that the uuid is not empty and is
+        # 32 characters in length
+        self.assertNotEqual(uuid, '')
+        self.assertEqual(len(uuid), 32)
```
DavisRayM: uuid is always 36 characters (32 hex digits + 4 dashes). Using the default way uuids are set for `DataDictionary` objects; which only returns the hex values.

faith-mutua:
@DavisRayM this has passed QA. All the answers to the questions have their own object.

onadata/apps/logger/tests/models/test_xform.py
ukanga: Yes, let's have a UUID for the merged form.

onadata/libs/serializers/floip_serializer.py
```diff
@@ -78,14 +79,34 @@ def parse_responses(responses, session_id_index=SESSION_ID_INDEX,
...
 
+class UUIDField(serializers.CharField):
+    """
+    Custom UUIDField that stores the values of a UUID as hex but
+    represents the as a complete UUID4 string
+    """
+    def to_representation(self, obj):
+        return uuid.UUID(obj)
+
+    def to_internal_value(self, obj):
```
pld, line 90: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/libs/serializers/floip_serializer.py
```diff
@@ -78,14 +79,34 @@ def parse_responses(responses, session_id_index=SESSION_ID_INDEX,
...
+        return uuid.UUID(obj).hex
+
+
+class ReadOnlyUUIDField(serializers.ReadOnlyField):
+    """
+    Custom ReadOnlyField for UUID
+    """
+    def to_representation(self, obj):
```
pld, line 98: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/libs/serializers/floip_serializer.py
```diff
@@ -78,14 +79,38 @@ def parse_responses(responses, session_id_index=SESSION_ID_INDEX,
...
 
+class UUIDField(serializers.CharField):
+    """
+    Custom UUIDField that stores the values of a UUID as hex but
+    represents the as a complete UUID4 string
+    """
+    def to_representation(self, obj):
+        return str(uuid.UUID(obj))
+
+    def to_internal_value(self, obj):
```
pld, line 90: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/libs/serializers/floip_serializer.py
```diff
@@ -78,14 +79,38 @@ def parse_responses(responses, session_id_index=SESSION_ID_INDEX,
...
+            return val
+        except TypeError:
+            raise serializers.ValidationError('Invalid uuid')
+
+
+class ReadOnlyUUIDField(serializers.ReadOnlyField):
+    """
+    Custom ReadOnlyField for UUID
+    """
+    def to_representation(self, obj):
```
pld, line 102: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/apps/api/viewsets/floip_viewset.py
```diff
@@ -2,7 +2,10 @@
 """
 FloipViewSet: API endpoint for /api/floip
 """
+import uuid as uu
 
+from django.db.models import Q
```
pld, line 7: Codacy Issue found: [Unused Q imported from django.db.models](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/libs/serializers/floip_serializer.py
pld, line 90: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/libs/serializers/floip_serializer.py
pld, line 102: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/libs/serializers/floip_serializer.py
pld, line 90: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

onadata/libs/serializers/floip_serializer.py
pld, line 102: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4906603)

faith-mutua:
@DavisRayM Flow data results for merged datasets now have data.

onadata/apps/logger/migrations/0061_auto_20200713_0814.py
```diff
@@ -0,0 +1,26 @@
...
+
+def generate_uuid_if_missing(apps, schema_editor):
+    """
+    Generate uuids for XForms without them
+    """
+    XForm = apps.get_model('logger', 'XForm')
+
+    for form in XForm.objects.filter(uuid=''):
+        form.uuid = get_uuid()
+        form.save()
```
ivermac, line 16: Won't this be expensive when running migrations - even though it might be one time? Can we do this in the XForm's `save` function; check if that form has a uuid and set one if it doesn't exist.
I guess my worry is just the app going through all XForms checking which one doesn't have a uuid

ivermac: May be not check on the XForm's instance save function but where the uuid is required

DavisRayM: I'm not entirely sure if this query would be expensive to run(_I'll try analyzing the query on a large database and get back to you_). The `XForm` save function has a check to set the UUID when missing through the `DataDictionary` model.

The main reason this was included was because of `MergedXForms` since they're not created through the `DataDictionary` model. But, I've also modified the `MergedXForm` save to set it when missing...

_I'm not sure at the moment about when `MergedXForms` are updated/saved again after creation_

ivermac: Cool. I guess i'm fine if retrieving xforms without uuid won't be expensive

DavisRayM: The query isn't expensive to run. _The index we currently have on the UUID field speeds up the query quite a lot_

Screenshot 2020-07-15 at 15 26 42

ivermac: Cool

onadata/apps/logger/migrations/0061_auto_20200713_0814.py
```diff
@@ -0,0 +1,26 @@
...
+from onadata.libs.utils.common_tools import get_uuid
+
+
+def generate_uuid_if_missing(apps, schema_editor):
+    """
+    Generate uuids for XForms without them
+    """
+    XForm = apps.get_model('logger', 'XForm')
+
+    for form in XForm.objects.filter(uuid=''):
```
ukanga, line 14: We tend to use `xform` instead of form as a variable of an `XForm` model object. Can we keep the same for consistency?

DavisRayM: Updated this to `xform` in the latest commits

PR #1775, opened 2020-01-31
Create consistent Enketo URLs

### LINKED ISSUE #1503: Create consistent Enketo URLs
### Actual behaviour

When a form is transferred to a different user account, the Enketo URI changes. 

### Expected behaviour

Retain the same Enketo URI across different user accounts.

### Implementation plan

Setup a formList URI endpoint that is linked to only the form and not the user account. e.g /e/[UUID]/[formList|submission|etc]

DavisRayM:

- Add new endpoints: `/enketo/<form_pk>/formList` and `/enketo/<form_pk>/submission`
- Update and add tests where necessary
- Update `get_form_url` function
- Update `get_enketo_preview` function
- Updated `create_enketo_urls` and `update_enketo_urls` management commands

- Added and updated tests

### Pending verification steps

- [x] Test flow locally
- [x] Pass QA
    - [x] Test one is able to create a form with the new enketo url set. One would need to check the Enketo store to confirm the `enketo_url` is set to `/e/<xform_id`
    - [x] Test one is able to make / edit submissions to forms with the new endpoint (Best way to do this would be to use a newly created XForm as we don't change the `enketo_url` for pre-existing XForms unless they're moved)
    - [x] ~Test that moving a form changes its `enketo_url` to the new endpoint `/e/<xform_id`~

- The url enketo we'll use to access form will have to change to the new endpoint. This can easily be done using the `update_enketo_urls` function

### Notes

The migration run in this PR creates an Index with the `xxx_pattern_ops` operator classes. These operator classes cannot utilize `<, <=, >, or >=` queries. I don't think we'll be running any of these queries on the `uuid` field at the moment... but it's subject to change.

Quote from [the postgresql docs](https://www.postgresql.org/docs/9.2/indexes-opclass.html):

_"Note that you should also create an index with the default operator class if you want queries involving ordinary <, <=, >, or >= comparisons to use an index. Such queries cannot use the xxx_pattern_ops operator classes. (Ordinary equality comparisons can use these operator classes, however.) It is possible to create multiple indexes on the same column with different operator classes."_

Closes #1503


onadata/libs/utils/viewer_tools.py
```diff
@@ -315,6 +316,8 @@ def get_form_url(request,
     if preview:
         url = '%s/preview' % url
 
+    if xform_uuid:
+        url = f"{url}/e/{xform_uuid}"
```
pld: Codacy Issue found: [invalid syntax](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4921990)

onadata/apps/main/management/commands/update_enketo_urls.py
```diff
@@ -45,25 +48,45 @@ def handle(self, *args, **options):
...
+
+                    if data_type == 'enketo_url':
+                        form_url = get_form_url(
+                            request, username, protocol=protocol,
+                            xform_pk=xform.pk)
+                        _enketo_url = enketo_url(form_url, id_string)
+                        MetaData.enketo_url(xform, _enketo_url)
+                    elif data_type == 'enketo_preview_url':
+                        _enketo_preview_url = get_enketo_preview_url(
+                            request, username, id_string, xform_pk=xform.pk)
```
pld: I don't think this matches the new signature, which takes `id_string` 2nd then an optional username

onadata/apps/main/management/commands/update_enketo_urls.py
```diff
@@ -45,25 +48,45 @@ def handle(self, *args, **options):
...
-                        xform_pk=xform.pk)
-                    _enketo_url = enketo_url(form_url, id_string)
-                    MetaData.enketo_url(xform, _enketo_url)
-                elif data_type == 'enketo_preview_url':
-                    _enketo_preview_url = get_enketo_preview_url(
-                        request, username, id_string, xform_pk=xform.pk)
-                    MetaData.enketo_preview_url(xform, _enketo_preview_url)
-                f.write('%s : %s \n' % (id_string, data_value))
+
+        if use_username:
```
pld: why not put this conditional inside the loop to remove duplication?

onadata/apps/main/management/commands/update_enketo_urls.py
```diff
@@ -45,25 +48,45 @@ def handle(self, *args, **options):
...
+                with open('/tmp/enketo_url', 'a') as f:
+                    if data_type == 'enketo_url':
+                        form_url = get_form_url(
+                            request, protocol=protocol,
+                            xform_uuid=xform.uuid)
+                        _enketo_url = enketo_url(form_url, id_string)
+                        MetaData.enketo_url(xform, _enketo_url)
+                    elif data_type == 'enketo_preview_url':
+                        _enketo_preview_url = get_enketo_preview_url(
+                            request, id_string, xform_uuid=xform.uuid)
```
pld: where is `id_string` defined?

onadata/apps/main/management/commands/update_enketo_urls.py
pld: can you remove all the new duplicate by having the condition affect whether `username` is `None` or not and the value of `id_string`? that looks doable to me

onadata/apps/main/management/commands/update_enketo_urls.py
DavisRayM: Changed this in the latest commits

onadata/apps/main/management/commands/update_enketo_urls.py
DavisRayM: Moved the conditional inside the loop in the latest commit.

Also, I've modified this code block removing the duplicate code. Using a conditional to effect the value of the `xform_uuid` passed into the functions.

onadata/apps/main/management/commands/update_enketo_urls.py
DavisRayM: The `id_string` was defined at the beginning of the `for` loop.  I've changed how it's being defined in the latest commits. Here is the [updated definition of the variable](https://github.com/onaio/onadata/blob/da4a526eb15e9e06a8da91563e6c4bcccd833239/onadata/apps/main/management/commands/update_enketo_urls.py#L53-L58)

onadata/libs/utils/viewer_tools.py
```diff
@@ -340,10 +343,17 @@ def get_enketo_edit_url(request, instance, return_url):
     return url
 
 
-def get_enketo_preview_url(request, username, id_string, xform_pk=None):
+def get_enketo_preview_url(
+        request, id_string, username=None, xform_pk=None, xform_uuid=None):
     """Return an Enketo preview URL."""
-    form_url = get_form_url(
-        request, username, settings.ENKETO_PROTOCOL, True, xform_pk=xform_pk)
+    if xform_pk and username:
```
pld: if we're adding the keys to kwargs below, we should definitely add them here too, eg say `username=username` etc.

But, I wonder if we should define kwargs all at once and have a single call to `get_form_url` like

```python
get_form_url_kwargs = {
    "protocol": settings.ENKETO_PROTOCOL,
    "preview": True
}
if xform_pk and username:
    get_form_url_kwargs.update{
        "username": username,
        "xform_pk": xform_pk
    }
elif xform_uuid:
    get_form_url_kwargs.update{
        "xform_uuid": xform_uuid
    }

form_url = get_form_url(request, **get_form_url_kwargs)
```

pld: I'm not sold on defining the kwargs actually being better, thoughts?

onadata/libs/serializers/data_serializer.py
```diff
@@ -28,10 +28,15 @@ def get_request_and_username(context):
...
     username = view.kwargs.get('username')
+    form_uuid = view.kwargs.get('form_uuid')
 
     if not username:
-        # get the username from the user if not set
-        username = (request.user and request.user.username)
+        # get the username from the XForm object if form_uuid is
+        # present else utilize the request users username
+        if form_uuid:
+            username = XForm.objects.get(uuid=form_uuid).user.username
```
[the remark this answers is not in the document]
DavisRayM: This state will most likely happen every time Enketo makes a submission. Other than that the only other time this would happen is if a client decided to submit date with the new endpoint.

onadata/libs/utils/viewer_tools.py
```diff
@@ -315,6 +316,8 @@ def get_form_url(request,
     if preview:
         url = '%s/preview' % url
 
+    if xform_uuid:
+        url = "{}/e/{}".format(url, xform_uuid)
     if username and xform_pk is None:
```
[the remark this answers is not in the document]
DavisRayM: String append would work perfectly here. I'll be changing those lines to:
```
url+= '/{}/{}'.format(username, xform_pk) if xform_pk else '/{}'.format(username)
```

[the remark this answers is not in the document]
DavisRayM: I'll also be changing line 320 to:

```
url += "/e/{}".format(xform_uuid) 
```
as suggested.

onadata/libs/utils/viewer_tools.py
DavisRayM: I think this would work fine. There are no downsides of using an array here, would help reduce the duplicate calls to `get_form_url` changing this to:

```
    form_url_kwargs = {
        "request": request,
        "protocol": settings.ENKETO_PROTOCOL,
        "preview": True
    }

    if xform_pk and username:
        form_url_kwargs.update({
            'xform_pk': xform_pk,
            'username': username
        })
    elif xform_uuid:
        form_url_kwargs.update({
            'xform_uuid': xform_uuid
        })

    form_url = get_form_url(**form_url_kwargs)
    values = {'form_id': id_string, 'server_url': form_url}
```

onadata/apps/main/management/commands/update_enketo_urls.py
```diff
@@ -45,26 +48,33 @@ def handle(self, *args, **options):
...
             xform = meta_data.content_object
             with open('/tmp/enketo_url', 'a') as f:
-
                 if data_type == 'enketo_url':
                     form_url = get_form_url(
-                        request, username, protocol=protocol,
-                        xform_pk=xform.pk)
+                        request, username=username, id_string=id_string,
+                        xform_pk=xform.pk,
+                        xform_uuid=xform.uuid if not username else None)
```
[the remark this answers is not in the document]
DavisRayM: I've changed this in the latest commit. Defined a variable to hold the `xform_uuid`, there is no point of time the `use_username` would evaluate to true and the `username` would evaluate to `False`; I've changed the conditional to now use the `use_username` boolean.

onadata/libs/serializers/data_serializer.py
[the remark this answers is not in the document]
pld: Can you please verify that `XForm` has an index on the `uuid` column

onadata/libs/utils/viewer_tools.py
pld: lgtm

onadata/libs/serializers/data_serializer.py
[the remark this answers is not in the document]
DavisRayM: > Can you please verify that `XForm` has an index on the `uuid` column

The `uuid` column is currently not indexed

pld:
Let's add that to this PR since now we expect to have significantly more lookups.

Are there any negative repercussions of adding this index we should be aware of?

DavisRayM:
> Let's add that to this PR since now we expect to have significantly more lookups. Are there any negative repercussions of adding this index we should be aware of?
> […](#)
> On Feb 4, 2020, at 12:10, Davis Raymond ***@***.***> wrote: ﻿ @DavisRayM commented on this pull request. In onadata/libs/serializers/data_serializer.py: > if not username: - # get the username from the user if not set - username = (request.user and request.user.username) + # get the username from the XForm object if form_uuid is + # present else utilize the request users username + if form_uuid: + username = XForm.objects.get(uuid=form_uuid).user.username Can you please verify that XForm has an index on the uuid column The uuid column is currently not indexed — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or unsubscribe.

I'm not quite sure on the repercussions of indexing the column. I'll make the change for now while I research more on the implications of it.

Keeping a `Do not merge` tag on this for now too. Will update this thread with my finding in a bit.

DavisRayM:
I've done a bit of research and consulted(@jasonrogena) before merging the migration we may need to figure out how much time it would take for the migration to run on pre-existing databases(The amount of data collected in them may affect how long the database would be inaccessible).

The other concern in indexing the field was the implication on memory usage. From my research it seems Postgresql only caches frequently used rows in memory. I'm honestly not quite sure on how this would affect memory usage.

_ **Interesting read/answer**: You can read more on how and when postgres caches data [here](https://dba.stackexchange.com/questions/25513/postgresql-index-caching)_

DavisRayM:
The migration introduced by this PR doesn't seem quite heavy, tried running it on a db with around 1200 rows of XForms took approximately `32ms`.

Screenshot 2020-02-07 at 09 44 33

CC: @pld @ivermac

onadata/libs/serializers/data_serializer.py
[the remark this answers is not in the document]
pld: added index migration looks good

pld:
> Think you can remove the DNM, a couple final comments

Which I see you've already done :)

onadata/apps/logger/models/xform.py
```diff
@@ -718,7 +718,9 @@ class XForm(XFormMixin, BaseModel):
     deleted_at = models.DateTimeField(blank=True, null=True)
     last_submission_time = models.DateTimeField(blank=True, null=True)
     has_start_time = models.BooleanField(default=False)
-    uuid = models.CharField(max_length=36, default=u'')
+    # We should probably change the default uuid value to uuid.uuid4().hex
+    # maybe ?
```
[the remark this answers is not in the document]
DavisRayM: The comment was mostly there as a reminder for me. I'll open a separate issue for this as I'm not sure of why it was initially set to an empty string.

I suspected the uuid.uuid4() function may generate duplicates possibly ? I'm not sure. I'll open an issue for this either way and remove the comment from the codebase.

onadata/apps/main/management/commands/update_enketo_urls.py
```diff
@@ -45,26 +48,35 @@ def handle(self, *args, **options):
...
+            if use_username:
+                username = meta_data.content_object.user.username
+                id_string = meta_data.content_object.id_string
+
             data_type = meta_data.data_type
             data_value = meta_data.data_value
             xform = meta_data.content_object
-            with open('/tmp/enketo_url', 'a') as f:
+            xform_pk = xform.pk
+            xform_uuid = xform.uuid if not use_username else None
```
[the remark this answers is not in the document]
DavisRayM: Not sure, If I understood this correctly. But I've changed this line to

```
xform_uuid = xform.uuid if not use_username and xform.uuid else None
```

onadata/apps/logger/models/xform.py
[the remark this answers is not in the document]
DavisRayM: Opened an issue for this here #1779

onadata/apps/main/management/commands/update_enketo_urls.py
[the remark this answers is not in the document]
pld: I'm suggesting to change to 

```python
xform_uuid = not use_username and xform.uuid
```

since this will evaluate to `False` when `useusername` is false, which is the correct behavior, we do not need to explicitly set `None` here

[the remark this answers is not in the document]
DavisRayM: The `xform_uuid` also acts as the container for the value (xform.uuid). We currently don't use this as a boolean.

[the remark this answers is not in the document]
pld: I understand, but this would not evaluate to a boolean if `not use_username` is true, it would be the intended value, no?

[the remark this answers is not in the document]
DavisRayM: My bad, yes this would be fine. Tested it out returns `False` when `not use_username` is False but that shouldn't be an issue as the functions will handle the value correctly either way.

I'll change this.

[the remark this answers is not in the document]
DavisRayM: Changed in the latest commits.

WNjihia:
LGTM 🙂 
Tested creating a form, submitting to a form, moving and replacing a form.

DavisRayM:
Another suggestion brought up by @ukanga was utilizing the `id` field to generate the enketo url. The `id` has the same properties we are looking for in order to generate the url. Which are:

- Its unique
- We'll never change

I think this change would be fine, it would also remove the need for the index on the `uuid` column too (Though I'm not sure if we should keep it for the future).

Thoughts @pld @ivermac ?

ivermac:
@DavisRayM sounds good to me

onadata/apps/api/viewsets/xform_list_viewset.py
```diff
@@ -73,15 +73,22 @@ def get_renderers(self):
 
     def filter_queryset(self, queryset):
         username = self.kwargs.get('username')
-        if username is None and self.request.user.is_anonymous:
+        form_uuid = self.kwargs.get('form_uuid')
+        if (username is None and form_uuid is None) and \
```
pld: do we need `is None`? `not` will work unless we want this to be False when `username` is False

DavisRayM: I think using `not` here is enough. I've changed this in the latest commits.

DavisRayM:
**Note for reviewers**: I've modified this branch to utilize the forms `id` instead of it's `uuid`.



> Another suggestion brought up by @ukanga was utilizing the `id` field to generate the enketo url. The `id` has the same properties we are looking for in order to generate the url. Which are:
> 
>     * Its unique
> 
>     * Unchanging; _The uuid could also change when one regenerates uuids for XForms_
> 
> 
> I think this change would be fine, it would also remove the need for the index on the `uuid` column too (Though I'm not sure if we should keep it for the future).
> 
> Thoughts @pld @ivermac ?

onadata/libs/filters.py
```diff
@@ -22,6 +22,12 @@ def filter_queryset(self, request, queryset, view):
         """
         form_id = view.kwargs.get(
             view.lookup_field, view.kwargs.get('xform_pk'))
+        lookup_field = view.lookup_field
```
pld: Codacy Issue found: [Unused variable 'lookup_field'](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4921990)

pld: Codacy Issue found: [Unused variable 'lookup_field'](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=4921990)

onadata/apps/logger/migrations/0060_auto_20200305_0357.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.9 on 2020-03-05 08:57
+
+from django.db import migrations, models
```
DavisRayM: I'm a bit conflicted on whether this should be in this PR. Feels like this index will be helpful in the future (FLOIP endpoint uses the `uuid` field quite a bit)

faith-mutua:
@DavisRayM this has passed QA. However, I have noted the following behavior; 
1. The Enketo url doesn't change after a form is moved to another project. 
2. Anonymous users cannot submit data to public forms as long as `require phone authentication`  is set to `Yes`. 

I believe this is the expected behavior.

DavisRayM:
> @DavisRayM this has passed QA. However, I have noted the following behavior;
> 
>     1. The Enketo url doesn't change after a form is moved to another project.
> 
>     2. Anonymous users cannot submit data to public forms as long as `require phone authentication`  is set to `Yes`.
> 
> 
> I believe this is the expected behavior.

Thank you @faith-mutua.

Yes those are expected. The enketo url showcased to the user shouldn't change as it links to the internal storage of enketo. The require authentication part is also intended.

PR #1783, opened 2020-02-17
Invalidate tokens on password change

### LINKED ISSUE #1782: Invalidate a users api and temporary token on password change
### Feature / Change Request

I think  we should consider invalidating / deleting a users `Token` and `TempToken` on password change.

### Benefits

- Increases account security

CC: @ivermac @pld @ukanga @WinnyTroy @lincmba

DavisRayM:
### Changes / Features 

- Invalidate a users api and temporary token when they're password
  changes.
- Add test to verify that the api and temporary tokens are renewed

- Added tests

Pending:
- Add new tests

- API Tokens will now need to be re-retrieved when one changes their password

Closes #1782

### DIFF
--- a/onadata/apps/api/viewsets/user_profile_viewset.py
+++ b/onadata/apps/api/viewsets/user_profile_viewset.py
@@ -37,6 +37,7 @@ from onadata.libs.utils.cache_tools import (cache,
                                             CHANGE_PASSWORD_ATTEMPTS,
                                             LOCKOUT_CHANGE_PASSWORD_USER)
 from onadata.libs import filters
+from onadata.libs.utils.user_auth import invalidate_and_regen_tokens
 from onadata.libs.mixins.authenticate_header_mixin import \
     AuthenticateHeaderMixin
 from onadata.libs.mixins.cache_control_mixin import CacheControlMixin
@@ -209,14 +210,20 @@ class UserProfileViewSet(
         if new_password:
             if not lock_out:
                 if user_profile.user.check_password(current_password):
+                    data = {
+                        'username': user_profile.user.username
+                    }
                     metadata = user_profile.metadata or {}
                     metadata['last_password_edit'] = timezone.now().isoformat()
                     user_profile.user.set_password(new_password)
                     user_profile.metadata = metadata
                     user_profile.user.save()
                     user_profile.save()
+                    data.update(invalidate_and_regen_tokens(
+                        user=user_profile.user))
 
-                    return Response(status=status.HTTP_204_NO_CONTENT)
+                    return Response(
+                        status=status.HTTP_200_OK, data=data)
 
                 response = change_password_attempts(request)
                 if isinstance(response, int):
--- a/docs/profiles.rst
+++ b/docs/profiles.rst
@@ -233,7 +233,11 @@ Response
 
 ::
 
-    HTTP 200 OK
+    {
+        "username": "demouser",
+        "access_token": "30557df81ac63729d2865e90f334a5d00fe5e57c",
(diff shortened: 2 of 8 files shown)


onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -209,14 +234,20 @@ def change_password(self, request, *args, **kwargs):  # noqa
...
+                        'username': user_profile.user.username
+                    }
                     metadata = user_profile.metadata or {}
                     metadata['last_password_edit'] = timezone.now().isoformat()
                     user_profile.user.set_password(new_password)
                     user_profile.metadata = metadata
                     user_profile.user.save()
                     user_profile.save()
+                    data.update(invalidate_and_regen_tokens(
+                        user=user_profile.user))
```
ukanga: Do we need to return the access and temp token at this point? Could on the UI side simply make the request to the API/user endpoint to retrieve the new information from there as it already does on login?

ivermac: The tokens that are in UI will already be invalidated so I don't think they can be used to make a request to the user endpoint unless the user logs out, no?

ukanga: Yes, not necessarily logging out but would need to log in with the new credentials. Will we be terminating the existing session and creating a new one? Essentially as you say, log out and log in a fresh?

ukanga: No much particular objection but simply wondering.

Should we update the documentation at https://api.ona.io/static/docs/profiles.html#change-authenticated-users-password to reflect that we also return the access and temp tokens?

DavisRayM: We don't necessarily have to return the access and temp token at this point, returning them just makes it easier for the API consumer to continue making requesting (Just removes the added step of requesting for a new access token).

UI's will have to change / modify their stored token values. They can decide to terminate a session or just update it's values.

DavisRayM: Yes, we should update the documentation. Thank you for bringing that up, let me make that change.

ivermac: I'd like to think we'll be terminating the existing session, right @DavisRayM ?

faith-mutua:
@DavisRayM The API token is only updated when a user changes their password from the **Settings ** page. Changing passwords from the `Forgot password` link doesn't invalidate the API token.

DavisRayM:
> @DavisRayM The API token is only updated when a user changes their password from the **Settings ** page. Changing passwords from the `Forgot password` link doesn't invalidate the API token.

Thank you for noting this @faith-mutua. The latest commits should handle this now.

faith-mutua:
@DavisRayM The API token is now being invalidated when a user changes their password via The `Forgot password` link or from the `Edit Account Settings` tab in the account settings page. This has passed QA.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -142,6 +144,29 @@ def change_password_attempts(request):
     return 1
 
 
+def invalidate_and_regen_tokens(user):
```
ukanga: Can we have this util function in `onadata.libs.authentication` module rather than a view set? Especially since it is also used by the `CustomPasswordResetTokenGenerator` class. I think it clutters the viewset.

DavisRayM: Yes, I'll make that change.

PR #1788, opened 2020-02-21
Upgrade dependencies

### LINKED ISSUE #1787: Investigate requirement upgrades required for the Django 2.2 upgrade
### Environmental Information

- Onadata version: v2.2.1

### Problem description

The current onadata installation has packages that _may_ be incompatible with Django v2.2.1 . The incompatibility may not be immediately noticeable `500` status codes or the link but may bring down functionality that we usually expect to work out of the box i.e Async functionality.

~One of the packages currently not compatible with Django v2.2.9 is [celery 4.2.1](https://github.com/onaio/onadata/blob/master/requirements/base.pip#L20) and possibly [raven](https://docs.sentry.io/clients/python/integrations/#django).~

It seems the incompatibility/failure is being caused by [PyXForm](https://github.com/XLSForm/pyxform/issues/401). This issue is fixed in their latest release.

_Celery may be the only package that is currently incompatible with Django v2.2.9. But, it won't hurt to thoroughly investigate further_

DavisRayM:

- Updated Requirements. Major updates:
    - `django-filter` 2.1.0 -> 2.2.0
    - `django-guardian` 1.5.1 -> 2.1.0
    - `django` 2.2.9 -> 2.2.10
    - `djangorestframework` 3.9.3 -> 3.11.0
- Install `djangorestframework-guardian`
- Change `MultiLookUpRouter` `base_name` argument to `basename`
- Modify `ProjectSerializer`
- Stop usage of `rest_framework.compat.six`

- Updated tests

- `CORS_ORIGIN_WHITELIST` items should have their protocol specified i.e http://localhost:3000

### Additional Information

This PR includes new migrations related to the following apps: `django_celery_results, guardian, registration.`

```
django_celery_results.0004_auto_20190516_0412
django_celery_results.0005_taskresult_worker
django_celery_results.0006_taskresult_date_created
django_celery_results.0007_remove_taskresult_hidden
guardian.0002_generic_permissions_index
registration.0005_activation_key_sha256
```

Part of #1787


DavisRayM:
Going to be downgrading from `django-guardian` `v2.2` to `v2.1` as I'm facing issues with the version ([Issue](https://github.com/django-guardian/django-guardian/pull/677)).

DavisRayM:
CI is currently failing on this error:

Screenshot 2020-03-02 at 17 14 38

The error is related to an old bug, that used to pop up regarding unique_group_names. I'm currently not able to recreate this issue with the provided `.xls`; should i:
a) Recreate the issue with another `.xls` file ? Feels like this wouldn't really test the initial bug IMO ? I'm also not sure why the fixture raised the issue.. as it doesn't seem to have two group_names of the same kind.
b) Remove the test case as it's no longer valid.

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -337,17 +337,21 @@ def create_async(self, request, *args, **kwargs):
...
-                       'data': request.FILES.get('xls_file').read()}
-                      if isinstance(request.FILES.get('xls_file'),
-                                    InMemoryUploadedFile) else
-                      {'name': fname,
-                       'path': request.FILES.get(
-                           'xls_file').temporary_file_path()})).task_id})
+                     request.user.id,
+                     request.POST,
+                     owner.id,
+                     {'name': fname, 'path': xls_file_path}).task_id})
```
DavisRayM: Had to modify this since celery tries to serialize the passed in arguments and fails. Not sure when the serialization of arguments was introduced... but passing in primitive / serializable data was the solution I thought would work here.

ukanga:
Can we have the pyxform upgrade as a separate PR from this one? 

Do the pyxform related changes take into account the existing forms when being replaced?

ukanga:
Perhaps it has to do with the change that makes all XForm XML root nodes to start with `/data/`.

> CI is currently failing on this test:
> 
> Screenshot 2020-03-02 at 17 14 38
> 
> The test is related to an old bug, that used to pop up due to unique group names. I'm currently not able to recreate this issue with the provided `.xls`. Should I:
> 
> a) Recreate the issue with another `.xls` file ? _Feels like this wouldn't really test the initial bug IMO_
> 
> _I'm also not sure why the fixture raised the issue before.. as it doesn't seem to have duplicate group names. Still tracking down any possible change on PyXForm that might've fixed it._
> 
> b) Remove the test case as it's no longer valid.

DavisRayM:
> Can we have the pyxform upgrade as a separate PR from this one?
> 
> Do the pyxform related changes take into account the existing forms when being replaced?

Yes, I'll move the pyxform changes to a separate branch.

Currently the PyXForm changes do not take into account the possible challenges, that may arise _I'm still testing this out to see what side effects the upgrade would bring. This will be easier to do once moved out to it's on separate branch; I'll also document the side effects introduced by that branch on its PR_

DavisRayM:
Yes, seems to be related to the root node change.

> Perhaps it has to do with the change that makes all XForm XML root nodes to start with `/data/`.
> 
> > CI is currently failing on this test:
> > Screenshot 2020-03-02 at 17 14 38
> > The test is related to an old bug, that used to pop up due to unique group names. I'm currently not able to recreate this issue with the provided `.xls`. Should I:
> > a) Recreate the issue with another `.xls` file ? _Feels like this wouldn't really test the initial bug IMO_
> > _I'm also not sure why the fixture raised the issue before.. as it doesn't seem to have duplicate group names. Still tracking down any possible change on PyXForm that might've fixed it._
> > b) Remove the test case as it's no longer valid.

faith-mutua:
@DavisRayM I have tested the following scenarios and they have passed QA
1. User can create a project
2. User can share a project
3. User can publish a form
4. User can create a form using the form builder
5. User can submit data via both Enketo and ODK Collect
6. User can export data.

PR #1796, opened 2020-03-03
Upgrade PyXForm dependency to v1.1.0

### LINKED ISSUE #1787: Investigate requirement upgrades required for the Django 2.2 upgrade

DavisRayM:

- Upgrade `pyxform` to `v1.1.0`

- Updated tests
- Tested functionality locally

#### Pending:
- [x] Test that this change is backwards compatible

_Still researching, impacts of this change_

Closes #1787


faith-mutua:
@DavisRayM I have tested the following scenarios and they have passed QA.
- User can publish a form
- User can replace a form - both old and new forms
- User can share a project
- User can create a form using the form builder
- User can edit forms created using the form builder
- Uploading forms with a field name `data` rises the error `There are two sections with the name data.`
- User can export data
- User can upload form media files
- User can upload XLS Report templates and generate reports.

PR #1798, opened 2020-03-12
Set deleted_by field when deleting XForms asynchronously

### LINKED ISSUE #1797: Track users who initiate asynchronous xform deletions
### Environmental Information

- Operating System: All
- Onadata version: Master
- Python version: All

### Feature request

Track users who initiate asynchronous xform deletions.

DavisRayM:

- Track users who initiate asynchronous deletion of XForms

- Updated tests

- None

Closes #1797

### DIFF
--- a/onadata/apps/api/viewsets/xform_viewset.py
+++ b/onadata/apps/api/viewsets/xform_viewset.py
@@ -710,7 +710,9 @@ class XFormViewSet(AnonymousUserPublicFormsMixin,
         if request.method == 'DELETE':
             xform = self.get_object()
             resp = {
-                u'job_uuid': tasks.delete_xform_async.delay(xform.pk).task_id,
+                u'job_uuid': tasks.delete_xform_async.delay(
+                    xform.pk,
+                    request.user.id).task_id,
                 u'time_async_triggered': datetime.now()}
             resp_code = status.HTTP_202_ACCEPTED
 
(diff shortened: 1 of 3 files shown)


onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -710,8 +710,11 @@ def delete_async(self, request, *args, **kwargs):
         if request.method == 'DELETE':
             xform = self.get_object()
             resp = {
-                u'job_uuid': tasks.delete_xform_async.delay(xform.pk).task_id,
-                u'time_async_triggered': datetime.now()}
+                u'job_uuid': tasks.delete_xform_async.delay(
+                    xform.pk,
+                    request.user.id).task_id,
+                u'time_async_triggered': datetime.now(),
+                u'triggered_by': request.user.username}
```
ivermac: Do we need `triggered_by` since we'll be saving the user who triggered the delete?

DavisRayM: Mostly kept it they're for information purposes (if that makes sense) wanted to be extra verbose. On the first async task status just in case.

TL;DR: We don't really need it

ivermac: Cool, let's remove it for now and may be add it when we need it

DavisRayM: I've removed the `triggered_by` field in the latest commits. Please have a look @ivermac

PR #1799, opened 2020-03-17
Rename variable 'gen_consistent' to 'consistent_urls'

DavisRayM:

- Rename variable `gen_consistent` to `consistent_urls`

CI Tests

None

### DIFF
--- a/onadata/apps/api/viewsets/xform_viewset.py
+++ b/onadata/apps/api/viewsets/xform_viewset.py
@@ -395,7 +395,7 @@ class XFormViewSet(AnonymousUserPublicFormsMixin,
         form_url = get_form_url(
             request, self.object.user.username,
             protocol=settings.ENKETO_PROTOCOL,
-            xform_pk=self.object.pk, gen_consistent=True)
+            xform_pk=self.object.pk, generate_consistent_urls=True)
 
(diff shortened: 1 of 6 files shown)


DavisRayM:
> The change looks good. Why is the value of `consistent_urls` a boolean?

It's used as a "flag" of sorts to determine whether we should generate the enketo urls consistently [here](https://github.com/onaio/onadata/pull/1799/files#diff-0c09bacc2f47a378a518ee619d348093R319).

DavisRayM:
> > The change looks good. Why is the value of `consistent_urls` a boolean?
> 
> It's used as a "flag" of sorts to determine whether we should generate the enketo urls consistently [here](https://github.com/onaio/onadata/pull/1799/files#diff-0c09bacc2f47a378a518ee619d348093R319).

I'm actually not entirely sure if we'll ever have to recreate the old urls ? Might just be better to fully phase out the old way of generating enketo urls ?

ivermac:
Can I suggest we rename `consistent_urls` to `generate_ consistent_urls`?

DavisRayM:
> Can I suggest we rename `consistent_urls` to `generate_consistent_urls` 🙈?

Sure, your suggestion seems more verbose I believe. I'll make this change.

PR #1802, opened 2020-03-31
[WIP] Allow users to resubmit soft-deleted instances

### LINKED ISSUE #1801: Authenticated user should only submit to forms they have access to.
### Environmental Information

- Operating System: Linux
- Onadata version: all
- Python version: Python 3.6+

### Problem description

An authenticated submission allows a user to submit to any account if the id_string and `formhub uuid `match even if they do not have explicit permissions to do that.

### Expected behaviour

The authenticated user should only submit to the forms they have explicit permissions to.

1. As user bob publish a form and make submissions to your (Bob) account
2. As user bob use ODK Briefcase to pull form submissions in your (Bob) account
3. As user Alice, use ODK Briefcase to push the submissions in your (Alice) account
4. As user Alice, you will notice the form is published but the submissions would not be reflecting.

### Additional Information

DavisRayM:

- Allow users to submit submissions that have been `soft-deleted`

- Updated tests

- Users will be able to submit new submissions with the same uuid as a submission in the system as long as the other submissions have been deleted.

Closes #1801

### DIFF
--- a/onadata/apps/logger/tests/test_form_submission.py
+++ b/onadata/apps/logger/tests/test_form_submission.py
@@ -215,6 +215,14 @@ class TestFormSubmission(TestBase):
         self._make_submission(xml_submission_file_path)
         self.assertEqual(self.response.status_code, 202)
 
+        # Ensure one can resubmit the data if submissions is soft-deleted
+        instance = Instance.objects.last()
+        instance.set_deleted()
+        instance.save()
+
+        self._make_submission(xml_submission_file_path)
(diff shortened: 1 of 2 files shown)


DavisRayM:
Decided to approach this by allowing the creation of the instance as long as the duplicate has been deleted. At the moment I'm a bit unsure about how this will affect some functionalities (Still looking into this)

DavisRayM:
Kinda missed the fact that this has to be published to a different account..... I'll try this out, hopefully my findings still apply to the same.

PR #1804, opened 2020-04-02
Fix an issue caused when a user tries to submit to a copy of a shared form

### LINKED ISSUE #1801: Authenticated user should only submit to forms they have access to.

DavisRayM:

- Modify the get_xform_from_submission function

- [x] Add/Update tests

### Pending changes

- [x] Support the old submission processing flow. _We should not be completely switching towards this new form of submission processing_

Closes #1801

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
@@ -181,7 +181,7 @@ class TestXFormSubmissionViewSet(TestAbstractViewSet, TransactionTestCase):
                 auth = DigestAuth('alice', 'bobbob')
                 request.META.update(auth(request.META, response))
                 response = self.view(request)
-                self.assertEqual(response.status_code, 403)
+                self.assertEqual(response.status_code, 404)
 
     def test_post_submission_authenticated_json(self):
         """
--- a/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
@@ -21,6 +21,7 @@ from onadata.apps.api.viewsets.xform_submission_viewset import \
     XFormSubmissionViewSet
 from onadata.apps.logger.models import Attachment, Instance, XForm
 from onadata.libs.permissions import DataEntryRole
+from onadata.libs.utils.common_tools import get_uuid
 
 
 # pylint: disable=W0201,R0904,C0103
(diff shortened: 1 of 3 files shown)


onadata/libs/utils/logger_tools.py
```diff
@@ -127,31 +127,54 @@ def get_uuid_from_submission(xml):
...
-
-            return xform
+            try:
+                check_submission_permissions(request, xform)
+            except PermissionDenied as e:
+                if xform.user == username:
+                    # If the retrieved form is the one we were looking for
+                    # raise a permission denied error
+                    raise e
+                pass
```
pld, line 165: Codacy Issue found: [Unnecessary pass statement](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5294022)

onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
```diff
@@ -181,7 +181,7 @@ def test_post_submission_uuid_other_user_username_not_provided(self):
                 auth = DigestAuth('alice', 'bobbob')
                 request.META.update(auth(request.META, response))
                 response = self.view(request)
-                self.assertEqual(response.status_code, 403)
+                self.assertEqual(response.status_code, 404)
```
DavisRayM, line 184: I feel like a `404` here is more appropriate... ? The user used the `/submission` endpoint which in this case means we have to find a form matching the submission data(uuid & id_string) that the user has permission to submit to.

_In this scenario we've located the XForm using `uuid` but the user doesn't have permission to it (Assumption made: `uuid` may not be reliable to find the actual form the user want's to submit to), as such we try to locating the XForm using the id_string within the users owned forms._

onadata/libs/utils/logger_tools.py
```diff
@@ -127,29 +127,54 @@ def get_uuid_from_submission(xml):
...
+                # Verify request user has permission to make submissions to the
+                # XForm
+                check_submission_permissions(request, xform)
+                return xform
+            except PermissionDenied as e:
+                # Check if the owner_username is equal to the XForm owner
+                # Assumption: If the owner_username is equal to the XForm
+                # owner we've retrieved the correct form.
+                if owner_username and xform.user.username == owner_username:
+                    raise e
```
ivermac, line 169: I'm confused a bit here. Why are we raising an error if the owner username is the same as the xform user?

DavisRayM: We raise the permission error since we've technically retrieved the correct form the user was trying to make a submission too and they don't have the permission required to make the submission

onadata/libs/utils/logger_tools.py
```diff
@@ -127,29 +127,54 @@ def get_uuid_from_submission(xml):
...
+    :param (str) owner_username: The owner of the target XForm. Default: None
+    """
+    if not uuid and not (owner_username or request.user):
         raise InstanceInvalidUserError()
 
+    username = None
+    if owner_username:
+        username = owner_username
+    elif request.user.is_authenticated:
+        username = request.user.username
```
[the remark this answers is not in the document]
DavisRayM: I don't necessarily want to set the `username` to the request user. The username is sort of used as the target account to retrieve the `XForm` if it fails on retrieving using `uuid`.

As such, I was thinking the `owner_username` if present should be the go-to username to use here... Though I'll look further into this, I believe the username passed into this function can be the `request.user.username` if the target username isn't set during the making of the submission.

onadata/apps/logger/tests/test_digest_authentication.py
```diff
@@ -60,10 +61,9 @@ def test_fail_authenticated_submissions_to_wrong_account(self):
         # Authentication required
         self.assertEqual(self.response.status_code, 401)
         auth = DigestAuth('dennis', 'dennis')
-        self._make_submission(xml_submission_file_path, add_uuid=True,
-                              auth=auth)
-        # Not allowed
-        self.assertEqual(self.response.status_code, 403)
+        with self.assertRaises(Http404):
+            self._make_submission(xml_submission_file_path, add_uuid=True,
+                                  auth=auth)
```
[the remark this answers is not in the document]
DavisRayM: Yes, this fails because they didn't provide the `username` of the user who owns the username and because we couldn't find a form matching the submission within the request users account.

onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
```diff
@@ -181,7 +182,58 @@ def test_post_submission_uuid_other_user_username_not_provided(self):
...
+                count = XForm.objects.all().count()
+                original_form_pk = self.xform.pk
+                duplicate_form = self.xform
+                duplicate_form.pk = None
+                duplicate_form.uuid = get_uuid()
+                duplicate_form.user = alice_profile.user
+                duplicate_form.save()
+
+                self.assertNotEqual(original_form_pk, duplicate_form.pk)
+                self.assertEqual(XForm.objects.all().count(), count+1)
```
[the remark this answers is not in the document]
DavisRayM: Yes, that's true. I'll change this

onadata/libs/utils/logger_tools.py
```diff
@@ -127,29 +127,48 @@ def get_uuid_from_submission(xml):
     return len(split_xml) > 1 and split_xml[1] or None
 
 
-def get_xform_from_submission(xml, username, uuid=None):
-    # check alternative form submission ids
-    uuid = uuid or get_uuid_from_submission(xml)
-
-    if not username and not uuid:
+def get_xform_from_submission(
+        xml, request, username, uuid=None):
```
ukanga, line 131: I think I am not redefining existing functions by introducing new parameters before existing parameters. I would perhaps go by adding a new function or adding the new parameter as optional. And perhaps work towards a deprecation process.
Can we leave it at?
```python
def get_xform_from_submission(xml, username, uuid=None, request=None):
...
if request:
   # new way
else:
   # existing old way.
```

DavisRayM: Sure, I'll make this change.

onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
```diff
@@ -181,7 +182,58 @@ def test_post_submission_uuid_other_user_username_not_provided(self):
...
+                count = XForm.objects.all().count()
+                original_form_pk = self.xform.pk
+                duplicate_form = self.xform
+                duplicate_form.pk = None
+                duplicate_form.uuid = get_uuid()
+                duplicate_form.user = alice_profile.user
+                duplicate_form.save()
+
+                self.assertNotEqual(original_form_pk, duplicate_form.pk)
+                self.assertEqual(XForm.objects.count(), count+1)
```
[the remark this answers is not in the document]
DavisRayM: I don't believe so... this should compare a number with a number `count` is initially set to `XForm.objects.count()` before duplicating the form. I'm not entirely sure if we should refresh from DB, but it wouldn't hurt to do so I believe.

_I'll make the changes and wait for CI to pass and merge this_

PR #1808, opened 2020-04-07
Refactor submission deletion queryset

KipSigei:
- Refactor to remove `.update()` since it doesn't call  the Instance model's. `save()` method
- `.update()` method doesn't send `post_save` signal required by google exports or any other micro service

Closes #


moshthepitt:
Should we add a test that it is now called?

PR #1814, opened 2020-04-20
Fix form uploads caching issue

KipSigei:
- Add signal handler to clear list of forms from cache when uploading a form to be able to show updated list with newly uploaded form

- Tests

Closes https://github.com/onaio/zebra/issues/5641


KipSigei:
Closing this PR so that we can address under https://github.com/onaio/onadata/pull/1819

PR #1820, opened 2020-04-30
OpenData endpoint updates

WinnyTroy:
This approach aims to ensure that the column header fields used to create tableau dashboards match the row header fields

Steps taken
- Removed metadata field from both the column headers and row headers
- Include column fields not present in the submission data
- Flatten repeat data in the submission data.

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
@@ -110,6 +110,36 @@ class TestOpenDataViewSet(TestBase):
         # cast generator response to list so that we can get the response count
         self.assertEqual(len(streaming_data(response)), 4)
 
+    def test_column_headers_are_synchronized(self):
+        self.view = OpenDataViewSet.as_view({
+            'get': 'schema'
+        })
+
+        _open_data = self.get_open_data_object()
+        uuid = _open_data.uuid
+        request1 = self.factory.get('/', **self.extra)
+        response1 = self.view(request1, uuid=uuid)
+        self.assertEqual(response1.status_code, 200)
+        self.assertListEqual(
+            ['column_headers', 'connection_name', 'table_alias'],
+            sorted(list(response1.data))
+        )
+
+        self._make_submissions()
+        self.view = OpenDataViewSet.as_view({
+            'get': 'data'
+        })
+        _open_data = self.get_open_data_object()
+        uuid = _open_data.uuid
+        request2 = self.factory.get('/', **self.extra)
+        response2 = self.view(request2, uuid=uuid)
+        self.assertEqual(response2.status_code, 200)
+        # cast generator response to list for easy manipulation
+        row_data = streaming_data(response2)
+        # assert that the length of the column_headers
+        # is equal to the row_headers randomly picked
+        self.assertEqual(len(response1.data['column_headers']), len(row_data[3]))
+
     def test_get_data_with_pagination(self):
         self._make_submissions()
         self.view = OpenDataViewSet.as_view({
--- a/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
@@ -111,70 +111,33 @@ class TestOpenDataViewSet(TestBase):
         # cast generator response to list so that we can get the response count
         self.assertEqual(len(streaming_data(response)), 4)
 
-    def test_tableau_data_fetch(self):  # pylint: disable=invalid-name
+    def test_tableau_get_group_data(self):
         """
-        Test that the row headers generated
-        match the column headers for the same form
+        Test tableau unpacks group data successfully
         """
-        self.view = OpenDataViewSet.as_view({
-            'get': 'schema'
-        })
-        self._submission_time = parse_datetime('2013-02-18 15:54:01Z')
-        self.fixture_dir = os.path.join(
-            self.this_directory, 'fixtures', 'csv_export')
-        path = os.path.join(self.fixture_dir, 'tutorial_w_repeats.xls')
-        self._publish_xls_file_and_set_xform(path)
-        path = os.path.join(self.fixture_dir, 'tutorial_w_repeats.xml')
-        self._make_submission(
-            path, forced_submission_time=self._submission_time)
-
-        _open_data = get_or_create_opendata(self.xform)
-        uuid = _open_data[0].uuid
-        request1 = self.factory.get('/', **self.extra)
-        response1 = self.view(request1, uuid=uuid)
-        self.assertEqual(response1.status_code, 200)
-        self.assertListEqual(
-            ['column_headers', 'connection_name', 'table_alias'],
-            sorted(list(response1.data))
-        )
-
-        expected_column_headers = [
-            '_gps_altitude',
-            '_gps_latitude',
-            '_gps_longitude',
-            '_gps_precision',
-            '_id',
-            'age',
-            'children_1__childs_age',
-            'children_1__childs_name',
-            'children_2__childs_age',
-            'children_2__childs_name',
-            'children_3__childs_age',
-            'children_3__childs_name',
-            'children_4__childs_age',
-            'children_4__childs_name',
-            'gps',
-            'has_children',
-            'meta_instanceID',
-            'name',
-            'picture',
-            'web_browsers_chrome',
-            'web_browsers_firefox',
-            'web_browsers_ie',
-            'web_browsers_safari'
-        ]
-
+        self._make_submissions()
         self.view = OpenDataViewSet.as_view({
             'get': 'data'
         })
-        request2 = self.factory.get('/', **self.extra)
-        response2 = self.view(request2, uuid=uuid)
-        self.assertEqual(response2.status_code, 200)
+        _open_data = self.get_open_data_object()
+        uuid = _open_data.uuid
+
+        request = self.factory.get('/', **self.extra)
+        response = self.view(request, uuid=uuid)
+        self.assertEqual(response.status_code, 200)
         # cast generator response to list for easy manipulation
-        row_data = streaming_data(response2)
-        row_data_fields = [k for k in row_data[0].keys()]
-        row_data_fields.sort()
-        self.assertEqual(row_data_fields, expected_column_headers)
+        row_data = streaming_data(response)
+        self.assertEqual(len(row_data), 4)
+        # Confirm group data is available in tableau
+        self.assertEqual(
+            row_data[3]['transport_available_'
+                        'transportation_types_to_referral_facility'],  # noqa
+            'taxi other')
+        self.assertEqual(
+            row_data[3][
+                'transport_loop_over_transport_'
+                'types_frequency_taxi_frequency_to_referral_facility'],  # noqa
+            'daily')
 
     def test_get_data_with_pagination(self):
         self._make_submissions()
@@ -344,3 +307,128 @@ class TestOpenDataViewSet(TestBase):
         request = self.factory.get('/', **self.extra)
         response = self.view(request, uuid=uuid)
         self.assertEqual(response.status_code, 404)
+
+
+class TestOpenData(TestBase):
+    """
+    Test tableau unpacks repeat data successfully
+    """
+
+    def setUp(self):
+        super(TestOpenData, self).setUp()
+        self._create_user_and_login()
+        self._submission_time = parse_datetime('2020-02-18 15:54:01Z')
+        self.fixture_dir = os.path.join(
+            self.this_directory, 'fixtures', 'csv_export')
+        path = os.path.join(self.fixture_dir, 'tutorial_w_repeats.xls')
+        self._publish_xls_file_and_set_xform(path)
+        path = os.path.join(self.fixture_dir, 'tutorial_w_repeats.xml')
+        self.factory = RequestFactory()
+        self.extra = {
+            'HTTP_AUTHORIZATION': 'Token %s' % self.user.auth_token}
+        self._make_submission(
+            path, forced_submission_time=self._submission_time)
+
+        self.view = OpenDataViewSet.as_view({
+            'post': 'create',
+            'patch': 'partial_update',
+            'delete': 'destroy',
+            'get': 'data'
+        })
+
+    def test_tableau_data_fetch(self):  # pylint: disable=invalid-name
+        """
+        Test that the row headers generated
+        match the column headers for the same form
+        """
+        self.view = OpenDataViewSet.as_view({
+            'get': 'schema'
+        })
+
+        _open_data = get_or_create_opendata(self.xform)
+        uuid = _open_data[0].uuid
+        request1 = self.factory.get('/', **self.extra)
+        response1 = self.view(request1, uuid=uuid)
+        self.assertEqual(response1.status_code, 200)
+        self.assertListEqual(
+            ['column_headers', 'connection_name', 'table_alias'],
+            sorted(list(response1.data))
+        )
+
+        expected_column_headers = [
+            '_gps_altitude',
+            '_gps_latitude',
+            '_gps_longitude',
+            '_gps_precision',
+            '_id',
+            'age',
+            'children_1__childs_age',
+            'children_1__childs_name',
+            'children_2__childs_age',
+            'children_2__childs_name',
+            'children_3__childs_age',
+            'children_3__childs_name',
+            'children_4__childs_age',
+            'children_4__childs_name',
+            'gps',
+            'has_children',
+            'meta_instanceID',
+            'name',
+            'picture',
+            'web_browsers_chrome',
+            'web_browsers_firefox',
+            'web_browsers_ie',
+            'web_browsers_safari'
+        ]
+
+        self.view = OpenDataViewSet.as_view({
+            'get': 'data'
+        })
+        request2 = self.factory.get('/', **self.extra)
+        response2 = self.view(request2, uuid=uuid)
+        self.assertEqual(response2.status_code, 200)
+        # cast generator response to list for easy manipulation
+        row_data = streaming_data(response2)
+        row_data_fields = [k for k in row_data[0].keys()]
+        row_data_fields.sort()
+        self.assertEqual(row_data_fields, expected_column_headers)
+
+    def test_tableau_get_repeat_data(self):
+        """
+        Test that data received from repeats is flattened
+        and the tablau data endpoint returns accurate data
+        for this question type
+        """
+        self.view = OpenDataViewSet.as_view({
+            'get': 'data'
+        })
+
+        repeat_data = [
+            {
+                'children/childs_age': 12,
+                'children/childs_name': 'Tom'
+            },
+            {
+                'children/childs_age': 5,
+                'children/childs_name': 'Dick'
+            }
+        ]
+
+        inst = Instance.objects.filter(
+            xform=self.xform).first()
+        self.assertEqual(inst.json['children'], repeat_data)
+
+        _open_data = get_or_create_opendata(self.xform)
+        uuid = _open_data[0].uuid
+        request = self.factory.get('/', **self.extra)
+        response = self.view(request, uuid=uuid)
+        self.assertEqual(response.status_code, 200)
+        # cast generator response to list for easy manipulation
+        row_data = streaming_data(response)
+        # Test that Tableau receives this data
+        self.assertEqual(
+            row_data[0]['children_1__childs_name'],
+            'Tom')
+        self.assertEqual(
+            row_data[0]['children_2__childs_age'],
+            5)
--- a/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
@@ -110,7 +110,11 @@ class TestOpenDataViewSet(TestBase):
         # cast generator response to list so that we can get the response count
         self.assertEqual(len(streaming_data(response)), 4)
 
-    def test_column_headers_are_synchronized(self):
+    def test_column_headers_match_row_heaaders(self):
+        """
+        Test that the row headers generated
+        match the column headers for the same form
+        """
         self.view = OpenDataViewSet.as_view({
             'get': 'schema'
         })
@@ -124,6 +128,8 @@ class TestOpenDataViewSet(TestBase):
             ['column_headers', 'connection_name', 'table_alias'],
             sorted(list(response1.data))
         )
+        expected_column_header_fields = [
+            k['id'] for k in response1.data['column_headers']]
 
         self._make_submissions()
         self.view = OpenDataViewSet.as_view({
@@ -136,10 +142,10 @@ class TestOpenDataViewSet(TestBase):
         self.assertEqual(response2.status_code, 200)
         # cast generator response to list for easy manipulation
         row_data = streaming_data(response2)
-        # assert that the length of the column_headers
-        # is equal to the row_headers randomly picked
+        row_data_fields = [k for k in row_data[2].keys()]
         self.assertEqual(
-            len(response1.data['column_headers']), len(row_data[3]))
+            row_data_fields.sort(),
+            expected_column_header_fields.sort())
 
     def test_get_data_with_pagination(self):
         self._make_submissions()
--- a/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
@@ -138,7 +138,8 @@ class TestOpenDataViewSet(TestBase):
         row_data = streaming_data(response2)
         # assert that the length of the column_headers
         # is equal to the row_headers randomly picked
-        self.assertEqual(len(response1.data['column_headers']), len(row_data[3]))
+        self.assertEqual(
+            len(response1.data['column_headers']), len(row_data[3]))
 
     def test_get_data_with_pagination(self):
         self._make_submissions()
--- a/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_open_data_viewset.py
@@ -2,10 +2,11 @@
 """
 OpenData tests.
 """
-
+import os
 import json
 
 from django.test import RequestFactory
+from django.utils.dateparse import parse_datetime
 from django.contrib.auth.models import User
 
 from onadata.apps.logger.models import OpenData, Instance
@@ -110,7 +111,7 @@ class TestOpenDataViewSet(TestBase):
         # cast generator response to list so that we can get the response count
         self.assertEqual(len(streaming_data(response)), 4)
 
-    def test_column_headers_match_row_heaaders(self):
+    def test_tableau_data_fetch(self):  # pylint: disable=invalid-name
         """
         Test that the row headers generated
         match the column headers for the same form
@@ -118,9 +119,17 @@ class TestOpenDataViewSet(TestBase):
         self.view = OpenDataViewSet.as_view({
             'get': 'schema'
         })
+        self._submission_time = parse_datetime('2013-02-18 15:54:01Z')
+        self.fixture_dir = os.path.join(
+            self.this_directory, 'fixtures', 'csv_export')
+        path = os.path.join(self.fixture_dir, 'tutorial_w_repeats.xls')
+        self._publish_xls_file_and_set_xform(path)
+        path = os.path.join(self.fixture_dir, 'tutorial_w_repeats.xml')
+        self._make_submission(
+            path, forced_submission_time=self._submission_time)
 
-        _open_data = self.get_open_data_object()
-        uuid = _open_data.uuid
+        _open_data = get_or_create_opendata(self.xform)
+        uuid = _open_data[0].uuid
         request1 = self.factory.get('/', **self.extra)
         response1 = self.view(request1, uuid=uuid)
         self.assertEqual(response1.status_code, 200)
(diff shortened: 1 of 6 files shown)


onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -42,6 +52,35 @@ class OpenDataViewSet(ETagsMixin, CacheControlMixin,
     MAX_INSTANCES_PER_REQUEST = 1000
     pagination_class = StandardPageNumberPagination
 
+    def clean_data_fields(self, data, xform):
```
pld, line 55: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5447927)

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -175,10 +225,16 @@ def schema(self, request, **kwargs):
...
-            headers = xform.get_headers() + ['_id']
+            headers = xform.get_headers()
             self.xform_headers = replace_special_characters_with_underscores(
                 headers)
 
+            # cache a clean list of column headers
+            # using the form id_string as the key
+            column_headers = '{}_{}'.format(
+                TABLEAU_COLUMN_HEADERS, xform.id_string)
+            cache.set(column_headers, self.xform_headers)
```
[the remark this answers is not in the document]
DavisRayM: Seems like the `replace_special_characters_with_underscores` doesn't remove the extra fields at all too ? Not sure where that's happening....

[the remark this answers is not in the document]
WinnyTroy: > Why did you chose to cache the column headers ? I think the cleaned headers are easily re-creatable ?

Yeah, I have removed using cache. since we already have the cform object, I have just re-created how we created headers.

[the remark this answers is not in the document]
WinnyTroy: > Seems like the `replace_special_characters_with_underscores` doesn't remove the extra fields at all too ? Not sure where that's happening....

Yeah, it was being done in the `self.flatten_xform_columns()` function. BUt I have moved this to this function, seems cleaner.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -42,6 +53,33 @@ class OpenDataViewSet(ETagsMixin, CacheControlMixin,
     MAX_INSTANCES_PER_REQUEST = 1000
     pagination_class = StandardPageNumberPagination
 
+    def clean_data_fields(self, data, xform):
```
pld, line 56: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5447927)

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -42,6 +53,33 @@ class OpenDataViewSet(ETagsMixin, CacheControlMixin,
     MAX_INSTANCES_PER_REQUEST = 1000
     pagination_class = StandardPageNumberPagination
 
+    def clean_data_fields(self, data, xform): #pylint disable=no-self-use
```
pld, line 56: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5447927)

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -138,11 +180,19 @@ def data(self, request, **kwargs):
 
             if should_paginate:
                 instances = self.paginate_queryset(instances)
+            # Remove metadata fields from the instance
+            for row in DataInstanceSerializer(instances, many=True).data:
```
DavisRayM, line 180: Could we exclude these fields within the serializer ? It would remove the need for us to manually filter them out here

onadata/apps/api/viewsets/open_data_viewset.py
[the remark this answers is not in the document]
DavisRayM: Mhh I think having the `flatten_xform_columns` function separate is fine.. I just didn't see where it was being called, might have missed that part the last time I reviewed this.

_Also wasn't expecting us to cache the raw form of the headers and then flatten the columns...._

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -42,6 +53,33 @@ class OpenDataViewSet(ETagsMixin, CacheControlMixin,
     MAX_INSTANCES_PER_REQUEST = 1000
     pagination_class = StandardPageNumberPagination
 
+    def clean_data_fields(self, data, xform):  # pylint disable=no-self-use
```
pld, line 56: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5447927)

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -42,6 +53,33 @@ class OpenDataViewSet(ETagsMixin, CacheControlMixin,
     MAX_INSTANCES_PER_REQUEST = 1000
     pagination_class = StandardPageNumberPagination
 
+    def clean_data_fields(self, data, xform):  # pylint disable=R0201
```
pld, line 56: Codacy Issue found: [Method could be a function](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5447927)

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: Moved that

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -29,9 +30,19 @@
 
 
 def replace_special_characters_with_underscores(data):
+    # Remove metadata fields from the column headers
+    remove_metadata_fields(data)
```
DavisRayM, line 34: Let's leave this out of `replace_special_characters_with_underscores`. _Keeping it separate is perfect, initially thought this function was supposed to clean the data_

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -29,9 +30,19 @@
...
     return [re.sub(r"\W", r"_", a) for a in data]
 
 
+def remove_metadata_fields(data):
+    for field in METADATA_FIELDS:
+        if field in data:
+            data.remove(field)
+        list(set(data))
+    return data
```
DavisRayM, line 43: Is it possible to have this functionality within the Data serializer ? I'm guessing that's where it's used ?

You can do so by using the [`exclude` variable](https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include)

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -42,6 +53,33 @@ class OpenDataViewSet(ETagsMixin, CacheControlMixin,
...
+                column_headers = set(tableau_headers)
+                row_headers = set(row)
+                diff = column_headers.difference(row_headers)
+
+                # set default values for these additional fields
+                left_out_fields = {
+                    field: "" for field in diff if not field.startswith('_')}
+                row.update(left_out_fields)
+                result.append(row)
+        return result
```
DavisRayM, line 81:
```suggestion
        result = []
        if data:
           headers = xform.get_headers()
           tableau_headers = remove_meta_fields(headers)
            for row in data:
                # Include fields from the column headers
                diff = set(tableau_headers).difference(set(row))

                # Shouldn't the headers here be cleared of all `_` fields ?
                # Other than '_id' ?
                left_out_fields = {
                    field: "" for field in diff if not field.startswith('_')}
                row.update(left_out_fields)
                result.append(row)
        return result
```

onadata/libs/serializers/data_serializer.py
```diff
@@ -110,6 +111,10 @@ def to_representation(self, instance):
         ret = super(DataInstanceSerializer, self).to_representation(instance)
         if 'json' in ret:
             ret = ret['json']
+            # Remove metadata fields from the instance
+            for field in METADATA_FIELDS:
+                if field in ret:
+                    del ret[field]
```
DavisRayM, line 117: Could we `exclude` these fields ?

onadata/apps/api/viewsets/open_data_viewset.py
DavisRayM: Is it possible to have this functionality within the Data serializer ? I'm guessing that's where it's used ?

You can do so by using the [`exclude` variable](https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include)

onadata/libs/serializers/data_serializer.py
WinnyTroy: Might need to move this to an entirely new serializer as this serializer is also used for the data endpoint.
I'll change this

WinnyTroy: Actually, I dont believe we can use `excclude` here
As we need to remove some of the fields we get from the json we receive for a specific submission
Exclude would have worked if this was removing some of the fields from the model.
Correct me if am wrong?

DavisRayM: Mhh yea, thats true

onadata/libs/utils/cache_tools.py
```diff
@@ -26,6 +26,31 @@
 XFORM_LINKED_DATAVIEWS = "xfs-linked_dataviews"
 PROJECT_LINKED_DATAVIEWS = "ps-project-linked_dataviews"
 
+# Cache names used in the open_data viewet
+METADATA_FIELDS = [
```
ukanga, line 30: I don't think these fields are being cached, perhaps move this to commono_tags.py?

onadata/libs/utils/cache_tools.py
```diff
@@ -26,6 +26,31 @@
...
+    '_notes',
+    '_uuid',
+    '_tags',
+    '_bamboo_dataset_id',
+    '_attachments',
+    '_geolocation',
+    '_media_count',
+    '_total_media',
+    '_submitted_by',
+    'meta_instanceID',
```
ukanga, line 46: The `meta/instanceID` is part of the submitted data, I don't think we should remove it. `meta/deprecatedID` is usually also part of the data but only shows up if a record has been edited.

onadata/libs/utils/cache_tools.py
WinnyTroy: Moved

onadata/libs/utils/cache_tools.py
WinnyTroy: Sure..seems consistent in both column data and row data from tests.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -139,10 +172,13 @@ def data(self, request, **kwargs):
...
+            cleaned_instance_fields = clean_data_fields(
+                TableauDataSerializer(instances, many=True).data, xform)
+
             csv_df_builder = CSVDataFrameBuilder(
                 xform.user.username, xform.id_string, include_images=False)
             data = csv_df_builder._format_for_dataframe(
-                DataInstanceSerializer(instances, many=True).data)
+                cleaned_instance_fields)
 
             return self._get_streaming_response(data)
```
ukanga, line 183: I am wondering whether we should implement `clean_data_fields()` as a generator function to limit the memory overhead it could have.  Seems we would put all data in memory for a form here, but it also seems that might have been the case with the previous implementation.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -84,6 +115,8 @@ def append_to_tableau_colulmn_headers(header, question_type=None):
                 'dataType': quest_type,
                 'alias': header
             })
+        # Remove metadata fields from the column headers
+        remove_metadata_fields(self.xform_headers)
```
ukanga, line 119: Should this be:

```python
self.xform_headers = remove_metadata_fields(self.xform_headers)
```
Or better yet

```python
xform_headers = remove_metadata_fields(self.xform_headers)
...
for header in xform_headers:
    ...
```
The function `remove_metadata_fields()` does return data.
```

File: onadata/apps/api/tests/viewsets/test_open_data_viewset.py [hunk 16]
```diff
@@ -110,6 +110,41 @@ def test_get_data_using_uuid(self):
...
+        request2 = self.factory.get('/', **self.extra)
+        response2 = self.view(request2, uuid=uuid)
+        self.assertEqual(response2.status_code, 200)
+        # cast generator response to list for easy manipulation
+        row_data = streaming_data(response2)
+        # assert that the length of the column_headers
+        # is equal to the row_headers randomly picked
+        self.assertEqual(
+            len(response1.data['column_headers']), len(row_data[3]))
+
```
ukanga, line 147: Would it be possible to compare actual fields to confirm that if you are expecting `'age', 'name', '_id'` you actually get those values? A length comparison only means you are expecting a specific length of the result, it does not ensure you are getting the data you intend to get.

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: yeah....changed that. Thanks for that.

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: Yes, that does make it more optimal.
I have changed that to a generator expression

onadata/apps/api/tests/viewsets/test_open_data_viewset.py
WinnyTroy: Sure, updated this to check for specific fields.

I initially was testing using length, in an attempt to make sure the same number of fields were present for both the row headers and the column headers

onadata/apps/api/tests/viewsets/test_open_data_viewset.py
```diff
@@ -110,6 +110,43 @@ def test_get_data_using_uuid(self):
...
+        uuid = _open_data.uuid
+        request2 = self.factory.get('/', **self.extra)
+        response2 = self.view(request2, uuid=uuid)
+        self.assertEqual(response2.status_code, 200)
+        # cast generator response to list for easy manipulation
+        row_data = streaming_data(response2)
+        row_data_fields = [k for k in row_data[2].keys()]
+        self.assertEqual(
+            row_data_fields.sort(),
+            expected_column_header_fields.sort())
```
DavisRayM, line 148: This test assumes the first response(`response1`) is always what we should be expecting to be correct ? 

_Could we change the `expected_headers` to be a static dict `['header1', 'header2']` to make this less dynamic ? I feel like this test will never fail...._

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -32,6 +33,36 @@ def replace_special_characters_with_underscores(data):
...
+        headers = xform.get_headers()
+        tableau_headers = remove_metadata_fields(headers)
+        for row in data:
+            # Include fields from the column headers
+            diff = set(tableau_headers).difference(set(row))
+            left_out_fields = {
+                field: None for field in diff}
+            row.update(left_out_fields)
+            result.append(row)
+    yield result
```
DavisRayM, line 52: Could we yield rows instead here ? _The generator as is will still cycle through the entire data object and return result as it was doing when it was a function.... I think it'll be better if we made this a generator that returns rows after cleaning them_

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -32,6 +33,36 @@ def replace_special_characters_with_underscores(data):
...
+
+def remove_metadata_fields(data):
+    """
+    Clean out unneccessary metadata fields
+    """
+    for field in METADATA_FIELDS:
+        if field in data:
+            data.remove(field)
+        list(set(data))
+    return data
```
DavisRayM, line 63:
```suggestion
    for field in METADATA_FIELDS:
        if field in data:
            data.remove(field)
    return list(set(data))
```

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,87 @@
...
+    """
+    for field in METADATA_FIELDS:
+        if field in data:
+            data.remove(field)
+    return list(data)
+
+
+def flatten_repeat_data(data):
+    result = []
+    index_tags=DEFAULT_INDEX_TAGS
```
pld, line 76: Codacy Issue found: [Exactly one space required around assignment](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5447927)

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: cool.. I yielded the rows  and then used this genexp to determine the results

onadata/apps/api/tests/viewsets/test_open_data_viewset.py
WinnyTroy: Sure. changed that.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +17,92 @@
...
+        # Yield the rows
+        rows = (row for row in data)
+        for row in rows:
+            # Include fields from the column headers
+            diff = set(tableau_headers).difference(set(row))
+            left_out_fields = {
+                field: None for field in diff}
+            row.update(left_out_fields)
+            result.append(row)
+    yield result
```
DavisRayM, line 60: This is a bit nit-picky but I don't see why we should have `result` here.... We don't really need to return a list with one row incremented I believe ?

```suggestion
    if data:
        headers = xform.get_headers()
        tableau_headers = remove_metadata_fields(headers)
        # Yield the rows
        rows = (row for row in data)
        for row in rows:
            # Include fields from the column headers
            diff = set(tableau_headers).difference(set(row))
            left_out_fields = {
                field: None for field in diff}
            row.update(left_out_fields)
            yield row
```

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +17,92 @@
...
+                                index=index,
+                                close_tag=index_tags[1])] + [
+                                    nested_key.split('/')[1]]
+                        xpaths = "/".join(xpaths)
+                        flat_dict.update({xpaths: nested_val})
+            else:
+                flat_dict.update({key: value})
+        # Add each row to resultant data
+        result.append(flat_dict)
+    return result
```
DavisRayM, line 103: Suggestion: Lets combine clean_data_fields and flatten_repeat_data into one function. That way we only loop through row data once, cleaning and flattening the data where needed. We would also just have one generator function in this case which might be good ?

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +17,92 @@
...
+    Loop through instance data and flatten repeat data
+    given the key "children/details" and nested_key/
+    abbreviated xpath "children/details/immunization/polio_1",
+    generate ["children", index, "immunization/polio_1"]
+    """
+    result = []
+    index_tags = DEFAULT_INDEX_TAGS
+    for record in data:
+        flat_dict = {}
+        for (key, value) in viewitems(record):
```
DavisRayM, line 84: I'm curious what does `viewitems` do ?

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,85 @@
...
+                                    index=index,
+                                    close_tag=index_tags[1])] + [
+                                        nested_key.split('/')[1]]
+                            xpaths = "/".join(xpaths)
+                            flat_dict.update({xpaths: nested_val})
+                    row.pop(key)
+                else:
+                    flat_dict.update({key: value})
+            # Include fields not present within the instance
+            {
```
pld, line 81: to avoid 81-83, can you define `flat_dict` to be the rhs of `left_out_fields` on line 56, removing line 55

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,85 @@
...
+
+
+def remove_metadata_fields(data):
+    """
+    Clean out unneccessary metadata fields
+    """
+    for field in METADATA_FIELDS:
+        if field in data:
+            data.remove(field)
+    return list(data)
```
pld, line 95: why the `list` conversion?

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -84,10 +147,12 @@ def append_to_tableau_colulmn_headers(header, question_type=None):
                 'dataType': quest_type,
                 'alias': header
             })
+        # Remove metadata fields from the column headers
+        xform_headers = set(remove_metadata_fields(self.xform_headers))
```
pld, line 151: when does `xform_headers` return duplicates?

onadata/libs/serializers/data_serializer.py
```diff
@@ -110,6 +111,27 @@ def to_representation(self, instance):
...
+    class Meta:
+        model = Instance
+        fields = ('json', )
+
+    def to_representation(self, instance):
+        ret = super(TableauDataSerializer, self).to_representation(instance)
+        if 'json' in ret:
+            ret = ret['json']
+            # Remove metadata fields from the instance
+            for field in METADATA_FIELDS:
```
pld, line 132: this can call the function you've defined, but to avoid circular imports you'll have to move that function to a utils namespace

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,85 @@
...
+                    flat_dict.update({key: value})
+            # Include fields not present within the instance
+            {
+                flat_dict.update({f: None}) for f in left_out_fields
+                if f not in flat_dict.keys()}
+            result.append(flat_dict)
+    return result
+
+
+def remove_metadata_fields(data):
```
pld, line 88: this should probably live in a utils namespace

onadata/libs/utils/common_tags.py
```diff
@@ -159,3 +159,26 @@
 XLS_DATE_FIELDS = ['date', 'today']
 SUBMISSION_REVIEW = 'submission_review'
 XLS_DATETIME_FIELDS = ['start', 'end', 'dateTime', '_submission_time']
+
+# Cache names used in the open_data viewet
```
pld, line 163: is this only related to the `open_data` viewset or are these all the general metadata fields? also typo in viewset

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: Changed that also. Thankyou

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,85 @@
...
+                                    key=nested_key.split('/')[0],
+                                    open_tag=index_tags[0],
+                                    index=index,
+                                    close_tag=index_tags[1])] + [
+                                        nested_key.split('/')[1]]
+                            xpaths = "/".join(xpaths)
+                            flat_dict.update({xpaths: nested_val})
+                    row.pop(key)
+                else:
+                    flat_dict.update({key: value})
```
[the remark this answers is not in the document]
WinnyTroy: I've changed that. Thankyou for this

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,85 @@
...
+    index_tags = DEFAULT_INDEX_TAGS
+    if data:
+        headers = xform.get_headers()
+        tableau_headers = remove_metadata_fields(headers)
+        for row in data:
+            diff = set(tableau_headers).difference(set(row))
+            left_out_fields = {
+                field: None for field in diff}
+            flat_dict = {}
+            for (key, value) in list(row.items()):
```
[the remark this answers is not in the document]
WinnyTroy: Actually Its no longer needed.
Removed that. Thanks for catching this

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,85 @@
...
+                            xpaths = [
+                                '{key}{open_tag}{index}{close_tag}'.format(
+                                    key=nested_key.split('/')[0],
+                                    open_tag=index_tags[0],
+                                    index=index,
+                                    close_tag=index_tags[1])] + [
+                                        nested_key.split('/')[1]]
+                            xpaths = "/".join(xpaths)
+                            flat_dict.update({xpaths: nested_val})
+                    row.pop(key)
```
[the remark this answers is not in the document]
WinnyTroy: Forgot to remove that. Thanks for catching this.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -16,22 +16,85 @@
...
+            for (key, value) in list(row.items()):
+                # capture and flatten repeat data here
+                # Loop through instance data and flatten repeat data
+                # given the key "children/details" and nested_key/
+                # abbreviated xpath "children/details/immunization/polio_1",
+                # generate ["children", index, "immunization/polio_1"]
+                if isinstance(value, list) and key not in [
+                        ATTACHMENTS, NOTES, GEOLOCATION]:
+                    for index, item in enumerate(value):
+                        index += 1
```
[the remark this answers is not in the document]
WinnyTroy: Yes, does make it much cleaner. Changed that. Thanks

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: Its not necessary at this time.
Had forgotten to remove that. I was previously using it to convert the set of column header fields back to a list.
Thanks for catching that

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: I noticed that there could be duplicates for group data(not repeat data)
if the `specify other` question type is used.

This can be observed when you run the `test_column_headers_endpoint` and place a debugger here.
The schema generated without removing duplicate columns is
```code
[
  'transport_available_transportation_types_to_referral_facility_ambulance',
  'transport_available_transportation_types_to_referral_facility_bicycle',
  'transport_available_transportation_types_to_referral_facility_boat_canoe',
  'transport_available_transportation_types_to_referral_facility_bus',
  'transport_available_transportation_types_to_referral_facility_donkey_mule_cart',
  'transport_available_transportation_types_to_referral_facility_keke_pepe',
  'transport_available_transportation_types_to_referral_facility_lorry',
  'transport_available_transportation_types_to_referral_facility_motorbike',
  'transport_available_transportation_types_to_referral_facility_taxi',
  'transport_available_transportation_types_to_referral_facility_other',
  'transport_available_transportation_types_to_referral_facility_other',
  'transport_loop_over_transport_types_frequency_ambulance_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_bicycle_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_boat_canoe_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_bus_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_donkey_mule_cart_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_keke_pepe_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_lorry_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_motorbike_frequency_to_referral_facility',
  'transport_loop_over_transport_types_frequency_taxi_frequency_to_referral_facility',
  'image1',
  'meta_instanceID',
  '_id'
]
```

This introduces an extra column, making column headers more than row headers.
This difference is what I am avoiding

onadata/libs/serializers/data_serializer.py
WinnyTroy: Much much cleaner. Appreciate this.

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: Moved this. Thankyou.

onadata/libs/utils/common_tags.py
WinnyTroy: It is comprised of the general metadata fields, yes.

I believe it is only the tableau functionality that removes metadata tags
So there is no where else that these tags are combined together in a list except here
I may however be wrong on this.

pld: OK, in that case I think you can remove this comment, the variable name is explanatory

onadata/apps/api/viewsets/open_data_viewset.py
pld: make sense, can you add a note to the comment explaining this, something like, "calling `set` to remove duplicates in group data"

onadata/libs/utils/opendata_tools.py
```diff
@@ -0,0 +1,13 @@
+from onadata.libs.utils.common_tags import METADATA_FIELDS
+
+def remove_metadata_fields(data):
```
pld, line 3: can you please move this to logger_tools instead of a new file, that seems like a nice more generic location for it

onadata/libs/utils/common_tags.py
WinnyTroy: Done

onadata/libs/utils/opendata_tools.py
WinnyTroy: Sure. Moved that.

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -13,25 +15,94 @@
...
+                field: None for field in diff}
+            for (key, value) in row.items():
+                if isinstance(value, list) and key not in [
+                        ATTACHMENTS, NOTES, GEOLOCATION]:
+                    for index, item in enumerate(value, start=1):
+                        # order repeat according to xform order
+                        item = get_ordered_repeat_value(key, item, index)
+                        flat_dict.update(item)
+                else:
+                    flat_dict.update({key: value})
```
pld, line 101: this should use assignment, since it's a single key and value

onadata/apps/api/viewsets/open_data_viewset.py
```diff
@@ -13,25 +15,94 @@
...
+                    data[xpaths] = nested_val
+        return data
+
+    result = []
+    if data:
+        headers = xform.get_headers()
+        tableau_headers = remove_metadata_fields(headers)
+        for row in data:
+            diff = set(tableau_headers).difference(set(row))
+            flat_dict = {
```
pld, line 91: instead of looping and assigning you can use a `defaultdict` with default value of None

WinnyTroy: I wasnt sure how to work with defaultdict for this, but I found another simple solution of using class Dict [.fromkeys](https://docs.python.org/2/library/stdtypes.html#dict.fromkeys) to unpack this set, assigning the default value

onadata/apps/api/viewsets/open_data_viewset.py
WinnyTroy: Sorry, changed that. Thanks

onadata/apps/api/viewsets/open_data_viewset.py
pld: yep good catch, defaultdict wouldn't have worked here, this is much better, you don't need to pass `None` since that's default but no harm in being explicit like you are

PR #1823, opened 2020-05-06
Introduce caching for user profile objects

### LINKED ISSUE #1821: Cache user profile on write
### Environmental Information
All
- Operating System: All
- Onadata version: v2.3.1
- Python version: 3.6

### Problem description
When a user profile is created or updated, it's not cached. This means subsequent requests to retrieve the user profile hit the database every time. I think we can improve performance by taking advantage of caching which already exists in onadata

### Expected behavior
Cache user profile after a write (creating and updating). 

### Additional Information
We can cache on creation and updating the profile and introduce `retrieve` function in `UserProfileViewSet` that would first check if there's data in cache before trying to retrieving the data from the database

WinnyTroy:
Cache user_profile object upon creation or update

Closes #1821

### DIFF
--- a/onadata/apps/api/viewsets/connect_viewset.py
+++ b/onadata/apps/api/viewsets/connect_viewset.py
@@ -25,6 +25,8 @@ from onadata.libs.serializers.project_serializer import ProjectSerializer
 from onadata.libs.serializers.user_profile_serializer import \
     UserProfileWithTokenSerializer
 from onadata.settings.common import DEFAULT_SESSION_EXPIRY_TIME
+from onadata.libs.utils.cache_tools import (cache,
+                                            USER_PROFILE_PREFIX)
 
 
 def user_profile_w_token_response(request, status):
@@ -42,8 +44,14 @@ def user_profile_w_token_response(request, status):
     try:
         user_profile = request.user.profile
     except UserProfile.DoesNotExist:
-        user_profile, _ = UserProfile.objects.get_or_create(
-            user=request.user)
+        user_profile = cache.get(
+            f'{USER_PROFILE_PREFIX}{request.user.username}')
+        if not user_profile:
+            user_profile, __ = UserProfile.objects.get_or_create(
+                user=request.user)
+            cache.set(
+                f'{USER_PROFILE_PREFIX}{request.user.username}',
+                user_profile)
 
     serializer = UserProfileWithTokenSerializer(
         instance=user_profile, context={"request": request})
--- a/onadata/apps/logger/views.py
+++ b/onadata/apps/logger/views.py
@@ -36,6 +36,8 @@ from onadata.apps.main.models import MetaData, UserProfile
 from onadata.libs.exceptions import EnketoError
 from onadata.libs.utils.decorators import is_owner
 from onadata.libs.utils.log import Actions, audit_log
+from onadata.libs.utils.cache_tools import (cache,
+                                            USER_PROFILE_PREFIX)
 from onadata.libs.utils.logger_tools import (
     BaseOpenRosaResponse, OpenRosaResponse, OpenRosaResponseBadRequest,
     PublishXForm, inject_instanceid, publish_form, remove_xform,
@@ -172,7 +174,11 @@ def formList(request, username):  # pylint: disable=C0103
     formList view, /formList OpenRosa Form Discovery API 1.0.
     """
     formlist_user = get_object_or_404(User, username__iexact=username)
-    profile, __ = UserProfile.objects.get_or_create(user=formlist_user)
+    profile = cache.get(
+        f'{USER_PROFILE_PREFIX}{formlist_user.username}')
+    if not profile:
+        profile, __ = UserProfile.objects.get_or_create(
+            user__username=formlist_user.username)
 
     if profile.require_auth:
         authenticator = HttpDigestAuthenticator()
@@ -231,7 +237,11 @@ def xformsManifest(request, username, id_string):  # pylint: disable=C0103
 
     xform = get_form(xform_kwargs)
     formlist_user = xform.user
-    profile, __ = UserProfile.objects.get_or_create(user=formlist_user)
+    profile = cache.get(
+        f'{USER_PROFILE_PREFIX}{formlist_user.username}')
+    if not profile:
+        profile, __ = UserProfile.objects.get_or_create(
+            user__username=formlist_user.username)
 
     if profile.require_auth:
         authenticator = HttpDigestAuthenticator()
--- a/onadata/libs/utils/cache_tools.py
+++ b/onadata/libs/utils/cache_tools.py
@@ -13,6 +13,7 @@ PROJ_OWNER_CACHE = "ps-project_owner-"
 
 # Cache names used in user_profile_serializer
 IS_ORG = "ups-is_org-"
+USER_PROFILE_PREFIX = "user_profile-"
 
 # Cache names user in xform_serializer
 XFORM_PERMISSIONS_CACHE = "xfs-get_xform_permissions"
--- a/onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
@@ -29,6 +29,7 @@ from onadata.apps.main.models.user_profile import \
 from onadata.libs.authentication import DigestAuthentication
 from onadata.libs.serializers.user_profile_serializer import \
     _get_first_last_names
+from onadata.libs.utils.cache_tools import USER_PROFILE_PREFIX, safe_key
 
 
 def _profile_data():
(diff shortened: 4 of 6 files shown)


onadata/libs/utils/cache_tools.py
```diff
@@ -41,3 +44,11 @@ def safe_delete(key):
 def safe_key(key):
     """Return a hashed key."""
     return hashlib.sha256(force_bytes(key)).hexdigest()
+
+def cache_single_object(key_prefix, id, query):
+    key = "".format(key_prefix, id)
```
pld, line 49: Codacy Issue found: [Too many arguments for format string](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5478081)

onadata/libs/utils/cache_tools.py
```diff
@@ -41,3 +44,11 @@ def safe_delete(key):
 def safe_key(key):
     """Return a hashed key."""
     return hashlib.sha256(force_bytes(key)).hexdigest()
+
+def cache_single_object(key_prefix, id, query):
```
pld, line 48: Codacy Issue found: [Redefining built-in 'id'](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5478081)

onadata/libs/utils/cache_tools.py
```diff
@@ -41,3 +44,11 @@ def safe_delete(key):
...
+    key = "".format(key_prefix, id)
+    object_list = cache.get(key, None)
+    if object_list is None:
+        object_list,_ = query
```
pld, line 52: Codacy Issue found: [Exactly one space required after comma](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5478081)

onadata/apps/main/views.py
```diff
@@ -1086,7 +1091,11 @@ def form_photos(request, username, id_string):
     image_urls = json.dumps(image_urls)
 
     data['images'] = image_urls
-    data['profile'], created = UserProfile.objects.get_or_create(user=owner)
+    data['profile'] = property(
+        lambda u: cache_single_object(
```
pld, line 1095: Codacy Issue found: [Unexpected keyword argument 'id' in function call](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5478081)

pld, line 1095: Codacy Issue found: [No value for argument 'userid' in function call](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5478081)

onadata/apps/api/permissions.py
```diff
@@ -513,7 +515,11 @@ def has_permission(self, request, view):
             else:
                 # Raises a permission denied exception, forces authentication
                 return False
-            profile, _ = UserProfile.objects.get_or_create(user=user)
+            profile = cache_single_object(
+                USER_PROFILE_PREFIX,
+                user.id,
+                UserProfile.objects.get_or_create(user=user)
+                )
```
DavisRayM, line 522: This would possibly also cache a User who is already present in the database right ? Not entirely sure if we need to cache a user that's already retrievable from the db in this case.... It might end up in a scenario where we'll always be caching all users who make authenticated submissions.

_I think we should only cache here if we're creating a Profile_

https://github.com/onaio/onadata/pull/1823/files#diff-356ba073bddf0df5ba9d9c050982939eR47-R51
https://github.com/onaio/onadata/pull/1823/files#diff-cec4d5661d03adabc14e2f6efc04e7b6R177-R181
https://github.com/onaio/onadata/pull/1823/files#diff-cec4d5661d03adabc14e2f6efc04e7b6R240-R243

Same case for the above ^ _I guess an easy change would be to make the `cache_single_object` function capable of knowing whether it should cache or not ? Possibly checking if the query returns something_

onadata/apps/api/viewsets/openid_connect_viewset.py
```diff
@@ -159,7 +161,12 @@ def create_or_get_user(
                                                    'username': username
                                                })
     if created:
-        UserProfile.objects.create(user=user)
+        property(
+            lambda u: cache_single_object(
+                USER_PROFILE_PREFIX,
+                user.id,
+                UserProfile.objects.get_or_create(user=user)
+                ))
```
DavisRayM, line 169: Why did we have to use a lamba function & property here ?

onadata/apps/logger/views.py
```diff
@@ -505,7 +521,12 @@ def enter_data(request, username, id_string):
     except EnketoError as e:
         data = {}
         owner = User.objects.get(username__iexact=username)
-        data['profile'], __ = UserProfile.objects.get_or_create(user=owner)
+        data['profile'] = property(
+            lambda u: cache_single_object(
+                USER_PROFILE_PREFIX,
+                owner.id,
+                UserProfile.objects.get_or_create(user=owner)
+                ))
```
DavisRayM, line 529: Why the lamba and property ?

onadata/apps/api/permissions.py
WinnyTroy: The idea was to avoid requests hitting the database for subsequent requests after a user profile is created. So, even for users who are saved in the db, in the first try to retrieve the UserProfile would hit the db, but subsequent requests wouldnt necessarily have to.

I cached the UserProfile objects here since at this point there are database requests to retrieve the UserProfile object for all these users, wouldn't we want to avoid these requests constantly hitting the db?

At the moment, the `cache_single_object` function checks if the user_profile cache for the user has been created, if not, then it makes the query. I was avoiding making requests to the db until absolutely necessary

onadata/apps/api/viewsets/openid_connect_viewset.py
WinnyTroy: An attempt to make it a getter function. I can remove them

onadata/libs/utils/cache_tools.py
```diff
@@ -41,3 +44,12 @@ def safe_delete(key):
...
     return hashlib.sha256(force_bytes(key)).hexdigest()
+
+
+def cache_single_object(key_prefix, userid, query):
+    key = "{}{}".format(key_prefix, userid)
+    object_list = cache.get(key, None)
+    if object_list is None:
+        object_list, _ = query
+        cache.set(key, object_list)
+    return object_list
```
ivermac, line 55: I think this function achieves the purpose of cache. I've noticed we are passing `query` as an arg and using it to set the cache object if doesn't exist. This means we'll always querying the db to retrieve the `query` object then passing it to this function, right? if that's the case, I guess that beats the purpose of caching in the first place, right? We would like to retrieve data from cache first and if doesn't exist from cache, we retrieve the data from database

onadata/apps/api/viewsets/openid_connect_viewset.py
DavisRayM: Mhh okay... I'm a bit confused on why we need the getter function ?

onadata/apps/api/permissions.py
DavisRayM: Okay.... but still this will make the request wont it ? `UserProfile.objects.get_or_create(user=user)` would make the db request as such it doesn't really decrease the amounts of request to the db....

WinnyTroy: I understand now..cool cool, i'll restructure that abit

onadata/apps/api/viewsets/connect_viewset.py
```diff
@@ -42,8 +44,11 @@ def user_profile_w_token_response(request, status):
     try:
         user_profile = request.user.profile
     except UserProfile.DoesNotExist:
-        user_profile, _ = UserProfile.objects.get_or_create(
-            user=request.user)
+        user_profile = cache.get(
+            f'{USER_PROFILE_PREFIX}{request.user.username}')
+        if not user_profile:
+            user_profile, __ = UserProfile.objects.get_or_create(
+                user=request.user)
```
ivermac, line 51: Should we cache the profile here as well after retrieving it? so that we don't have the db again if a new requests is made before the cache expires.

ivermac: This just means that `user_profile` won't be empty and we won't have to hit the db

WinnyTroy: Made this change.

onadata/apps/api/viewsets/connect_viewset.py
```diff
@@ -42,8 +44,15 @@ def user_profile_w_token_response(request, status):
...
+
+    cache.set(
+        f'{USER_PROFILE_PREFIX}{request.user.username}',
+        user_profile)
```
ivermac, line 55: I think we want to move this inside the `if` block after line 50 and 51 when we have retrieved the `user_profile` object. As it is at the moment, we are setting the cache value every time this function is called.

WinnyTroy: Yeah, moved that. Thanks

PR #1825, opened 2020-05-07
Update Enketo kwargs arguments

### LINKED ISSUE #1824: Enketo data submission requires login
## Problem description
Enketo links require a user to log in to submit data even though require authentication is turned off on the user's setting page.

### Expected behavior
A user should be able to submit data without logging in, if require authentication is turned off

WinnyTroy:
closes #1824

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
@@ -352,7 +352,7 @@ class TestOrganizationProfileViewSet(TestAbstractViewSet):
         request = self.factory.get('/', **self.extra)
         response = view(request)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.data, [expected_data])
+        self.assertEqual(dict(response.data[0]), expected_data)
 
     def test_role_for_org_non_owner(self):
         # creating org with member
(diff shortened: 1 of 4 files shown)


DavisRayM:
Could we also add a test for this scenario... A bit scary that we didn't notice that passing `form_pk` there would not work.

Ideally the test would ensure anonymous users can make submissions via the enketo url

onadata/apps/api/tests/viewsets/test_organization_profile_viewset.py
```diff
@@ -352,7 +352,7 @@ def test_member_sees_orgs_added_to(self):
         request = self.factory.get('/', **self.extra)
         response = view(request)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.data, [expected_data])
+        self.assertEqual(dict(response.data[0]), expected_data)
```
DavisRayM, line 355: Why did we have to change this ?

WinnyTroy: yeah for some reason, the response_data returned an OrderedDict object.
I didnt see the correlation with the changes made however.

DavisRayM: Mhh okay, was the test failing because of it ?

WinnyTroy: lemme confirm this

WinnyTroy: For clarification, this change occurs because as a result of a  sorting issue and does not fail because of the changes to the enketo kwargs

PR #1830, opened 2020-06-11
Include preview domain

WinnyTroy:
This allows generating enketo urls for the preview api instance.

### DIFF
--- a/onadata/apps/main/management/commands/create_enketo_express_urls.py
+++ b/onadata/apps/main/management/commands/create_enketo_express_urls.py
@@ -6,6 +6,7 @@ from onadata.apps.logger.models import XForm
 from onadata.libs.utils.model_tools import queryset_iterator
 from onadata.libs.utils.viewer_tools import (
     enketo_url, get_enketo_preview_url, get_form_url)
+from onadata.libs.utils.common_tags import APP_DOMAIN
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


onadata/apps/main/management/commands/create_enketo_express_urls.py
```diff
@@ -38,7 +38,11 @@ def handle(self, *args, **options):
             raise CommandError(
                 'please provide a server_name, a server_port and a protocol')
 
-        if server_name not in ['api.ona.io', 'stage-api.ona.io', 'localhost']:
+        if server_name not in [
+                'api.ona.io',
+                'stage-api.ona.io',
+                'localhost',
+                'api.preview.ona.io']:
```
DavisRayM, line 45: I think it would be better if we introduced a setting like `APP_DOMAIN` that has the value that should be used in this function... I don't believe we ever need to create enketo express urls for other domains other than the one the application is running on... thought ?

DavisRayM: Would also ensure that this command is able to run on any domain

WinnyTroy: cool cool....sounds cleaner

onadata/apps/main/management/commands/create_enketo_express_urls.py
```diff
@@ -38,7 +39,7 @@ def handle(self, *args, **options):
             raise CommandError(
                 'please provide a server_name, a server_port and a protocol')
 
-        if server_name not in ['api.ona.io', 'stage-api.ona.io', 'localhost']:
+        if server_name not in APP_DOMAIN:
```
DavisRayM, line 42: Lets make `APP_DOMAIN` just a string. I don't believe there is a scenario we'll need more than one domain.

```suggestion
        if server_name == APP_DOMAIN:
```

_Actually if possible let's not limit the `server_name`... I think making it a required argument might be better ? Thoughts ?_

WinnyTroy:
Closed in #1857

PR #1831, opened 2020-06-12
Use last name as first name if first name isn't sent

DavisRayM:

- During the Open ID Connect authentication process use the last name as the first name if not retrieved from token

- Updated tests

### DIFF
--- a/onadata/apps/api/viewsets/openid_connect_viewset.py
+++ b/onadata/apps/api/viewsets/openid_connect_viewset.py
@@ -126,8 +126,10 @@ class OpenIDConnectViewSet(viewsets.ViewSet):
                     return Response(
                         data, template_name='missing_oidc_detail.html')
 
-                first_name = claim_values.get(FIRST_NAME)
                 last_name = claim_values.get(LAST_NAME)
+                first_name = claim_values.get(FIRST_NAME)
+                if not first_name:
+                    first_name = last_name
                 user = create_or_get_user(first_name, last_name, email,
                                           username)
         else:
--- a/onadata/apps/api/viewsets/openid_connect_viewset.py
+++ b/onadata/apps/api/viewsets/openid_connect_viewset.py
@@ -1,4 +1,5 @@
 import secrets
+import json
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


onadata/apps/api/viewsets/openid_connect_viewset.py
```diff
@@ -126,8 +126,10 @@ def callback(self, request, **kwargs):  # pylint: disable=no-self-use
                     return Response(
                         data, template_name='missing_oidc_detail.html')
 
-                first_name = claim_values.get(FIRST_NAME)
                 last_name = claim_values.get(LAST_NAME)
+                first_name = claim_values.get(FIRST_NAME)
+                if not first_name:
+                    first_name = last_name
```
ukanga, line 132: What about the case when we do not have the `last_name` and the `first_name`?

DavisRayM: Mhh I'll implement functionality to fail the authentication process(_I think this was the agreed upon approach ?_) if both fields are missing.

ukanga: Yes.

onadata/apps/api/viewsets/openid_connect_viewset.py
```diff
@@ -129,7 +129,10 @@ def callback(self, request, **kwargs):  # pylint: disable=no-self-use
                 last_name = claim_values.get(LAST_NAME)
                 first_name = claim_values.get(FIRST_NAME)
                 if not first_name:
-                    first_name = last_name
+                    if last_name:
+                        first_name = last_name
+                    else:
+                        return HttpResponseBadRequest()
```
ukanga: Let's have a descriptive error message. The pattern established in this file should be something we should change direction by returning descriptive messages.

Probably also make it a bit terse:
```python
if not first_name and not last_name:
    raise HttpResponseBadRequest(json.dumps({"error": _("Missing required fields: %s, %s." % (FIRST_NAME, LAST_NAME)}))
else:
    first_name = first_name or last_name
```

PR #1834, opened 2020-06-15
Add enketo_meta_usename cookie

### LINKED ISSUE #1832: Set cookie `__enketo_meta_username`
# Problem description
Enketo currently get's the username metadata from a cookie called `__enketo_meta_username`.
This cookie is not set by onadata. 

A similar cookie that is currently set is `__enketo_meta_uid` that contains the username

Create a form with a username field and then use enketo to collect data for the form.
Notice that the username field is set to *'username not found'*

FrankApiyo:
Add `__enketo_meta_username` cookie with the same value as the `enketo_meta_uid` cookie

Closes #1832


faith-mutua:
@FrankApiyo @DavisRayM This hasn't been fixed yet. I am still getting `username not found` under the username field on stage.ona.io.

PR #1844, opened 2020-07-06
Delete User Functionality Updates

### LINKED ISSUE #1843: Delete Users Functionality Updates
### Environmental Information
Production

### Problem description
Once a user has been soft deleted,the email address initially used by the user is still associated with the inactive user account. This means that the user cannot use this email in creating another account

### Expected behavior
Once a user has been soft deleted, they should be able to create a user account with the same email and username they had initially.

### Additional Information
Also need to include a crontab to delete users periodically

WinnyTroy:
Included deletion suffix to email associated with the user account
Included cron task to periodically perform task after 30 days

Closes #1843

### DIFF
--- a/onadata/apps/api/tests/management/commands/test_delete_users.py
+++ b/onadata/apps/api/tests/management/commands/test_delete_users.py
@@ -38,6 +38,9 @@ class DeleteUserTest(TestBase):
             "User bruce deleted successfully.",
             out.getvalue())
 
+        with self.assertRaises(User.DoesNotExist):
+            User.objects.get(email="bruce@gmail.com")
+
     @mock.patch(
         "onadata.apps.api.management.commands.delete_users.input")
     def test_delete_users_no_input(self, mock_input):  # pylint: disable=R0201
--- a/onadata/apps/api/management/commands/delete_users.py
+++ b/onadata/apps/api/management/commands/delete_users.py
@@ -60,6 +60,7 @@ def inactivate_users(users, user_input):
                     # to make the initial username available
                     deletion_suffix = timezone.now().strftime('-deleted-at-%s')
                     user.username += deletion_suffix
+                    user.email += deletion_suffix
 
                     user.save()
                     sys.stdout.write(
--- a/onadata/apps/api/tasks.py
+++ b/onadata/apps/api/tasks.py
@@ -64,9 +64,12 @@ def delete_xform_async(xform_id, user_id):
     xform.soft_delete(user)
 
 
+@task()
 def delete_user_async():
     """Delete inactive user accounts"""
-    users = User.objects.filter(active=False)
+    users = User.objects.filter(active=False,
+                                username__contains="deleted-at",
+                                email__contains="deleted-at")
     for user in users:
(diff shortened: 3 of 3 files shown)


onadata/settings/common.py
```diff
@@ -509,6 +510,31 @@ def configure_logging(logger, **kwargs):
...
+        'options': {'ignore_result': True}
+    },
+    'Delete-Inactive-Users': {
+        'task': 'onadata.apps.api.tasks.delete_user_async',
+        'schedule': crontab(
+            0, 0, day_of_month='30'),  # Execute on the 30th of every month.
+        'options': {'ignore_result': True}
+    },
+}
+
```
DavisRayM, line 537: I don't think this configuration should be placed within the `common` settings..... Celery beat scheduling seems like something the user should configure within their own local setting IMO

onadata/settings/common.py
```diff
@@ -509,6 +510,31 @@ def configure_logging(logger, **kwargs):
...
+    },
+    'celery.backend_cleanup': {
+        'task': 'celery.backend_cleanup',
+        'schedule': crontab(hour='*', minute='30'),
+        'options': {'ignore_result': True}
+    },
+    'Delete-Inactive-Users': {
+        'task': 'onadata.apps.api.tasks.delete_user_async',
+        'schedule': crontab(
+            0, 0, day_of_month='30'),  # Execute on the 30th of every month.
```
ivermac, line 533: @WinnyTroy What about the month of February? Can we consider may be the 1st day of every month?

ivermac: I agree with @DavisRayM on having the configuration on a user's local settings file

ivermac: That said, I like the fact that functionality will be scheduled and executed once a month

onadata/settings/common.py
WinnyTroy: Places this here since the tasks are things that should run in all environments/deployments of onadata

Im unsure which settings file would be a better fit for this configurations.
Cc: @ivermac

onadata/settings/common.py
WinnyTroy: Alright, sure, I'll remove this configurations from here. 
And I see no problem with having this run on the 1st of every month. Let me change that

onadata/apps/api/tests/management/commands/test_delete_users.py
```diff
@@ -56,3 +57,30 @@ def test_delete_users_no_input(self, mock_input):  # pylint: disable=R0201
...
+            'delete_users',
+            user_details=new_user_details,
+            user_input='True',
+            stdout=out)
+
+        self.assertEqual(
+            "User Ken deleted successfully.",
+            out.getvalue())
+
+        with self.assertRaises(ObjectDoesNotExist):
```
DavisRayM, line 85: Tiny thing/change.... Please use `User.DoesNotExist`

WinnyTroy: Cool cool. Changed that.

PR #1847, opened 2020-07-16
Add management command to increase an ODKTokens lifetime

### LINKED ISSUE #1747: Add a management command to elongate the expiry date of an ODKToken
### Environmental Information

- Operating System: All
- Onadata version: master
- Python version: Python3.6+

### Feature description

In some scenarios administrators may want to elongate the lifetime of an ODK Token. Adding a way to easily do so through a management command for a particular user would be nice.

### Expected behavior of the new feature

- One should be able to run a command like `./manage.py elongate_odk_token <username> <amount of days>`

DavisRayM:

- Added `expires` date-time field for ODKToken class
- Added a management command `increase_odk_token_lifetime`

- Added tests

- All `ODKTokens` created before this change will not have the `expires` date set. Which might cause issues. _All ODKTokens should be invalidated/regenerated._ 

Closes #1747


onadata/apps/api/models/odk_token.py
```diff
@@ -3,12 +3,13 @@
 """
 import binascii
 import os
-from datetime import timedelta
+from datetime import timedelta, datetime
```
pld, line 6: Codacy found an issue: [Unused datetime imported from datetime](https://app.codacy.com/manual/Ona/onadata/pullRequest?prid=5862118)

onadata/apps/api/models/odk_token.py
```diff
@@ -99,5 +94,16 @@ def _post_save_persist_partial_digests(sender, instance=None, **kwargs):
         _persist_partial_digests(instance.user)
 
 
+def _post_save_set_expiry_date(sender, instance=None, **kwargs):
+    if instance and not instance.expires:
+        instance.expires = instance.created + timedelta(
+            days=ODK_TOKEN_LIFETIME)
+        instance.expires = instance.expires.astimezone(
+            instance.created.tzinfo)
```
ivermac, line 102: Why are we overriding `instance.expires` in line 101? Also, will `instance.expires` in line 99 still have the value assigned to it even though it hasn't been saved and it's being used in line 101? Let me know if my questions make sense

onadata/libs/utils/logger_tools.py
```diff
@@ -112,10 +112,41 @@ def _get_instance(xml, new_uuid, submitted_by, status, xform, checksum):
...
+                try:
+                    form = XForm.objects.filter(
+                        id_string__iexact=form_id,
+                        user__username__iexact=username,
+                        deleted_at__isnull=True).first()
+                    root = form.survey.name
+                except XForm.DoesNotExist:
+                    root = 'data'
+        else:
+            root = 'data'
```
ivermac, line 147: Why have we changed the value assigned to `root` from `form_id` (which seems to be an xform object) to a string value (`data`)? Also, i'm not very conversant with this part of the code. I'd like @ukanga's eye on this change as well

DavisRayM: My bad, these are changes meant for another pull request. I didn't notice the commits were rebased into this one. I'll update the PR

onadata/apps/api/models/odk_token.py
DavisRayM: Overriding the value to make sure that the datetime object saved in the field is timezone aware(_Usually Django saves it's `datetime` objects with the UTC timezone when Timezones are enabled. Didn't want to deviate from that_)

The instance will still have the `expires` value assigned to the object in line 99 and the value will be accessible in line 101, I believe. I'll update this though and set the `expires` field in one line.

PR #1890, opened 2020-09-03
Organization profile caching fixes

FrankApiyo:
- Add `is_org` SerializerMethodField to OrganizationSerializer
- Checked for cached organization profile in `retrieve` method of `UserProfileViewSet`
- Tested locally
Closes https://github.com/onaio/onadata/issues/1898

### DIFF
--- a/onadata/apps/api/viewsets/user_profile_viewset.py
+++ b/onadata/apps/api/viewsets/user_profile_viewset.py
@@ -37,7 +37,8 @@ from onadata.libs.utils.email import (get_verification_email_data,
 from onadata.libs.utils.cache_tools import (safe_delete,
                                             CHANGE_PASSWORD_ATTEMPTS,
                                             LOCKOUT_CHANGE_PASSWORD_USER,
-                                            USER_PROFILE_PREFIX)
+                                            USER_PROFILE_PREFIX,
+                                            ORG_PROFILE_CACHE)
 from onadata.libs import filters
 from onadata.libs.utils.user_auth import invalidate_and_regen_tokens
 from onadata.libs.mixins.authenticate_header_mixin import \
@@ -210,6 +211,8 @@ class UserProfileViewSet(
         """ Get user profile from cache or db """
         username = kwargs.get('user')
         cached_user = cache.get(f'{USER_PROFILE_PREFIX}{username}')
+        if not cached_user:
+            cached_user = cache.get(f'{ORG_PROFILE_CACHE}{username}')
         if cached_user:
             return Response(cached_user)
         response = super(UserProfileViewSet, self)\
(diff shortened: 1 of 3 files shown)


onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -210,6 +211,8 @@ def retrieve(self, request, *args, **kwargs):
         """ Get user profile from cache or db """
         username = kwargs.get('user')
         cached_user = cache.get(f'{USER_PROFILE_PREFIX}{username}')
+        if not cached_user:
+            cached_user = cache.get(f'{ORG_PROFILE_CACHE}{username}')
```
WinnyTroy, line 215: Why are we introducing this block here?
Isnt this being done in line 213?

WinnyTroy (changes requested):

WinnyTroy (approved):

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -210,6 +211,8 @@ def retrieve(self, request, *args, **kwargs):
         """ Get user profile from cache or db """
         username = kwargs.get('user')
         cached_user = cache.get(f'{USER_PROFILE_PREFIX}{username}')
```
DavisRayM, line 213:
```suggestion
        cached_user = cache.get(f'{USER_PROFILE_PREFIX}{username}') or cache.get(f'{ORG_PROFILE_CACHE}{username}')
```

onadata/libs/serializers/organization_serializer.py
```diff
@@ -32,6 +32,7 @@ class OrganizationSerializer(serializers.HyperlinkedModelSerializer):
     creator = serializers.HyperlinkedRelatedField(
         view_name='user-detail', lookup_field='username', read_only=True)
     users = serializers.SerializerMethodField()
+    is_org = serializers.SerializerMethodField()
```
DavisRayM, line 35: What's the need for the `is_org` field? Since this is the Organization serializer... I'd expect this to only be used for organizations....? Is that not the case?

onadata/apps/api/viewsets/user_profile_viewset.py
DavisRayM: https://github.com/onaio/onadata/pull/1890/files#diff-c857800a61f0303b9833f6aa94d239e1R220 Do we need to also set the Org Profile cache here? Could we just have this cached using a single prefix instead i.e `ACCOUNT_PROFILE_PREFIX` ?

DavisRayM (changes requested):
https://github.com/onaio/onadata/issues/1898#issuecomment-692593021

FrankApiyo:
It seems we do not need this PR because the issue it fixes does not occur anymore

PR #1895, opened 2020-09-10
Update submission metrics collection

### LINKED ISSUE #1891: Collect project information and differentiate ODK and Enketo Submissions
### Feature request

Currently, our analytics functionality is tracking the form, user & organization information of each submission it would be awesome if we also collected project information and  if we differentiated ODK Submissions and Enketo submissions

WinnyTroy:
### Changes Introduced
Introduce collecting project information for every submission created
Differentiate submission sources for each submission.
Reconstruct the context object being sent to segment.
Introduce tracking of newly created forms, projects and user profile accounts

- [x] Confirming data sent to data analytics is as expected
- [x] Updating Tests

Unsure of how the context changes may appear on google analytics

Closes #1891


onadata/libs/utils/analytics.py
```diff
@@ -172,12 +207,11 @@ def decorator(obj, *args):
 def track(user, event_name, properties=None, context=None, request=None):
     """Record actions with the track() API call to the analytics agents."""
     if _segment or appoptics_api:
-        context = context or {}
-        context['source'] = settings.HOSTNAME
```
DavisRayM, line 176: Why are we no longer tracking the `source`?

onadata/libs/utils/analytics.py
```diff
@@ -143,12 +154,36 @@ def get_event_label(self) -> str:
...
         self.set_user()
         event_name = self.get_event_name()
         label = self.get_event_label()
         tracking_properties = self.get_tracking_properties(label=label)
+        try:
+            if tracking_properties['from'] == 'XML Submissions':
+                # Only introduce an `event_source` field for submissions
+                # created xml. This helps differentiate Enketo and ODK Collect
+                # instances. Otherwise, use the `from` tag in properties object
+                tracking_properties = self.get_request_origin(
```
DavisRayM, line 183: I still don't see the need to have both the `from` and `event_source` fields... Both refer to the same aspect of an event. I'd suggest we use the `from` field as that's already configured in Segment

DavisRayM (changes requested):

onadata/libs/utils/analytics.py
WinnyTroy: Changed this, remained with the `event_source` tag.

onadata/libs/utils/analytics.py
WinnyTroy: It appears with the `campaign` [field](https://github.com/onaio/onadata/pull/1895/files#diff-9ddcf8387d802e0eed517864f979f2ff08dd99dbfabf287382548c1ef5771bfeR228) that segment configured.

WinnyTroy: There is no `source` field known to Segment. The field to best represent this information is in the `campaign` object that we defined [here](https://github.com/onaio/onadata/pull/1895/files#diff-9ddcf8387d802e0eed517864f979f2ff08dd99dbfabf287382548c1ef5771bfeR194)

onadata/apps/api/tests/viewsets/test_abstract_viewset.py
```diff
@@ -244,6 +251,13 @@ def _publish_xls_form_to_project(self, publish_data={}, merge=True,
                 post_data = {'xls_file': xls_file}
                 request = self.factory.post(
                     '/', data=post_data, **self.extra)
+                request.META['HTTP_DATE'] = '2020-09-10T11:56:32.424726+00:00'
+                request.META['HTTP_REFERER'] = settings.HOSTNAME +\
+                    f':8000'
+                request.META['HTTP_USER_AGENT'] =\
+                    'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit'\
+                    '/537.36 (KHTML, like Gecko) Chrome'\
+                    '/83.0.4103.61 Safari/537.36'
```
DavisRayM, line 260: Why was this necessary to add?

WinnyTroy: Since we introduced tracking newly created forms and projects, a KeyError is raised when making this request on the test.
I'd figure real time requests would have this already set, so I updated the test to include these header variables

DavisRayM: Mihm instead of adding this in, could we change [these lines](https://github.com/onaio/onadata/pull/1895/files#diff-9ddcf8387d802e0eed517864f979f2ff08dd99dbfabf287382548c1ef5771bfeR192-R197) to use `.get()`? In some cases, the requests may not have those values I believe...

onadata/libs/tests/utils/test_analytics.py
```diff
@@ -31,7 +32,7 @@ def test_get_user_id(self):
         self.assertTrue(len(user2.email) > 0)
         self.assertEqual(get_user_id(user2), user2.email)
 
-    @override_settings(SEGMENT_WRITE_KEY='123', HOSTNAME='test-server')
+    @override_settings(SEGMENT_WRITE_KEY='123')
```
DavisRayM, line 35: Could we add back the `HOSTNAME` setting? Would be nice to see how the Analytics payload looks with it.

DavisRayM (approved):

ukanga (approved):

ivermac (approved):

PR #1938, opened 2020-10-29
Check submission encryption status before Instance creation

### LINKED ISSUE #1934: Restrict unencrypted submissions from being made to an encrypted form
### Environmental Information

- Onadata version: master

### Problem description

Currently, unencrypted submissions can be made to forms that are encrypted. We should stop users from being able to do so and require all encrypted form submissions to be encrypted and vice-versa.

### Expected behavior

Encrypted forms should only receive encrypted submissions and Unencrypted forms should only accept unencrypted submissions

1. Create an unencrypted form
2. Draft a submission for the form
3. Update the form & encrypt it
4. Submit the draft unencrypted submission

### Additional Information

Related to #1763

DavisRayM:

- Check submission encryption status and confirm it matches the XForms encryption status before creation.

- Added Tests

- Encrypted submissions will now be rejected when they are made to unencrypted forms and vice versa(unencrypted submission -> encrypted form)

Closes #1934


ivermac (dismissed):

onadata/libs/utils/logger_tools.py
```diff
@@ -275,6 +277,25 @@ def check_submission_permissions(request, xform):
...
+    submission_element = ET.fromstring(xml)
+    encryption_status = submission_element.attrib.get('encrypted') == "yes"
+
+    if encryption_status != xform.encrypted:
+        form_status = "encrypted" if xform.encrypted else "unencrypted"
+        submisison_status = "Encrypted" if encryption_status else "Unencrypted"
+        raise InstanceEncryptionError(
+            f"{submisison_status} submissions are not"
+            f" allowed for {form_status} forms.")
+
```
ukanga, line 297: Can we use translation strings here to allow this error message to be translated. And please remember to run the command for updating translation files,

onadata/libs/utils/logger_tools.py
```diff
@@ -468,6 +490,8 @@ def safe_create_instance(username, xml_file, media_files, uuid, request):
     except InstanceEmptyError:
         error = OpenRosaResponseBadRequest(
             _(u"Received empty submission. No instance was created"))
+    except InstanceEncryptionError as e:
+        error = OpenRosaResponseBadRequest(e)
```
ukanga, line 494: Would you also not need to apply `text(e)` as is the pattern in this section, see the trend below where the error variable `e` is concerned.

onadata/libs/utils/logger_tools.py
```diff
@@ -275,6 +277,25 @@ def check_submission_permissions(request, xform):
...
+def check_submission_encryption(xform: XForm, xml: bytes) -> NoReturn:
+    """
+    Check that the submission is encrypted or unencrypted depending on the
+    encryption status of an XForm.
+
+    The submission is invalid if the XForm's encryption status is different
+    from the submissions
+    """
+    submission_element = ET.fromstring(xml)
+    encryption_status = submission_element.attrib.get('encrypted') == "yes"
```
ukanga, line 289: I wonder, is this the only attribute that we need to check to determine a submission is encrypted? Should we also confirm presence of or lack of `base64EncryptedKey` and `encryptedXmlFile`.

ukanga (changes requested):

onadata/libs/utils/logger_tools.py
DavisRayM: All three [attributes are required](https://getodk.github.io/xforms-spec/encryption) in a properly built Encrypted submission; _Initially thought checking for just one would be okay. But, now I believe we can check if all three are present and also add functionality to determine whether the encrypted submission is valid_

I'll add these checks and re-request a review

onadata/libs/utils/logger_tools.py
```diff
@@ -275,6 +280,37 @@ def check_submission_permissions(request, xform):
...
+    encryption_status = False
+    submission_element = ET.fromstring(xml)
+    encrypted_attrib = submission_element.attrib.get('encrypted')
+    required_encryption_elems = [
+        elem.tag for elem in submission_element
+        if elem.tag in REQUIRED_ENCRYPTED_FILE_ELEMENTS]
+
+    # Check the validity of the submission
+    if encrypted_attrib == "yes" or len(required_encryption_elems) > 1:
+        if (not len(required_encryption_elems) == 2 or
```
ukanga, line 300: The expression `len(required_encryption_elems)` is used more than once here, can we assign it to variable and use the variable instead?

onadata/libs/utils/logger_tools.py
```diff
@@ -275,6 +280,37 @@ def check_submission_permissions(request, xform):
...
+                not encrypted_attrib == "yes") and xform.encrypted:
+            raise InstanceFormatError(
+                "Encrypted submission incorrectly formatted.")
+        encryption_status = True
+
+    if encryption_status != xform.encrypted:
+        form_status = "encrypted" if xform.encrypted else "unencrypted"
+        submisison_status = "Encrypted" if encryption_status else "Unencrypted"
```
ukanga, line 308: Since this is part of output to a user, you probably want to use the translated values.

Again, the only situation that matters is the "encrypted" form, if the form is not encrypted there is no need to reject the submission on this basis.

ukanga (changes requested):

onadata/libs/utils/logger_tools.py
DavisRayM: > Again, the only situation that matters is the "encrypted" form, if the form is not encrypted there is no need to reject the submission on this basis.

How about in a situation where an encrypted submission is made to an unencrypted form? _Doubt it would happen in a normal submission flow but is it something we should reject?_

ukanga: No, not really. It is not going to affect exports adversely. That element does bring to question what other validations we need to do.

ukanga (approved):

PR #1953, opened 2020-11-23
Add enketo-preview url routed to PreviewXFormListViewSet

### LINKED ISSUE #1952: Private forms not on FormList for authenticated users but available for anonymous users
### Problem description
Authenticated users with can view permissions are not able to open enketo preview links for private forms because the forms are not in their FormList when they go to `<onadata_url>/enketo/<form_pk>/formList?formID=<id_string>`

This is most likely caused by: https://github.com/onaio/onadata/blob/master/onadata/libs/filters.py#L77 which is used here: https://github.com/onaio/onadata/blob/24b743fc1acd13d7cab8e776a0afdbcd508ae679/onadata/apps/api/viewsets/xform_list_viewset.py#L45
### Expected behavior
Users with `can view` permissions on a form/project, should have the form on their formList so that they are able to see the form preview

FrankApiyo:
Add a new URL that is routed to `PreviewXFormListViewSet`
This has been tested on the staging environment and works as expected
Closes #1952


ukanga (commented):
Just concerned we do not have test to perhaps check this will work as expected Perhaps in `onadata/apps/api/tests/viewsets/test_xform_list_viewset.py` as was the case when we introduced the [/enketo path](https://github.com/onaio/onadata/commit/e82747827f522bb60f191d09617e9f9fa18de1d6#diff-b8a6e4ab17d4bfdc0b7715074bfbe0b8919d38166d7e6e45cf91dad979d1a27d).

ukanga (approved):

ukanga (commented):
Could we also add some documentation in the docs for this new endpoint as well perhaps on the enketo endpoint. Check with @DavisRayM on the full docs for both endpoints.

docs/formlist.rst
```diff
@@ -0,0 +1,147 @@
...
+
+GET /<username>/<form_pk>/formList
+
+**Pass username and form pk**
+
+Example
+^^^^^^^
+::
+
+    curl -X GET https://stage-api.ona.io/<user_name>/<form_pk>/formList
```
ivermac, line 70: Can we use `api.ona.io` on our example urls?

docs/formlist.rst
```diff
@@ -0,0 +1,147 @@
...
+
+
+**Use enketo/<xform_pk> endpoint**
+
+Example
+^^^^^^^
+
+::
+
+    curl -X GET https://stage-api.ona.io/enketo/<form_pk>/formList
```
ivermac, line 100: Same comment as above

docs/formlist.rst
```diff
@@ -0,0 +1,147 @@
...
+GET /enketo-preview/<xform_pk>/formList
+
+**Use enketo-preview/<xform_pk> endpoint**
+
+Example
+^^^^^^^
+
+::
+
+    curl -X GET https://stage-api.ona.io/enketo-preview/<form_pk>/formList
```
ivermac, line 130: Same comment as above?

ivermac (changes requested):
Is it correct to assume that `PreviewXFormListViewSet` has already been implemented?

FrankApiyo:
> Is it correct to assume that `PreviewXFormListViewSet` has already been implemented?

@ivermac yes! https://github.com/onaio/onadata/blob/master/onadata/apps/api/viewsets/xform_list_viewset.py#L167

ivermac (approved):

ukanga (approved):
Can you ensure that all checks are passing?

PR #1966, opened 2020-12-09
Data viewset retrieval optimisations

DavisRayM:

- Introduced new `SUBMISSION_RETRIEVAL_THRESHOLD` setting
- Disable Queryset ordering after an XForms' submission count surpasses the `SUBMISSION_RETRIEVAL_THRESHOLD` setting
- Disable ETag functionality after an XForms' submission count surpasses the `SUBMISSION_RETRIEVAL_THRESHOLD` setting
- Add `CountOverridablePaginator` and `CountOverridablePageNumberPagination`
- Switch the default pagination class for the Data viewset to `CountOverridablePageNumberPagination`

### TODO

- Set a limit on the number of records that can be returned at one time when pagination is not used; _Separate PR... Will most likely break 3rd Party applications that rely on this functionality_

- N/A

- After a certain threshold data retrieval (even via pagination) is not ordered and the ETag functionality is disabled

### DIFF
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -62,6 +62,9 @@ from onadata.libs.utils.model_tools import queryset_iterator
 from onadata.libs.utils.viewer_tools import get_form_url, get_enketo_urls
 
 SAFE_METHODS = ['GET', 'HEAD', 'OPTIONS']
+SUBMISSION_RETRIEVAL_THRESHOLD = getattr(settings,
+                                         "SUBMISSION_RETRIEVAL_THRESHOLD",
+                                         10000)
 BaseViewset = get_baseviewset_class()
 
 
@@ -469,8 +472,12 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
     def set_object_list(
             self, query, fields, sort, start, limit, is_public_request):
         try:
+            enable_etag = True
             if not is_public_request:
                 xform = self.get_object()
+                self.data_count = xform.num_of_submissions
+                enable_etag = self.data_count <\
+                    SUBMISSION_RETRIEVAL_THRESHOLD
 
             where, where_params = get_where_clause(query)
             if where:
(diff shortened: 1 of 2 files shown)


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -409,12 +413,23 @@ def list(self, request, *args, **kwargs):
...
-                xforms = MergedXForm.objects.filter(
-                    pk=xform_id, xforms__deleted_at__isnull=True)\
-                    .values_list('xforms', flat=True)
-                pks = [pk for pk in xforms if pk] or [xform_id]
+                merged_form = MergedXForm.objects.get(pk=xform_id)
+                pks = merged_form.xforms.filter(
+                    deleted_at__isnull=True).values_list('id', flat=True)
+                num_of_submissions = sum(
+                    merged_form.xforms.values_list(
+                        'num_of_submissions', flat=True))
```
ivermac, line 421: Should the above also filter by deleted_at is not null i.e.
```python
merged_form.xforms.filter(deleted_at__isnull=True).values_list('num_of_submissions', flat=True)
```
If the above is true, can the query in line 417 and 420 be merged into 1 to retrieve both the `id` and `num_of_submissions `?

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -409,12 +413,23 @@ def list(self, request, *args, **kwargs):
...
+            else:
+                num_of_submissions = XForm.objects.get(
+                    id=xform_id).num_of_submissions
             self.object_list = Instance.objects.filter(
-                xform_id__in=pks, deleted_at=None).only('json').order_by('id')
+                xform_id__in=pks, deleted_at=None).only('json')
+
+            # Disable ordering for XForms with Submissions that are more
+            # than the SUBMISSION_RETRIEVAL_THRESHOLD
```
ivermac, line 429: Or
```python
# Enable ordering for XForms with Submissions that are less
# than the SUBMISSION_RETRIEVAL_THRESHOLD
```
😄

ivermac (approved):

Issue #1948, opened 2020-11-13
When a user verifies their email the metadata is not always updated to reflect the verification status


samwata:
### Environmental Information
Production

### Problem description
When a user creates a new account, the verification email is sent to their email and after they have verified their email, the `is_email_verified` metadata data doesn't get updated.

### Steps to reproduce the behavior
1. Create a new account
2. Verify the email from the link received
3. Check your new account if the intercom chatbot is available

PR #1970, opened 2020-12-14
Clear cache and refresh user profile on email verification

DavisRayM:

- Refresh user profile from database before updating metadata
- Clear user-profiles cache on email verification

- N/A

- N/A

Possible fix for #1948

### DIFF
--- a/onadata/apps/api/viewsets/user_profile_viewset.py
+++ b/onadata/apps/api/viewsets/user_profile_viewset.py
@@ -100,7 +100,8 @@ def serializer_from_settings():
 
 
 def set_is_email_verified(profile, is_email_verified):
-    profile.metadata.update({"is_email_verified": is_email_verified})
+    profile.refresh_from_db()
+    profile.metadata.update({'is_email_verified': is_email_verified})
     profile.save()
 
 
@@ -214,7 +215,6 @@ class UserProfileViewSet(
             return Response(cached_user)
         response = super(UserProfileViewSet, self)\
             .retrieve(request, *args, **kwargs)
-        cache.set(f'{USER_PROFILE_PREFIX}{username}', response.data)
         return response
 
     def create(self, request, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


FrankApiyo (approved):

ivermac (dismissed):

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -100,7 +100,8 @@ def serializer_from_settings():
 
 
 def set_is_email_verified(profile, is_email_verified):
-    profile.metadata.update({"is_email_verified": is_email_verified})
+    profile.refresh_from_db()
```
ukanga, line 103: Can we make this call before we call `set_is_email_verified()`? I see a reason why the `refresh_from_db` may need to happen on the GET request to verify, but perhaps not on the POST request. I would hope all requests on a POST use the master DB. 

Is this issue due to a read replica update delay?

DavisRayM: Yes, I can change this.

DavisRayM: Doesn't seem to be a read replica update delay; the change never shows up. Noticed accounts that had been verified a while back(2-3 days) but the metadata wasn't updated

ukanga: What then guarantees this issue is resolve considering the statement has always been there? The only new statements is the cache deletion and the refresh_from_db.

DavisRayM: Changed this in the latest commits. The main issue seems to have been the `RegistrationProfile` retrieval in the user profile viewset.

ivermac (dismissed):

ivermac (approved):

ukanga (approved):

PR #1981, opened 2021-01-08
Add support for querying a column with multiple conditions

### LINKED ISSUE #1979: Filtering data using different conditions on same column only filters data using the second condition
I want to filter data using two conditions from the same column.
For instance am filtering data when enumerator name is `samwata` or `amwata` from `enumerator_name` column using the following query.

`https://api.ona.io/api/v1/data/531845?query={"$or":[{"enumerator_name":"amwata","enumerator_name":"samwata"}]} 
`

This only filters data with enumerator name `samwata` only but it does not include data with enumarator name as `amwata`
I also tried to separate the conditions to different dictionaries but the result is the same

` https://api.ona.io/api/v1/data/531845?query={"$or":[{"enumerator_name":"amwata"}, {"enumerator_name":"samwata"}]} `

DavisRayM:

- Added support for multiple query conditions on the same column

- Added Test

- N/A

Closes #1979

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_data_viewset.py
@@ -11,7 +11,6 @@ import geojson
 import requests
 from django.core.cache import cache
 from django.db.utils import OperationalError
-from django.db.models import Q
 from django.test import RequestFactory
 from django.test.utils import override_settings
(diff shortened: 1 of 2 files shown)


ukanga (commented):
Looks good, fix checks.

ivermac (approved):

PR #1988, opened 2021-01-12
Retrieve user profile using case insensitive username filter

### LINKED ISSUE #1982: <username>/formList endpoint is not case insensitive
### Environmental Information

- Onadata version: latest

### Problem description

Currently, the `<username>/formList` endpoint is not case insensitive when it searches for the `username` parameter. Which leads to scenarios where `davisraym/formList` would return 200s while `davisRayM/formList` would return 404s.

### Expected behavior

The endpoint should be case-insensitive like all other Aggregate endpoints i.e `<username>/submission`

1. Create a user with username `bob`
2. (Optional) Upload a XForm
3. Try accessing `/Bob/formList`

### Additional Information

DavisRayM:

- Retrieve user profile using case insensitive username filter

- Added test

All requests made to `<username>/formList` will now search for the `user` via a case-insensitive search using the username.

Closes #1982


ivermac (approved):
Looks good to me

ukanga (approved):

PR #2002, opened 2021-01-25
Use master database when updating XForm Submission Count

### LINKED ISSUE #1987: Incorrect submission count on XForms
### Environmental Information
production
- Onadata version: v2.4.9

### Problem description
The `num_of_submissions` in the XForm model has incorrect values for some forms

It seems `num_of_submissions` is not always updated correctly here https://github.com/onaio/onadata/blob/master/onadata/apps/logger/models/instance.py#L178

### Expected behavior
The `num_of_submissions` field should have the same value as the number of submissions that have not been deleted

DavisRayM:

- Avoids cases where the signal is fired before a read replica catches
  up to the main database

Closes #1987

### DIFF
--- a/onadata/apps/logger/models/instance.py
+++ b/onadata/apps/logger/models/instance.py
@@ -177,40 +177,44 @@ def _update_submission_count_for_today(
 @transaction.atomic()
 def update_xform_submission_count(instance_id, created):
     if created:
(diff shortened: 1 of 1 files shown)


ivermac (approved):
looks good to me

PR #2003, opened 2021-01-25
Capture attachment file names whose name exceeds 100 chars

### LINKED ISSUE #1939: Return user friendly warning when an attachment name exceeds 100 characters
### Environmental Information
Production
Stage
Preview

- Onadata version: v2.4.4

### Problem description
When a user uploads a media file using enketo that has a name exceeding 100 characters, they get the error as shown in the image
picturemessage_z21wl3n3 kre

### Expected behavior
We expect to have a user friendly error that the user can take to action.

Upload a file that collects media attachments
Try upload an image names with more than 100 characters

WinnyTroy:
- Include check to confirm attachment file name length before saving submission

- Added Tests

Closes #1939

### DIFF
--- a/onadata/libs/utils/logger_tools.py
+++ b/onadata/libs/utils/logger_tools.py
@@ -41,7 +41,7 @@ from onadata.apps.logger.xform_instance_parser import (
     DuplicateInstance, InstanceEmptyError, InstanceInvalidUserError,
     InstanceMultipleNodeError, InstanceEncryptionError, NonUniqueFormIdError,
     InstanceFormatError, clean_and_parse_xml, get_deprecated_uuid_from_xml,
-    get_submission_date_from_xml, get_uuid_from_xml)
+    get_submission_date_from_xml, get_uuid_from_xml, AttachmentNameError)
 from onadata.apps.messaging.constants import XFORM, \
     SUBMISSION_EDITED, SUBMISSION_CREATED
 from onadata.apps.messaging.serializers import send_message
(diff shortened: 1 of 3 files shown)


DavisRayM (approved):

onadata/libs/utils/logger_tools.py
```diff
@@ -336,6 +336,10 @@ def save_attachments(xform, instance, media_files, remove_deleted_media=False):
             xform.instances_with_osm = True
             xform.save()
         filename = os.path.basename(f.name)
+        # Validate Attachment file name length
+        if len(filename) > 99:
```
ivermac, line 340: Based on the error message, should the number in the condition be 100 instead of 99? At the moment, if the size of the filename is 100, the error message will be triggerd, no?

ivermac (commented):
It generally looks good. I have a question though

onadata/libs/utils/logger_tools.py
WinnyTroy: Made the change

ivermac (approved):
lgtm

DavisRayM (approved):

onadata/libs/utils/logger_tools.py
```diff
@@ -336,6 +336,10 @@ def save_attachments(xform, instance, media_files, remove_deleted_media=False):
             xform.instances_with_osm = True
             xform.save()
         filename = os.path.basename(f.name)
+        # Validate Attachment file name length
+        if len(filename) > 100:
+            raise AttachmentNameError(
+                _('Attachment file name exceeds 100 characters'))
```
DavisRayM, line 342: Let's ensure this exception is caught and an appropriate status code is returned with the reason.

DavisRayM (changes requested):

DavisRayM (approved):

PR #2007, opened 2021-01-29
2006 trigger error on url in xform title

### LINKED ISSUE #2006: Trigger error when xform title matches a url
### Environmental Information
All
### Onadata version:
master 

### Problem description
Prevent HTML injection

### Expected behavior
Trigger an error when xform title matches url

Use PUT request to update form

### Additional Information

ivermac:
Trigger error when xform title matches a url

Added a test to check that an error is triggered when xform title matches a url

Closes #2006


onadata/apps/logger/models/xform.py
```diff
@@ -818,6 +818,13 @@ def _set_title(self):
                 _("Title shouldn't have any invalid xml "
                   "characters ('>' '&' '<')"))
 
+        pattern = str("http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|"
```
DavisRayM, line 821: Should we update the pattern to also match URLs that don't include the HTTP/HTTPS protocols?

Screenshot 2021-02-01 at 12 10 31 PM

DavisRayM (approved):
LGTM, just a few suggestions

onadata/apps/logger/models/xform.py
ivermac: Yes we should. Let me update

DavisRayM (approved):
LGTM

PR #2009, opened 2021-02-04
Support query by date_modified field

### LINKED ISSUE #2001: Ability to search by last modified date for both new and edited onadata form instances
To my understanding, the last_modified field is not exposed for querying in the v1 data/ api - this makes it awkward to paginate ona data form instances by creation or update time: 

https://github.com/onaio/onadata/blob/20a4dece8fc7bee435e3512266649deec5200c66/onadata/apps/logger/models/instance.py#L486

It's possible to paginate by submission time or last edited time separately, for example:
https://api.ona.io/api/v1/data/558517?query={%22_last_edited%22:{%22$gte%22:%222020-01-19T11:40:06%22}}

But the last_edited field only exists for form instances that have actually been edited, so no unedited instances appear, and merging two sets of differently-paginated results is pretty complex to get efficient and correct.

Also, from my understanding of the code the last_modified field is tied to the data structure and auto-populated by django libraries, which may make it more robust vs the last_edited which seems to be linked to another history data structure.  It may be more natural to use for ETL.

If there's functionality I'm missing in this direction it would definitely be good to know, a request for discussion.

Much easier to paginate form instance data by modification time, easier to ETL.

DavisRayM:

- Add `_date_modified` into the generated Instance JSON
- Update tests
- Add `created` attribute to the data list XML response  Instance representations

- Updated tests

- Introduces new migrations that regenerate the JSON representation of all un-deleted Instances. Possible chance of downtime while the migration takes place

Closes #2001


ivermac (dismissed):
Looks good. Quick question, why do we need the migration file?

onadata/apps/logger/models/instance.py
```diff
@@ -389,6 +386,7 @@ def get_full_dict(self, load_existing=True):
                 self.date_created = submission_time()
 
             doc[SUBMISSION_TIME] = self.date_created.strftime(MONGO_STRFTIME)
+            doc[DATE_MODIFIED] = self.date_modified.strftime(MONGO_STRFTIME)
```
ukanga, line 389: We do have this check https://github.com/onaio/onadata/pull/2009/files#diff-98ae21dd41854ad6194bdbca52b97c4d2b0231f6d06b3ee5ee73c4c88a135e27R385-R386 for `date_created` ensuring it is not null, should we have the same for `date_modified`?

DavisRayM:
> Looks good. Quick question, why do we need the migration file?

It's there so that the JSON is regenerated to enable users to start using the `_date_modified` filter. _The JSON doesn't seem like it's updated till a Submission is edited_

onadata/apps/logger/migrations/0062_auto_20210202_0248.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('logger', '0061_auto_20200713_0814'),
+    ]
+
+    operations = [
+        migrations.RunPython(regenerate_instance_json)
+    ]
```
ukanga, line 25: We will need a strategy to run this in our production environment, other deployments might be okay.

onadata/apps/logger/models/instance.py
```diff
@@ -388,6 +385,12 @@ def get_full_dict(self, load_existing=True):
             if not self.date_created:
                 self.date_created = submission_time()
 
+            if not self.date_modified:
+                doc[DATE_MODIFIED] = ""
```
ukanga, line 389: Could we simply not include the field or simply use the `date_created` value, since that would ideally be the same value on the initial creation of the record.

onadata/apps/logger/models/instance.py
```diff
@@ -388,6 +385,12 @@ def get_full_dict(self, load_existing=True):
             if not self.date_created:
                 self.date_created = submission_time()
 
+            if not self.date_modified:
+                doc[DATE_MODIFIED] = self.date_created
```
ukanga, line 389: Should this be getting, the date format treatment as below or is it handled elsewhere?
```python
doc[DATE_MODIFIED] = self.date_created.strftime(MONGO_STRFTIME)
```

onadata/apps/logger/models/instance.py
```diff
@@ -388,6 +385,13 @@ def get_full_dict(self, load_existing=True):
             if not self.date_created:
                 self.date_created = submission_time()
 
+            if not self.date_modified:
+                doc[DATE_MODIFIED] = self.date_created.strftime(
+                    MONGO_STRFTIME)
+            else:
+                doc[DATE_MODIFIED] = self.date_modified.strftime(
+                    MONGO_STRFTIME)
+
```
ukanga, line 394: Simplify

```python
            if not self.date_modified:
                self.date_modified= self.date_created
           
           doc[DATE_MODIFIED] = self.date_modified.strftime(
                    MONGO_STRFTIME)
```

ukanga (approved):

PR #2010, opened 2021-02-04
Paginate data list responses after a configurable threshold 

### LINKED ISSUE #2008: Data ViewSet: Automatically paginate requests after a certain submission threshold

Currently, the data viewset returns all the instances present in a form if the `page_size` query param is not specified. It would be nice if the response was paginated if a form has a lot of submissions (10k+ perhaps ?) in order to reduce the amount of processing it may take to retrieve all the records.

DavisRayM:

- Paginate responses when requested data is more than the retrieval threshold

- Added tests

Closes #2008

### DIFF
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -617,20 +617,32 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
         self.set_object_list(
             query, fields, sort, start, limit, is_public_request)
 
+        xform = self.get_object()
+        retrieval_threshold = getattr(
+            settings, "SUBMISSION_RETRIEVAL_THRESHOLD", 10000)
         pagination_keys = [self.paginator.page_query_param,
                            self.paginator.page_size_query_param]
         query_param_keys = self.request.query_params
         should_paginate = any([k in query_param_keys for k in pagination_keys])
+
+        if not should_paginate:
+            # Paginate requests that try to retrieve data that surpasses
+            # the submission retrieval threshold
+            num_of_submissions = xform.num_of_submissions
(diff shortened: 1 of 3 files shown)


ivermac (approved):
Looks good to me

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -617,17 +617,29 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
...
             query, fields, sort, start, limit, is_public_request)
 
+        retrieval_threshold = getattr(
+            settings, "SUBMISSION_RETRIEVAL_THRESHOLD", 10000)
         pagination_keys = [self.paginator.page_query_param,
                            self.paginator.page_size_query_param]
         query_param_keys = self.request.query_params
         should_paginate = any([k in query_param_keys for k in pagination_keys])
+
+        if not should_paginate and not is_public_request:
```
ukanga, line 627: Does this mean we will not paginate public forms that exceed the threshold?

DavisRayM: At the moment yes; didn't think of this while implementing the feature... Any thoughts/reservations with always paginating public requests? _Seems a bit painful to calculate the entire count(getting all public XForms and summing the `num_of_submissions` of public data that's currently available..._

ukanga: This is for a single form request at a time, not all public forms at the same. Also, we could go with `num_of_submissions` that is not calculated on the fly, it is a safe estimate in most situations.

ukanga: I withdraw my statement, I see that is_public_request represents access to the `/api/v1/data/public` and not to one single form data.

ukanga (approved):

PR #2011, opened 2021-02-05
Check if XForm is a MergedXForm and merge field choices if it is(a MergedXForm)

### LINKED ISSUE #2013: Charts endpoint does not return correct data for MergedXForms
### Environmental Information

- Onadata version: `2.5.0`

### Problem description
For a merged XForm consisting of two XForms with the following markdown:
```
MD = """
| survey  |
|         | type              | name  | label   |
|         | select one fruits | fruits | Fruit   |

| choices |
|         | list name         | name   | label  |
|         | fruits            | orange | Orange |
|         | fruits            | mango  | Mango  |
"""


MD2 = """
| survey  |
|         | type              | name  | label   |
|         | select one fruits | fruits | Fruit   |

| choices |
|         | list name         | name   | label  |
|         | fruits            | apple  | Apple  |
|         | fruits            | cherries  | Cherries  |
"""

```
Incorrect data is returned from `/api/v1/charts/pk?field_xpath=fruits` endpoint

### Expected data

`{'data': [{'fruits': ['Apple', 'Cherries'], 'count': 1},
                         {'fruits': ['Orange', 'Mango'], 'count': 1}]}`

### Returned data

`{'data': [{'fruits': ['apple cherries'], 'count': 1},
                         {'fruits': ['Orange', 'Mango'], 'count': 1}]}`

Create a merged Dataset with two forms that look similar to the above markup and query `/api/v1/charts/pk?field_xpath=fruits` with your merged dataset's `pk`

FrankApiyo:
- Added a test case

Closes #2013

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_charts_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_charts_viewset.py
@@ -3,19 +3,46 @@ import os
 import mock
 
 from django.utils import timezone
+from django.core.cache import cache
 from rest_framework.test import APIClient
 from rest_framework.test import APIRequestFactory
 from rest_framework.test import force_authenticate
 from onadata.apps.api.viewsets.charts_viewset import ChartsViewSet
+from onadata.apps.api.viewsets.merged_xform_viewset import MergedXFormViewSet
 from onadata.apps.main.tests.test_base import TestBase
 from onadata.apps.logger.models.instance import Instance
 from django.db.utils import DataError
 from onadata.libs.utils.timing import calculate_duration
+from onadata.libs.utils.user_auth import get_user_default_project
 from onadata.libs.renderers.renderers import DecimalJSONRenderer
 
 
 def raise_data_error(a):
     raise DataError
+
+
+MD = """
+| survey  |
+|         | type              | name  | label   |
+|         | select one fruits | fruits | Fruit   |
+
+| choices |
+|         | list name         | name   | label  |
+|         | fruits            | orange | Orange |
+|         | fruits            | mango  | Mango  |
+"""
+
+
+MD2 = """
+| survey  |
+|         | type              | name  | label   |
+|         | select one fruits | fruits | Fruit   |
+
+| choices |
+|         | list name         | name   | label  |
+|         | fruits            | apple  | Apple  |
+|         | fruits            | cherries  | Cherries  |
+"""
 
 
 class TestChartsViewSet(TestBase):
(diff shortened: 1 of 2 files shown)


onadata/libs/utils/chart_tools.py
```diff
@@ -435,6 +436,16 @@ def get_field_label(field, language_index=0):
...
 
 
+def _get_form_field_from_name_or_xpath(xform, name=None, xpath=None):
+    """
+    Given a name or xpath, get xform field
+    """
+    if name:
+        return get_field_from_field_name(name, xform)
+    elif xpath:
+        return get_field_from_field_xpath(xpath, xform)
```
ivermac, line 446:
```suggestion
    return get_field_from_field_xpath(name or xpath, xform)
```

onadata/libs/utils/chart_tools.py
```diff
@@ -458,10 +466,26 @@ def get_chart_data_for_field(field_name,
...
+                                               name=field_name,
+                                               xpath=field_xpath)
 
     choices = get_field_choices(field, xform)
+    # check if xform is a MergedXForm and if it is, merge the field's choices
+    try:
+        merged_xform = MergedXForm.objects.get(pk=xform.pk)
+    except MergedXForm.DoesNotExist:
+        # this is not  a MergedXForm
+        pass
```
ivermac, line 479: XForm has `is_merged_dataset` property. We could re-write the above as
```python
if xform.is_merged_dataset:
    merged_xform = MergedXForm.objects.get(pk=xform.pk)
```

onadata/apps/api/tests/viewsets/test_charts_viewset.py
```diff
@@ -47,6 +74,59 @@ def setUp(self):
                 os.path.dirname(__file__), '..', 'fixtures', 'forms',
                 'tutorial', 'instances', '3.xml'))
 
+    def test_correct_data_for_charts(self):
```
ivermac, line 77: Should we change the test name to something like `test_correct_merged_dataset_data_for_charts`?

ivermac (changes requested):
I've suggested a few changes but it generally looks good

onadata/libs/utils/chart_tools.py
FrankApiyo: I'm not sure if this would work, because we need to call a different method when we have `name` from when we have `xpath`

ivermac: Aah...cool. My bad,  I thought it was the same function

ivermac (approved):
looks good to me. I'd like @DavisRayM to check this out as well

DavisRayM (approved):

PR #2014, opened 2021-02-11
Use language parameter to create exports

### LINKED ISSUE #2017: Use query parameter passed to exports endpoint to create exports with language other than the default

Use language query parameter passed to export builder [here](https://github.com/onaio/onadata/blob/master/onadata/libs/utils/export_tools.py#L191) to create exports in a language different from the default

FrankApiyo:
Closes #2017


onadata/apps/logger/models/xform.py
```diff
@@ -534,13 +534,16 @@ def get_language(self, languages, language_index=0):
...
+    def get_label(self, abbreviated_xpath, elem=None, language=None):
         elem = self.get_element(abbreviated_xpath) if elem is None else elem
 
         if elem:
             label = elem.label
 
-            if isinstance(label, dict):
+            if language and isinstance(label, dict):
+                label = label[language]
+
```
DavisRayM, line 545: Should we handle cases where a user passes a language that doesn't exist on the XForm to the `get_label` function ? _Seems the `self.get_language` function has a list of languages... or we could just use `label.get(language, <default>)`.._

Or is that something that's handled else where ?

FrankApiyo: Such a situation has not been taken in to account; going to tweak this a bit

DavisRayM (approved):
LGTM

ivermac (dismissed):
looks good to me

ivermac (approved):

DavisRayM (approved):

ivermac (approved):
Looks good to me

PR #2019, opened 2021-02-16
Add support for XForm versioning

### LINKED ISSUE #1999: Add ability to retrieve prior XForm versions

DavisRayM:

- Add `XFormVersion` model
- Add `versions` endpoint to XForm viewset

- Added tests

- N/A

### Notes for reviewers

The current implementation does not support the creation of XForm Versions that were replaced before this implementation was in place.

Closes #1999


onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -749,6 +761,40 @@ def destroy(self, request, *args, **kwargs):
 
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    @action(methods=['GET', 'POST'], detail=True)
+    def version(self, request, *args, **kwargs):
+        xform = self.get_object()
+        version_id = kwargs.get('version_id')
+        job_uuid = kwargs.get('job_uuid')
+        version = None
```
ivermac, line 769: Are we passing the `job_uuid` as a query? something like the following
`/api/v1/forms/<form-id>/version/<version>?job_uuid=<job_uuid>`

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -749,6 +761,40 @@ def destroy(self, request, *args, **kwargs):
...
+                    'xml': version.xml
+                }
+                status_code = status.HTTP_200_OK
+            else:
+                task_id = tasks.find_and_create_xls_form_version_async.delay(
+                    xform, version_id).task_id
+                data = {
+                    'job_uuid': task_id
+                }
+                status_code = status.HTTP_202_ACCEPTED
```
ivermac, line 790: What should happen when the version_id doesn't actually exist or is invalid?

ivermac (commented):
A couple of comments but it generally looks good

onadata/apps/api/viewsets/xform_viewset.py
DavisRayM: Yes, I've updated this piece of code to retrieve the value from the `query_params`

onadata/apps/logger/models/xform_version.py
```diff
@@ -0,0 +1,30 @@
...
+    the previous XForm versions XML or JSON.
+    """
+    xform = models.ForeignKey(
+            'logger.XForm', on_delete=models.CASCADE, related_name='versions')
+    # TODO: Test to see if the file is re-uploaded or renamed during the
+    # creation process. The main idea is just to link to the existing file
+    # and not recreate it
+    xls = models.FileField()
+    version = models.CharField(max_length=100)
+    date_created = models.DateTimeField(auto_now_add=True)
```
ukanga, line 23: Could we ask who created/uploaded the version/XLSForm? Have a `date_modified` field?

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -749,6 +763,74 @@ def destroy(self, request, *args, **kwargs):
...
+        # Retrieve a list of all available versions from the XForms'
+        # instances if the number of submissions are not more than the
+        # retrieval threshold
+        if not xform.num_of_submissions > getattr(
+            settings, "SUBMISSION_RETRIEVAL_THRESHOLD", 10000
+        ) and not versions:
+            versions = list(
+                xform.instances.filter(
+                    deleted_at__isnull=True
+                ).values('version').annotate(total=Count('version'))
```
ukanga, line 784: It does not seem like we use the version counts in any of the responses below, why do we need this here?

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -749,6 +763,74 @@ def destroy(self, request, *args, **kwargs):
...
+        if versions and version_id not in versions:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+        elif job_uuid:
+            resp = tasks.get_async_status(job_uuid)
+            return Response(data=resp, status=status.HTTP_202_ACCEPTED)
+        else:
+            if version_id == xform.version:
+                data = {
+                    'xml': xform.xml
+                }
```
ukanga, line 813: I see the need to get the XForm XML when you hit `/api/v1/forms/[xform pk]/versions/[version pk].xml. Why would we want the XML as part of a JSON response?

ukanga (commented):
I am not sure that all aspects of the new endpoint have been tested, or it is not clear to me what the API responses should be. Could you add documentation to aid with that understanding? Something that would be part of the API docs.

onadata/apps/api/tasks.py
```diff
@@ -116,3 +120,47 @@ def send_verification_email(email, message_txt, subject):
 @app.task()
 def send_account_lockout_email(email, message_txt, subject):
     send_generic_email(email, message_txt, subject)
+
+
+def find_and_create_xls_form_version(xform: XForm, version: str) -> None:
+    """
+    Locates a specific XForm version in the Storage backend and
+    creates an XForm version for the version
+    """
```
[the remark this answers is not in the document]
DavisRayM: This was added in order to support XForms uploaded before the change. Yes, I can remove the support for the previously uploaded forms. Will update the PR in my upcoming commits

onadata/apps/api/tests/viewsets/test_xform_viewset.py
```diff
@@ -3350,6 +3352,13 @@ def test_different_form_versions(self):
                 request = self.factory.patch('/', data=post_data, **self.extra)
                 response = view(request, pk=self.xform.pk)
                 self.assertEqual(response.status_code, 200)
+                # ensure that an XForm version object is created
+                versions = XFormVersion.objects.filter(xform=self.xform)
+                self.assertEqual(versions.count(), 1)
```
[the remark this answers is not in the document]
DavisRayM: Yes; I'll make changes to create an `XFormVersion` object on form uploads. Ideally we'd only need to check the `XFormVersion` for the versions of the form; this may not be the case for forms uploaded previously before the changes... for those forms we'd need to query the XForm instances.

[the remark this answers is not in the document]
ukanga: We can handle that via a migration, by creating for all existing forms. What I had concerns about is reprocessing XLSForm that had been uploaded, I do not think we should reprocess XLSForms but we can create versions of existing forms.

onadata/apps/logger/migrations/0064_auto_20210304_0314.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def create_initial_xform_version(apps, schema_editor):
+    """
+    Creates an XFormVersion object for an XForm that has no
+    Version
+    """
+    for xform in XForm.objects.filter(
+            downloadable=True,
+            deleted_at__isnull=True):
```
ukanga, line 16: Should we use a queryset iterator here? For a large number of forms, these may have adverse effects on a production system.

onadata/apps/api/viewsets/xform_viewset.py
```diff
@@ -749,6 +752,24 @@ def destroy(self, request, *args, **kwargs):
...
+        xform = self.get_object()
+        version_id = kwargs.get('version_id')
+        requested_format = kwargs.get('format') or 'json'
+
+        stored_versions = XFormVersion.objects.filter(
+            xform=xform).values_list('version')
+        stored_versions = [version[0] for version in stored_versions]
+
+        if not version_id:
+            return Response(data=stored_versions, status=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
DavisRayM: Changed this in the latest commits. I've introduced a serializer for this as suggested; I've also added the `url` field to reference the resource...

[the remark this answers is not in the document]
DavisRayM: The `xml` url might be a nice to have.... looking into how I can implement that

onadata/apps/logger/migrations/0064_auto_20210304_0314.py
```diff
@@ -0,0 +1,32 @@
...
+    Creates an XFormVersion object for an XForm that has no
+    Version
+    """
+    queryset = XForm.objects.filter(
+        downloadable=True,
+        deleted_at__isnull=True
+    )
+    for xform in queryset.iterator():
+        if xform.version and not XFormVersion.objects.filter(
+                xform=xform, version=xform.version).first():
```
ukanga, line 20: This seems kind of expensive, for 40k forms this will be 1 select query + 40k select queries + 40k insert queries, could we have a unique filter so that in the event the record exists it will fail and hence have 1 select query + 40k insert queries?

docs/forms.rst
```diff
@@ -1390,6 +1390,150 @@ Response
...
+        curl -X GET https://api.ona.io/api/v1/forms/1/versions
+
+Response
+^^^^^^^^
+
+::
+
+    [
+        "202103011453",
+        "202103011520"
```
ukanga, line 1412: The data format changed, could this be updated to reflect the new format?

onadata/apps/api/tools.py
```diff
@@ -76,6 +77,22 @@ def _get_id_for_type(record, mongo_field):
...
+    if xform.version and not XFormVersion.objects.filter(
+            xform=xform, version=xform.version).first():
+        return XFormVersion.objects.create(
+            xform=xform,
+            xls=xform.xls,
+            json=xform.json,
+            version=xform.version,
+            created_by=user,
+            xml=xform.xml
+        )
```
ukanga, line 93: I think the norm is to create the record and have it fail on an `IntegrityError` exception if the record already exists, instead of checking before inserting. We are likely to have duplicate records when two separate process run at about the same time.

onadata/apps/logger/migrations/0064_auto_20210304_0314.py
ukanga: Also, this check is also being done by the `create_xform_version()` function, so we currently are going to have 1 select query + 40k select queries + 40k select queries (done by `create_xform_version()`)  + 40k insert queries

onadata/apps/api/tools.py
```diff
@@ -340,10 +357,16 @@ def add_team_to_project(team, project):
...
     """
-    return do_publish_xlsform(request.user, request.data, request.FILES, owner,
-                              id_string, project)
+    survey = do_publish_xlsform(
+        request.user, request.data, request.FILES, owner, id_string, project)
+
+    # Create an XFormVersion object if the returned object is an
+    # XForm object
+    if isinstance(survey, XForm):
+        create_xform_version(survey, request.user)
```
ukanga, line 368: I wonder whether this should have been called or addressed at https://github.com/onaio/onadata/blob/2d039f2335bc3cf572f5923616f744e7360f4c6f/onadata/libs/utils/logger_tools.py#L655 and https://github.com/onaio/onadata/blob/2d039f2335bc3cf572f5923616f744e7360f4c6f/onadata/libs/utils/logger_tools.py#L683 such that all cases where a form is published will be captured. If we do publish an XForm with ODK Briefcase, will a version be created in this case?

onadata/apps/logger/migrations/0064_auto_20210304_0314.py
DavisRayM: I've changed this in the latest commit. I've removed the `XFormVersion.objects....` line. In favor of the mechanism in `create_xform_version()`; Just passes on an integrity error....

ukanga (approved):

PR #2022, opened 2021-02-25
Fix Charts endpoint JSON response not rendering

DavisRayM:

- Set `xform` data key to the XForm ID when no `accepted_format` is given

- Updated tests

- When no accepted format is given the return `xform` value shall be the ID

### DIFF
--- a/onadata/libs/utils/chart_tools.py
+++ b/onadata/libs/utils/chart_tools.py
@@ -473,7 +473,7 @@ def get_chart_data_for_field(field_name,
     except DataError as e:
         raise ParseError(text(e))
     else:
-        if accepted_format == 'json':
+        if accepted_format == 'json' or not accepted_format:
             xform = xform.pk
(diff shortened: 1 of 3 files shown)


onadata/libs/utils/chart_tools.py
```diff
@@ -473,7 +473,7 @@ def get_chart_data_for_field(field_name,
     except DataError as e:
         raise ParseError(text(e))
     else:
-        if accepted_format == 'json':
+        if accepted_format == 'json' or not accepted_format:
```
ivermac, line 476: Why do we need `or not accepted_format`?

ivermac (approved):
Looks good to me. I've added a question though

onadata/libs/utils/chart_tools.py
DavisRayM: Sometimes the `format` kwarg is empty; the format kwarg is the value we use to determine what type of response to return(but by default we return JSON)

PR #2028, opened 2021-02-26
Include ability to delete KPI tables and data from Ona db

WinnyTroy:
Use django ORM to remove KPI related tables and data from the Ona db

Closes onaio/infrastructure#3682

### DIFF
--- a/onadata/apps/main/management/commands/remove_kpi_tables.py
+++ b/onadata/apps/main/management/commands/remove_kpi_tables.py
@@ -1,0 +1,81 @@
+from django.db import connection
+from django.core.management.base import BaseCommand
+from django.utils.translation import ugettext_lazy
+
+
+def drop_table(table):
+    cursor = connection.cursor()
+    # Using DROP instead od Delete to remove entire table from DB
+    # Calling cascade to delete data referenced from other tables
+    sql = f"DROP TABLE {table} CASCADE;"
+    cursor.execute(sql)
+
+
+class Command(BaseCommand):
+    help = ugettext_lazy("Drop Formbuilder tables and data from Onadata DB")
+
+    def handle(self, *args, **kwargs):
+        kpi_tables = [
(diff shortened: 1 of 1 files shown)


onadata/apps/main/management/commands/remove_kpi_tables.py
```diff
@@ -0,0 +1,81 @@
...
+    def handle(self, *args, **kwargs):
+        kpi_tables = [
+            'django_migrations',
+            'django_content_type',
+            'auth_user',
+            'auth_group',
+            'auth_permission',
+            'auth_group_permissions',
+            'auth_user_groups',
+            'auth_user_user_permissions',
```
WinnyTroy, line 26: Although these auth tables they were required when doing the data migration, I'm weary on droping the entire table.
There could be records referenced while dropping the other tables, but I believe we would be cascade deleting those specific records

WinnyTroy:
> Some of these are regular Django tables. Will deleting them not break Onadata?

@moshthepitt I have highlighted afew tables I had in mind that may fail with this,
Apart from those, do you mind sharing which others you feel would break the Onadata service?

onadata/apps/main/management/commands/remove_kpi_tables.py
WinnyTroy: I'll commit out the auth tables.

moshthepitt: So the way the table names work is that the app name is included e.g. `auth_user` means that this is the User model from the auth app.

You can use that scheme to only target tables from the kpi app (and any other specific app).

WinnyTroy:
Closing this out after discussions with @ivermac on scope and implementation
This needs further discussion to introduce a policy to delete users and their resources both on Onadata and Formbuilder. This issue has been defined on the roadmap

PR #2029, opened 2021-03-02
Fix "Different root node name" issue

DavisRayM:

- Return data without modification if data is not a list; _Avoids encapsulating the XML within the `submission-batch` root tag_

- Q/A

- N/A

Closes #


ivermac (approved):
Looks good to me

PR #2039, opened 2021-03-18
Add support for sort and handle streaming of empty datasets on XML Responses

### LINKED ISSUE #2000: Data Endpoint: Support pagination of XML Response

DavisRayM:

- Handle streaming empty datasets using XML Renderer
- Add `_date_modified` to the NONE_JSON_FIELD list
- Sort none & filter none JSON fields using the model field

- Added tests

- N/A

Closes #2000

### DIFF
--- a/onadata/libs/models/sorting.py
+++ b/onadata/libs/models/sorting.py
@@ -1,5 +1,6 @@
 import json
 import six
+from typing import Dict
 
 
 def sort_from_mongo_sort_str(sort_str):
@@ -21,11 +22,16 @@ def sort_from_mongo_sort_str(sort_str):
     return sort_values
 
 
-def json_order_by(sort_list):
+def json_order_by(
+        sort_list, none_json_fields: Dict = {}, model_name: str = ""):
     _list = []
 
     for field in sort_list:
-        _str = u" json->>%s"
+        field_key = field.lstrip('-')
+        _str = u" json->>%s"\
+            if field_key not in none_json_fields.keys() else\
+            f'"{model_name}"."{none_json_fields.get(field_key)}"'
+
         if field.startswith('-'):
             _str += u" DESC"
         else:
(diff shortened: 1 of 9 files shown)


onadata/libs/models/sorting.py
```diff
@@ -21,11 +22,16 @@ def sort_from_mongo_sort_str(sort_str):
...
     _list = []
 
     for field in sort_list:
-        _str = u" json->>%s"
+        field_key = field.lstrip('-')
+        _str = u" json->>%s"\
+            if field_key not in none_json_fields.keys() else\
+            f'"{model_name}"."{none_json_fields.get(field_key)}"'
+
         if field.startswith('-'):
```
ukanga, line 35: Should we add or update the unit tests for the `json_order_by()` function?

onadata/libs/models/sorting.py
```diff
@@ -38,10 +44,12 @@ def json_order_by(sort_list):
...
 
-def json_order_by_params(sort_list):
+def json_order_by_params(sort_list, none_json_fields: Dict = {}):
     params = []
 
     for field in sort_list:
-        params.append(field.lstrip('-'))
+        field = field.lstrip('-')
+        if field not in none_json_fields.keys():
+            params.append(field.lstrip('-'))
```
ukanga, line 53: Should we add or update the unit tests for the `json_order_by_params()` function?

ukanga (commented):
Could we also revisit `<submission-item created=` to use `<submission-item dateCreated=` to be more consistent in how the attributes are defined?

onadata/libs/models/sorting.py
DavisRayM: Added in the latest commits

onadata/libs/models/sorting.py
DavisRayM: Added in the latest commits

ukanga (approved):

gstuder-ona:
@DavisRayM sorry I didn't get to this earlier, but we're using the endpoint from stage successfully now

PR #2044, opened 2021-03-25
2037 add review date

### LINKED ISSUE #2037: Add submission review date on data endpoint
Just as we have the review comment/status, we would like to have a review date as well on the data endpoint
data-endpoint

WinnyTroy:
Adds submission review dates to data endpoint response

Updated tests
QA

Closes #2037

### DIFF
--- a/onadata/apps/logger/models/instance.py
+++ b/onadata/apps/logger/models/instance.py
@@ -5,6 +5,7 @@ Instance model class
 import math
 import pytz
 from datetime import datetime
+from deprecated import deprecated
 
 from django.conf import settings
 from django.contrib.auth.models import User
@@ -444,11 +445,11 @@ class InstanceBaseClass(object):
         # pylint: disable=no-member
         return [note.get_data() for note in self.notes.all()]
 
+    @deprecated(version='2.5.3',
+                reason="Deprecated in favour of `get_latest_review`")
     def get_review_status_and_comment(self):
         """
         Return a tuple of review status and comment.
-        Deprecated in favour of `get_latest_review`
-        TODO: Clean code of this unused function.
         """
         try:
             # pylint: disable=no-member
--- a/onadata/apps/logger/models/instance.py
+++ b/onadata/apps/logger/models/instance.py
@@ -380,7 +380,7 @@ class InstanceBaseClass(object):
                 doc[REVIEW_STATUS] = status
                 if comment:
                     doc[REVIEW_COMMENT] = comment
-                doc[REVIEW_DATE] = review_date
+                doc[REVIEW_DATE] = review_date.strftime(MONGO_STRFTIME)
 
             # pylint: disable=attribute-defined-outside-init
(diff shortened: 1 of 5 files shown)


onadata/apps/logger/models/instance.py
```diff
@@ -441,15 +442,17 @@ def get_notes(self):
...
-            status = self.reviews.latest('date_modified').status
-            comment = self.reviews.latest('date_modified').get_note_text()
-            return status, comment
+            review = self.reviews.latest('date_modified')
+            status = review.status
+            comment = review.get_note_text()
+            review_date = review.date_created
+            return status, comment, review_date
         except SubmissionReview.DoesNotExist:
             return None
```
ukanga, line 457: Seems the None portion is not being tested, it could be a source of potential bug, I would expect 3 None's to be returned.

onadata/apps/logger/models/instance.py
```diff
@@ -441,15 +442,17 @@ def get_notes(self):
         # pylint: disable=no-member
         return [note.get_data() for note in self.notes.all()]
 
-    def get_review_status_and_comment(self):
+    def get_review_details(self):
```
ukanga, line 445: I tend to be hesitant in renaming functions that have been around for a long time, in this case about 3 years. I would perhaps recommend a new method:

```python
def get_latest_review(self):
    try:
        return self.reviews.latest('date_modified')
    except SubmissionReview.DoesNotExist:
        return None
````
Which then we can use as
```python
review = self.get_latest_review()
if review:
    doc[REVIEW_STATUS] = review.status
    doc[REVIEW_COMMENT] = review.comment
    doc[REVIEW_DATE] = review.date_created.strftime(MONGO_STRFTIME)
```
I think this approach is much cleaner and clearer.

ukanga: It also means you do not end up having to rename old functions, you could mark the old function deprecated so that we eventually change its reference everywhere.

onadata/apps/logger/models/instance.py
```diff
@@ -443,7 +446,9 @@ def get_notes(self):
 
     def get_review_status_and_comment(self):
         """
-        Return a tuple of review status and comment
+        Return a tuple of review status and comment.
+        Deprecated in favour of `get_latest_review`
+        TODO: Clean code of this unused function.
```
DavisRayM, line 451: I wonder if we should use something like the [Deprecated](https://pypi.org/project/Deprecated/) library or a [custom decorator](https://stackoverflow.com/a/30253848) like this; to mark deprecated functions ?

_TODOs usually get forgotted IMO_

WinnyTroy: Thanks for this.
I've used the pip deprecated package for this. Seemed cleaner.

ukanga (approved):

DavisRayM (approved):

PR #2068, opened 2021-05-03
Add pagination for the messaging endpoint

### LINKED ISSUE #2064: Introduce pagination on the messaging endpoint

As of now the messaging endpoint(`/api/v1/messaging`) has no pagination; All the data available it returned in one request. In order to improve performance in the future it would serve as well to introduce pagination on the endpoint and set a reasonable default limit for the number of notifications that can be retrieved in one request i.e 10,000 messages max ?

- Better performance for forms that might have a lot of messages/notifications

N/A

## Suggested additions

- Addition of `page` & `page_size` query params
- Addition of a configurable default number of records that can be returned from the endpoint when `page` & `page_size` is not defined
- Inclusion of the `Link` header in response to allow for easier navigation

DavisRayM:

- Add pagination for the messaging endpoint. Added `page` and `page_size` query params
- Add message retrieval limit
- Add `Link` header to paginated responses
- Add `generate_pagination_headers` utility function

- Add/Updated tests

The messaging viewset will now not return all the messages that are currently available if the number of messages surpasses the retrieval threshold.

- [x] Included tests 
- [x] Updated documentation

Closes #2064

### DIFF
--- a/onadata/apps/messaging/viewsets.py
+++ b/onadata/apps/messaging/viewsets.py
@@ -17,7 +17,8 @@ from onadata.apps.messaging.filters import (
     UserFilterBackend)
 from onadata.apps.messaging.permissions import TargetObjectPermissions
 from onadata.apps.messaging.serializers import MessageSerializer
-from onadata.libs.pagination import StandardPageNumberPagination
+from onadata.libs.pagination import (
+    StandardPageNumberPagination, generate_pagination_headers)
 
 
 # pylint: disable=too-many-ancestors
@@ -38,6 +39,7 @@ class MessagingViewSet(mixins.CreateModelMixin, mixins.ListModelMixin,
 
     def list(self, request, *args, **kwargs):
         queryset = self.filter_queryset(self.get_queryset())
+        no_of_records = queryset.count()
         retrieval_threshold = getattr(
             settings, "MESSAGE_RETRIEVAL_THRESHOLD", 10000)
         pagination_keys = [self.paginator.page_query_param,
@@ -45,7 +47,14 @@ class MessagingViewSet(mixins.CreateModelMixin, mixins.ListModelMixin,
         query_param_keys = self.request.query_params
         should_paginate = any(
             [k in query_param_keys for k in pagination_keys]) or \
-            queryset.count() > retrieval_threshold
+            no_of_records > retrieval_threshold
+        current_page_size = self.request.query_params.get(
+            self.paginator.page_size_query_param) or retrieval_threshold
+        current_page = self.request.query_params.get(
+            self.paginator.page_query_param) or 1
+        headers = generate_pagination_headers(
+            self.request, no_of_records, current_page_size, current_page
+        )
 
         if should_paginate and \
                 "page_size" not in self.request.query_params.keys():
@@ -57,4 +66,4 @@ class MessagingViewSet(mixins.CreateModelMixin, mixins.ListModelMixin,
         else:
             serializer = self.get_serializer(queryset, many=True)
 
-        return Response(serializer.data)
+        return Response(serializer.data, headers=headers)
--- a/onadata/apps/messaging/viewsets.py
+++ b/onadata/apps/messaging/viewsets.py
@@ -8,6 +8,7 @@ from actstream.models import Action
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import mixins, viewsets
 from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
 
 from onadata.apps.messaging.constants import MESSAGE_VERBS
 from onadata.apps.messaging.filters import (
@@ -15,6 +16,7 @@ from onadata.apps.messaging.filters import (
     UserFilterBackend)
 from onadata.apps.messaging.permissions import TargetObjectPermissions
 from onadata.apps.messaging.serializers import MessageSerializer
+from onadata.libs.pagination import StandardPageNumberPagination
 
 
 # pylint: disable=too-many-ancestors
@@ -31,3 +33,15 @@ class MessagingViewSet(mixins.CreateModelMixin, mixins.ListModelMixin,
     filter_backends = (TargetTypeFilterBackend, TargetIDFilterBackend,
                        UserFilterBackend, DjangoFilterBackend)
     filterset_class = ActionFilterSet
+    pagination_class = StandardPageNumberPagination
+
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+        else:
+            serializer = self.get_serializer(queryset, many=True)
+
+        return Response(serializer.data)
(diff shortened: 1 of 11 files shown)


onadata/apps/messaging/viewsets.py
```diff
@@ -31,3 +35,35 @@ class MessagingViewSet(mixins.CreateModelMixin, mixins.ListModelMixin,
...
+
+        if should_paginate and \
+                "page_size" not in self.request.query_params.keys():
+            self.paginator.page_size = retrieval_threshold
+
+        if should_paginate:
+            page = self.paginate_queryset(queryset)
+            serializer = self.get_serializer(page, many=True)
+        else:
+            serializer = self.get_serializer(queryset, many=True)
```
denniswambua, line 67: Does this return everything if pagination params are not provided?

DavisRayM: It'll return everything if the threshold is not passed i.e if the threshold is 10,000 records; It'll return everything up to 10,000 records and provide pagination headers for easier navigation

onadata/libs/pagination.py
```diff
@@ -57,3 +62,73 @@ def paginate_queryset(self, queryset, request, view, count=None):
...
+    base_url = f"{url.scheme}://{url.netloc}{url.path}?"
+    if url.query:
+        query_params = parse_qs(url.query)
+        query_string = None
+        for param_key, param_values in query_params.items():
+            if param_key not in ['page_size', 'page']:
+                param_value = ','.join(param_values)
+                query = f'{param_key}={param_value}&'
+                if not query_string:
+                    query_string = query
```
denniswambua, line 81: Can this just be
```
query_string = query
```

onadata/libs/pagination.py
```diff
@@ -57,3 +62,73 @@ def paginate_queryset(self, queryset, request, view, count=None):
...
+    if current_page != 1:
+        first_page_url = (
+            f"{base_url}page=1&page_size={current_page_size}"
+        )
+
+    for rel, link in (
+            ('prev', prev_page_url),
+            ('next', next_page_url),
+            ('last', last_page_url),
+            ('first', first_page_url)):
```
WinnyTroy, line 130: For this, I'm mostly curious, I'm not sure which data structure this is 
```code
(('prev', prev_page_url),
('next', next_page_url),
('last', last_page_url),
('first', first_page_url))
```
Why did we use this struct over an array of tuples?

onadata/libs/pagination.py
```diff
@@ -57,3 +62,73 @@ def paginate_queryset(self, queryset, request, view, count=None):
...
+) -> dict:
+    url = urlparse(request.build_absolute_uri())
+    base_url = f"{url.scheme}://{url.netloc}{url.path}?"
+    if url.query:
+        query_params = parse_qs(url.query)
+        query_string = None
+        for param_key, param_values in query_params.items():
+            if param_key not in ['page_size', 'page']:
+                param_value = ','.join(param_values)
+                query = f'{param_key}={param_value}&'
```
WinnyTroy, line 79: Ii've tested this locally, after looping through all the query params passed with the request, the result url seems to appear as such
``` code
'http://localhost:8000/api/v1/messaging?target_type=xform&target_id=42&'
```
The trailing `&`, is that something that may be a proble in future?

WinnyTroy, line 79: I've tested this locally, after looping through all the query params passed with the request, the result  base url is
``` code
'http://localhost:8000/api/v1/messaging?target_type=xform&target_id=42&'
```
The trailing `&`, is that something that may be a problem in future?

onadata/libs/pagination.py
```diff
@@ -57,3 +62,73 @@ def paginate_queryset(self, queryset, request, view, count=None):
 
         self.request = request
         return list(self.page)
+
+
+def generate_pagination_headers(
+    request: Request, no_of_records: int, current_page_size: Union[int, str],
+    current_page: Union[int, str]
+) -> dict:
```
WinnyTroy, line 70: Could we maybe add docstings for this helper function?
I am also curious, with regards to the `current_page_size` and `current_page` parameters we pass to this function. Why/When would we expect this to get returned as a string? I see from the tests that they are passed in as ints, also in the same function on line 93 and 99, we specifically convert them to ints, so why the Union?

onadata/apps/messaging/viewsets.py
```diff
@@ -31,3 +35,35 @@ class MessagingViewSet(mixins.CreateModelMixin, mixins.ListModelMixin,
...
+        should_paginate = any(
+            [k in query_param_keys for k in pagination_keys]) or \
+            no_of_records > retrieval_threshold
+        current_page_size = self.request.query_params.get(
+            self.paginator.page_size_query_param) or retrieval_threshold
+        current_page = self.request.query_params.get(
+            self.paginator.page_query_param) or 1
+        headers = generate_pagination_headers(
+            self.request, no_of_records, current_page_size, current_page
+        )
```
WinnyTroy, line 57: It seems this is run everytime we hit this endpoint.
I see above in line 48, we define if we `should_paginate` flag.
I'm curious if there is need to generate the pagination headers if `should_paginate` is False?

onadata/libs/pagination.py
DavisRayM: It's a "nested" tuple in some sense... not entirely sure if that's the best way to call it though. Using this since it seems easier to loop through it's tuple objects and values. I could use a struct or dict in this case but this seems simpler IMO

onadata/libs/pagination.py
DavisRayM: It's not really a problem; that base_url is only used to generate pagination links as such if there is no need to create one of the pagination links it won't be used anywhere & if it needs to be used the pagination query params will be appended after the `&`

onadata/libs/pagination.py
DavisRayM: In some cases(especially when the value is retrieved from `request.query_param` the value is a str; this is just a way to support that it's not really necessary to have though). I'll add a docstring to this

onadata/apps/messaging/viewsets.py
DavisRayM: No, it shouldn't. I'll update the code

onadata/libs/pagination.py
```diff
@@ -57,3 +62,73 @@ def paginate_queryset(self, queryset, request, view, count=None):
...
+
+    if isinstance(current_page_size, str):
+        try:
+            current_page_size = int(current_page_size)
+        except ValueError:
+            return
+
+    if (current_page * current_page_size) < no_of_records:
+        next_page_url = (
+            f"{base_url}page={current_page + 1}&"
```
denniswambua, line 107: I see the parent class has `get_next_link` Could we re-use that functionality here and the get previous link
Ref https://github.com/encode/django-rest-framework/blob/master/rest_framework/pagination.py#L271

DavisRayM: This function isn't really in the Pagination class at the moment... I can add it to the Pagination class though

denniswambua: i meant in the `PageNumberPagination` class

denniswambua (approved):

PR #2097, opened 2021-06-02
Add `service_health` view function

### LINKED ISSUE #2053: Add a health check endpoint

Currently the application provides no way of knowing whether all the core parts(Database connection, Cache connection, e.t.c) are working as expected. It would help quite a lot if an endpoint was added that provided such information rather than requiring a user to hit a data endpoint or something of the sort

- Easier to retrieve application status

- N/A

DavisRayM:

- Add `service_health` view function
- Register new `/health` endpoint
- Modify docker-compose file

```
Add `notification` service & Configure the service as a notification backend; Users of the docker compose file can now subscribe to the service via mosquitto pub or any other mqtt client and listen to the notification
```

- Add `django-redis` as a requirement

N/A

- [x] Included tests 
- [ ] ~Updated documentation~

Closes #2053


onadata/apps/main/views.py
```diff
@@ -1423,6 +1424,41 @@ def enketo_preview(request, username, id_string):
...
+        except Exception as e:
+            service_statuses[f'{database}-Database'] = f'Degraded state; {e}'
+            service_degraded = True
+        else:
+            service_statuses[f'{database}-Database'] = 'OK'
+
+    # Check if cache is accessible
+    try:
+        cache.set('ping', 'pong')
+        service_degraded = not (cache.get('ping') == 'pong')
```
denniswambua, line 1449: Do we really want to restart a pod or create a new pod when we don't have access to the cache?

DavisRayM: I believe so; most of the processing intensive functionalities like user permission generation & other project calculations are stored and tracked using the cache. I think if their down we'd experience some impact on the service... though this is just an assumption

onadata/apps/main/urls.py
```diff
@@ -335,7 +335,10 @@
     re_path(r'^favicon\.ico',
             RedirectView.as_view(url='/static/images/favicon.ico',
                                  permanent=True)),
-    re_path(r'^static/(?P<path>.*)$', staticfiles_views.serve)
+    re_path(r'^static/(?P<path>.*)$', staticfiles_views.serve),
+
+    # Health check
+    re_path(r'^service_health$', main_views.service_health)
```
denniswambua, line 341: Can we just use `/health` endpoint

DavisRayM: Renamed in the latest commit

onadata/apps/main/views.py
denniswambua: It is better to have a degraded service than not to have one all. Because if we lose the cache we will lose the whole application, kubernetes will mark all the pod unhealthy

denniswambua (approved):

PR #2113, opened 2021-06-29
Ensure Pagination and sorting are implemented on the data endpoint

### LINKED ISSUE #2108: Sorted submissions are not paginated correctly
### Environmental Information

- Onadata version: `v2.5.7`

### Problem description
Submissions on data endpoint are not paginated correctly when you pass `sort` query parameter

Without `sort` query parameter:

Screenshot from 2021-06-22 12-47-08

With `sort` query parameter:
Screenshot from 2021-06-22 12-46-51

### Expected behavior
Pagination query parameters should work the same with or without the order query parameters

WinnyTroy:
- Ensure sort filter is applied for paginated responses from the data endpoint.

- [x] Included tests

- [x] Included tests 

Closes #2108

### DIFF
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -649,8 +649,11 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
             if should_paginate:
                 self.paginator.page_size = retrieval_threshold
 
-        if not isinstance(self.object_list, types.GeneratorType) and \
+        if isinstance(self.object_list, types.GeneratorType) and \
                 should_paginate:
+            # Unpack generator object to list
+            self.object_list = list(self.object_list)
+        else:
             current_page = query_param_keys.get(
                 self.paginator.page_query_param, 1)
             current_page_size = query_param_keys.get(
@@ -663,10 +666,10 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                 current_page_size=current_page_size
             )
 
-            try:
-                self.object_list = self.paginate_queryset(self.object_list)
-            except OperationalError:
-                self.object_list = self.paginate_queryset(self.object_list)
+        try:
+            self.object_list = self.paginate_queryset(self.object_list)
+        except OperationalError:
+            self.object_list = self.paginate_queryset(self.object_list)
 
         STREAM_DATA = getattr(settings, 'STREAM_DATA', False)
         if STREAM_DATA:
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -649,6 +649,7 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
             if should_paginate:
                 self.paginator.page_size = retrieval_threshold
 
+        # import ipdb; ipdb.set_trace()
         if isinstance(self.object_list, types.GeneratorType) and \
                 should_paginate:
             # Unpack generator object to list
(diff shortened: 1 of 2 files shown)


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -647,8 +647,11 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
             if should_paginate:
                 self.paginator.page_size = retrieval_threshold
 
-        if not isinstance(self.object_list, types.GeneratorType) and \
+        if isinstance(self.object_list, types.GeneratorType) and \
                 should_paginate:
+            # Unpack generator object to list
+            self.object_list = list(self.object_list)
+        else:
```
DavisRayM, line 654: Seems like the pagination headers are skipped if the object_list is a generator object?

WinnyTroy: Yes.
I assume, from the previously changes I observed a scenerio whereby data returned in JSON format was to be streamed, with no pagination.
I left the else condition to continue to handle such kinds of requests, i.e that the data returned is a generator objects and we do not want to paginate the returned response data.
A good example can be found on [this](https://github.com/onaio/onadata/blob/e2ca6f770466c42773a87031a3bcf1b6966b98e4/onadata/apps/api/tests/viewsets/test_data_viewset.py#L755) test, we first return a generator object, and we dont want to paginate it, but use the limit flag instead.

DavisRayM: IMO we should always paginate requests if requested no matter if we're streaming the data... It's still going to be a pain if a user requests a lot of records while streaming data is enabled

DavisRayM: We also have a few tests that use pagination and stream data https://github.com/onaio/onadata/blob/e2ca6f770466c42773a87031a3bcf1b6966b98e4/onadata/apps/api/tests/viewsets/test_data_viewset.py#L653-L654

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -647,8 +647,10 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
             if should_paginate:
                 self.paginator.page_size = retrieval_threshold
 
-        if not isinstance(self.object_list, types.GeneratorType) and \
-                should_paginate:
+        if isinstance(self.object_list, types.GeneratorType):
+            # Unpack generator object to list
+            self.object_list = list(self.object_list)
+        elif should_paginate:
```
DavisRayM, line 653: What happens if it's a generator and the user asked for pagination ?

WinnyTroy: It will mostly result to a generator object if sort query param is passed for the JSON result.
If `page` and `page_size` headers are passed, the result will be unpacked and paginated appropriately.

DavisRayM: Still seems like if it's a generator object the pagination headers won't be included.... Seems like it's unpacked but the pagination headers are not included

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -647,8 +647,10 @@ def _get_data(self, query, fields, sort, start, limit, is_public_request):
             if should_paginate:
                 self.paginator.page_size = retrieval_threshold
 
-        if not isinstance(self.object_list, types.GeneratorType) and \
-                should_paginate:
+        if isinstance(self.object_list, types.GeneratorType):
+            # Unpack generator object to list
+            self.object_list = list(self.object_list)
```
denniswambua, line 652: This converts the stream to a list and breaks the stream data functionality

denniswambua: Can we paginate when fetching from the db?

denniswambua (changes requested):

denniswambua (dismissed):

denniswambua (approved):

Issue #2104, opened 2021-06-11
Include management command to convert stored oauth2client creds to google-auth creds


WinnyTroy:
## Suggested Feature / Enhancement
Create management command to convert stored oauth2client credentials to google-auth creds
Populate the Oauth2TokenStorage model with google-auth creds

## Benefits of implementing the feature/enhancement
New google-auth creds are essential in successfully completing the Oauth2 Flow initialized by the google-auth lib

Part of #1992

PR #2128, opened 2021-07-23
Create oauth2 token storage model…

### LINKED ISSUE #2103: Create Oauth2TokenStorage model
- This new model is to store oauth2 (openid? or google) credentials

- Allow us to easily and efficiently migrate from deprecated oauth2client credentials

Part of #1992

WinnyTroy:
Create oauth2 token storage model for google auth credentials persistent storage

- New migrations for an empty db table 

Closes #2103


onadata/apps/api/models/oauthtokenstorage.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+Oauth2 token storage model class
+"""
+
+from django.db import models
+from jsonfield import JSONField
+from django.conf import settings
+
+
+class OauthStorageModel(models.Model):
```
DavisRayM, line 11: Is the plan for this model to be used to store different OAuth credentials? Not only google? MIght be nice to have an easy way to differentiate which provider the credentials are for i.e google, microsoft, e.t.c

WinnyTroy: At the moment, its mainly for the google Oauth2, but any other Oauth2 provider credentials may possibly also be persisted within this model.
How would you suggest we differentiat the provider?, include a separate third text field for oauth provider? - I dont see why not

onadata/apps/api/migrations/0009_oauthstoragemodel_provider.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.23 on 2021-07-28 13:49
+
```
DavisRayM, line 2: Lets combine the migrations in this PR into one migration

denniswambua:
@WinnyTroy Do we need a script to migrate the old tokens to this new token storage?

onadata/apps/api/migrations/0009_oauthstoragemodel_provider.py
WinnyTroy: done

WinnyTroy:
@denniswambua populating this model will be handled in [this](https://github.com/onaio/onadata/issues/2104) separate issue.

DavisRayM:
Superseded by #2236

PR #2132, opened 2021-08-10
Ensure exports are generated on submission delete

### LINKED ISSUE #2130: Update exports when records are deleted
### Environmental Information

- Onadata version: v2.5.8

### Problem description

Deleted records for forms that had data exports generated before the deletion occurred are not being removed from the created exports. It appears that data exports are not being updated because when the user tries to prepare an export after deleting records, no new export is generated. The user is redirected to the previously created export which includes the deleted records. This issue is affecting other functionalities like, pulling data in Ona R and pulling data from linked datasets where deleted records in the linked form are still available.

### Expected behavior

When records are deleted from a form, any exports generated before the deletion happened should be updated to remove the deleted records. We could also have a new data export generated altogether in such cases. 

- Create a data export in any of the exports formats
- Delete some records from the form
- Try creating a new export in any of the formats used in step 1
- Use this [test form](https://docs.google.com/spreadsheets/d/1-7exP8rhR3YQLI2GxyhhK2nXd6P4tiD7ZSgy7WzX0ek/edit#gid=41563628) that's pulling data from this other [test form](https://docs.google.com/spreadsheets/d/1Th_3VBq6OBuxxS5DciqBW-ZWvrGFeKZf2-gZzHP-IJ8/edit#gid=940000897). 

### Additional Information

DavisRayM:

- Ensure exports are generated when a submission is deleted

- Added test

N/A

- [x] Included tests 
- [ ] Updated documentation

Closes #2130

### DIFF
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -312,6 +312,7 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                     self.object.instances.filter(
                         deleted_at__isnull=True).update(
                             deleted_at=timezone.now(),
+                            date_modified=timezone.now(),
                             deleted_by=request.user)
                 else:
                     instance_ids = [
@@ -327,6 +328,7 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
                         deleted_at__isnull=True
                     ).update(
                         deleted_at=timezone.now(),
+                        date_modified=timezone.now(),
                         deleted_by=request.user)
 
                 # updates the num_of_submissions for the form.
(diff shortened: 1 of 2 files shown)


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -312,6 +312,7 @@ def destroy(self, request, *args, **kwargs):
                     self.object.instances.filter(
                         deleted_at__isnull=True).update(
                             deleted_at=timezone.now(),
+                            date_modified=timezone.now(),
                             deleted_by=request.user)
```
denniswambua, line 316: @DavisRayM date_modified is supposed to be auto-filled by the database

DavisRayM: @denniswambua `date_modified` is [auto filled by django](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.DateField.auto_now) when you call the `save()` function. `update()` usually doesn't call the `save()` function

denniswambua (approved):

Mstjamush:
@DavisRayM 
On deletion, there's no consistency on the items deleted and the number of remaining records. 
E.g on deleting one record, the total number of records drops by more than 3 record. Sometimes 3, sometimes 10 or more

PR #2136, opened 2021-08-18
Delete xform linked resources upon xform deletion

### LINKED ISSUE #1029: Deleting a form should also apply to linked resources like linked media files and data views.

WinnyTroy:
- Clean out Merged Datasets upon xform deletion.
- Clean out Form Media Files upon form deletion.
- Add documentation on clean up of filtered datasets

- [x] Updated tests 

Closes #1029

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_metadata_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_metadata_viewset.py
@@ -192,6 +192,35 @@ class TestMetaDataViewSet(TestAbstractViewSet):
             self.assertEqual(response.status_code, 204)
             self.assertEqual(count, MetaData.objects.count())
 
+    def test_delete_xform_deletes_media_metadata(self):
+        self._add_test_metadata()
+        self.view = MetaDataViewSet.as_view({'get': 'list'})
+        data = {'xform': self.xform.pk}
+        request = self.factory.get('/', data, **self.extra)
+        response = self.view(request)
+        meta_obj = self.xform.metadata_set.all()[0]
+        expected_metadata = {
+            'id': meta_obj.id,
+            'xform': self.xform.id,
+            'data_value': 'screenshot.png',
+            'data_type': meta_obj.data_type,
+            'data_file': f'http://localhost:8000/media/{meta_obj.data_file.name}',  # noqa
+            'data_file_type': 'image/png',
+            'media_url': None,
+            'file_hash': 'md5:09b9e5e3278ac2a43fce25681ffada85',
+            'url': f'http://testserver/api/v1/metadata/{meta_obj.id}',
+            'date_created': meta_obj.date_created
+            }
+        self.assertEqual(response.status_code, 200)
+        self.assertEquals(dict(response.data[0]), expected_metadata)
+
+        # Soft delete xform
+        self.xform.soft_delete()
+        # Confirm that all metadata was deleted
+        response2 = self.view(request)
+        self.assertEqual(response2.status_code, 200)
+        self.assertEquals(response2.data, [])
+
     def test_windows_csv_file_upload_to_metadata(self):
         data_value = 'transportation.csv'
         path = os.path.join(self.fixture_dir, data_value)
--- a/onadata/apps/logger/models/xform.py
+++ b/onadata/apps/logger/models/xform.py
@@ -953,8 +953,17 @@ class XForm(XFormMixin, BaseModel):
             update_fields.append('deleted_by')
 
         self.save(update_fields=update_fields)
+        # Delete associated filtered datasets
         for dataview in self.dataview_set.all():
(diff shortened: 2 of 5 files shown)


onadata/apps/logger/models/xform.py
```diff
@@ -953,8 +953,17 @@ def soft_delete(self, user=None):
...
         for dataview in self.dataview_set.all():
             dataview.soft_delete(user)
+        # Delete associated Merged-Datasets
+        for merged_dataset in self.mergedxform_ptr.all():
+            merged_dataset.soft_delete(user)
+        # Delete associated Form Media Files
+        for metadata in self.metadata_set.all():
+            # Only target media metadata
+            if metadata.data_type == 'media':
+                metadata.delete()
```
DavisRayM, line 966: We should look into soft-deleting metadata perhaps? This seems like it would cause issues when the soft-deleted XForm is restored...

onadata/apps/logger/models/xform.py
```diff
@@ -953,8 +953,17 @@ def soft_delete(self, user=None):
             update_fields.append('deleted_by')
 
         self.save(update_fields=update_fields)
+        # Delete associated filtered datasets
         for dataview in self.dataview_set.all():
             dataview.soft_delete(user)
```
DavisRayM, line 958: I wonder if we have a class function that can be used to un-delete an XForm? Seems like a nice to have IMO since this complicates the recovery process a bit

DavisRayM (commented):
Looks good just a few concerns

onadata/apps/logger/models/xform.py
WinnyTroy: Yeah, so these changes do increase complexity if a form is to be restored. I agree that we may need a management command that can reverse this, maybe a different issue can capture this?
To uncomplicate this though, I felt the inline comments help one understand the difference of linked datasets and merged datasets.
I also only removed the form media files, I left the other form metadata e.g enketo links, in the event a user may want to have the form restored.

onadata/apps/logger/models/xform.py
WinnyTroy: I considered this, and I saw it wouldnt really be necessary.
Metadata holds form extras link enketo links and submission review enabling/disabling. These are often never deleted.
Its only now we are introducing deleting of media files associated with a form, and I dont believe we would want to restore these?, maybe have the user re-create them? I could be wrong

DavisRayM: I think it'll lead to a bad user experience... Users should be aware that deleting their data does have some consequences but yea I'm not entirely sure about this

Thoughts @denniswambua ?

onadata/apps/logger/models/xform.py
DavisRayM: Yea, we don't have to implement a migration command for this... I think an instance method should do the trick. I'm fine with this being included in a separate PR.

I think re-creating the enketo links is something we can easily do on restoration of the form... The media files I'm iffy on

onadata/apps/logger/models/xform.py
denniswambua: @WinnyTroy Not all metadata can be regenerated. So we will need the soft delete on metadata.

WinnyTroy: @denniswambua  made this change

onadata/apps/api/tests/viewsets/test_metadata_viewset.py
```diff
@@ -192,6 +192,18 @@ def test_delete_metadata(self):
...
+        for data_type in ['supporting_doc', 'media', 'source']:
+            self._add_form_metadata(self.xform, data_type,
+                                    self.data_value, self.path)
+        # Soft delete xform
+        self.xform.soft_delete()
+        # Confirm that media files were deleted
+        form_metadata = [
+            mt.data_type for mt in self.xform.metadata_set.filter(
+                deleted_at__isnull=True)]
+        self.assertNotIn('media', form_metadata)
```
DavisRayM, line 205: Shouldn't we be checking that the response from the viewset in this case is empty ?

DavisRayM (approved):

onadata/apps/logger/models/xform.py
```diff
@@ -953,8 +953,15 @@ def soft_delete(self, user=None):
...
+        # Delete associated Merged-Datasets
+        for merged_dataset in self.mergedxform_ptr.all():
+            merged_dataset.soft_delete(user)
+        # Delete associated Form Media Files
+        for metadata in self.metadata_set.all():
+            metadata.soft_delete()
```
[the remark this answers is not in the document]
WinnyTroy: Great catch! Thanks
Made this change.

ukanga (dismissed):

denniswambua (approved):

PR #2137, opened 2021-08-19
Generate xform headers in csv exports for XForms without submissions

WinnyTroy:
- Update functionality to now export xform headers for forms without submissions
- Updates XForms linked datasets creation functionality for forms without submissions. Now exports are generated containing xform headers for these linked media files.

- Updated tests

- [x] Included tests
- [x] Updated older tests
- [ ] Deployed and tested to stage

### How to test these changes.
- Confirm that the data received when hitting the [data endpoint](https://ona.io/static/docs/data.html#stream-xform-submission-data-in-csv-format) for a form without submissions returns the xform headers.
- Generating a csv export for a form without submissions should generate an export file containing xform headers

Closes onaio/zebra#6445


onadata/apps/viewer/tasks.py
```diff
@@ -96,18 +96,25 @@ def _create_export(xform, export_type, options):
...
+                export.save()
+                report_exception(export.error_message, e, sys.exc_info())
+                raise Export.ExportConnectionError
+        else:
+            raise ExportTypeError
     else:
-        raise ExportTypeError
+        export.error_message = "No records found for this export"
+        export.internal_status = Export.FAILED
+        export.save()
```
DavisRayM, line 116: I think we should out-right block the creation of an Export object if the target XForm has no Submissions; The request export task should not be queued

onadata/libs/serializers/metadata_serializer.py
```diff
@@ -184,6 +184,10 @@ def validate(self, attrs):
                             _(u"User has no permission to "
                               "the dataview.")
                         })
+                    # ensure received xform contains submissions
```
DavisRayM, line 187: Why are we blocking a user from adding a media metadata if the XForm has no submissions ? This seems weird to me; In a scenario where a user has an external CSV won't this block them from uploading it ?

DavisRayM (changes requested):

onadata/libs/serializers/metadata_serializer.py
DavisRayM: My bad seems this is the linked dataset...?

denniswambua: It should only restrict linked dataset.

onadata/apps/viewer/tasks.py
ukanga: @DavisRayM When the media file is not generated it means the user may not even be able to download the form for data collection. Generating a file with only headers should be a valid export imo.

onadata/libs/serializers/metadata_serializer.py
ukanga: I think this applies to situation where we are using a linked media file. However, I do agree that we should not prevent a user from creating a link to a separate form due to no submissions on the form they are linking to.

ukanga: Also note, if the form being linked has submissions, it may not always be a guaranteed if we have submissions the user who downloads them will get any data to download. There is a case, where the additional restrictions on the form to be added as media requires that you only access records you are the submitter. In that situation, a user who had not made any submissions to the target form will have zero records.

onadata/apps/viewer/tasks.py
WinnyTroy: Included changes that generate csv export file with only the headers.
Removed the checks below as well.
Thanks for this!

onadata/libs/utils/csv_builder.py
```diff
@@ -551,30 +555,36 @@ def _update_columns_from_data(self, cursor):
...
+                self._tag_edit_string(record)
+                # re index repeats
+                for (key, value) in iteritems(record):
+                    self._reindex(
+                        key, value, self.ordered_columns, record, self.dd,
+                        include_images=image_xpaths,
+                        split_select_multiples=self.split_select_multiples,
+                        index_tags=self.index_tags,
+                        show_choice_labels=self.show_choice_labels,
+                        language=self.language)
```
ukanga, line 581: Are we  handling headers in this function? Could we do without calling this function and hence no need to add the additional check?

WinnyTroy: Sure. This change is breaking existing tests, working to resolve those now

WinnyTroy: Should be good now.

onadata/libs/utils/csv_builder.py
```diff
@@ -533,17 +533,21 @@ def _build_ordered_columns(cls, survey_element, ordered_columns,
...
-                    self.ordered_columns[key] = \
-                        remove_dups_from_list_maintain_order(
-                            [choice.replace('/' + name, '/' + label)
-                             if self.show_choice_labels else choice
-                             for choice, name, label in choices])
+                self.ordered_columns[key] = \
+                    remove_dups_from_list_maintain_order(
+                        [choice.replace('/' + name, '/' + label)
+                         if self.show_choice_labels else choice
+                         for choice, name, label in choices])
```
ukanga, line 550: The if check was removed, is there a change here that gives us confidence hat we will not encounter a `KeyError` exception or are we handling the exception elsewhere?

onadata/libs/utils/csv_builder.py
```diff
@@ -622,38 +609,44 @@ def export_to(self, path, dataview=None):
...
             columns += [col for col in self.extra_columns]
+
             for field in self.dd.get_survey_elements_of_type('osm'):
                 columns += OsmData.get_tag_keys(self.xform,
                                                 field.get_abbreviated_xpath(),
                                                 include_prefix=True)
-            cursor = self._query_data(self.filter_query)
-            if isinstance(cursor, QuerySet):
-                cursor = cursor.iterator()
-            data = self._format_for_dataframe(cursor)
```
ukanga, line 656: It seems we are loosing the optimization of using the `Queryset.iterator()` which in this case was being used to re-evaluate the records afresh since while generating the columns the data might have been traversed entirely and we need to repeat the same traversal while now sending the data to be written.  Did this have to be moved up?

WinnyTroy: I noticed the data was already converted to an iterator in line 633 - 634. This seemed to be like duplication of code.
I also didnt see any consequence by having `data` definition moved up to line 636. I can move it back to the bottom after the columns has been defined though.

onadata/libs/utils/csv_builder.py
WinnyTroy: Ah, that was unintentionally removed. I've reinstated that check.
Thanks for catching that.

onadata/libs/utils/csv_builder.py
```diff
@@ -574,31 +570,24 @@ def _update_columns_from_data(self, cursor):
...
+                    self.VALUE_SELECT_MULTIPLES,
                     show_choice_labels=self.show_choice_labels)
-            # check for gps and split into components i.e. latitude, longitude,
+            # check for gps and split into
+            # components i.e. latitude, longitude,
             # altitude, precision
             self._split_gps_fields(record, self.gps_fields)
             self._tag_edit_string(record)
+            image_xpaths = [] if not self.include_images \
+                else self.dd.get_media_survey_xpaths()
```
ukanga, line 590: This seems to be running inside the loop yet before https://github.com/onaio/onadata/pull/2137/files#diff-f8ec937286aa9f6477be6329d2714eb5e4825a4d2c5e6bd0e8df3a18dd285874L588-L590 it was just outside the loop, is this change necessary?

WinnyTroy: Yeah, it is necessary. The value generated is used in line 596 below in the `self._reindex` function.
I moved this to within the loop during the refactor of the `_format_for_dataframe` function because of this

ukanga: @WinnyTroy in line 596 we have an assignment but not a traversal of the XML like what happens with line 590 for every record being iterated on. Also, 596 has not changed in this PR. So, that is not a movement.

onadata/libs/tests/utils/test_csv_builder.py
```diff
@@ -629,26 +630,11 @@ def test_csv_export_with_labels(self):
             '_review_status', '_review_comment', '_review_date'
         ]
         self.assertEqual(expected_header, header)
-        labels = next(csv_reader)
-        self.assertEqual(len(labels), 17 + len(csv_df_builder.extra_columns))
-        expected_labels = [
```
DavisRayM, line 634: Why do we no longer test for `expected_labels` ?

onadata/libs/tests/utils/test_csv_builder.py
```diff
@@ -629,26 +630,11 @@ def test_csv_export_with_labels(self):
...
-            '_review_comment', '_review_date'
-        ]
-        self.assertEqual(expected_labels, labels)
         rows = []
         for row in csv_reader:
             rows.append(row)
-        self.assertEqual(len(rows), 7)
-        self.assertEqual(rows[4][5], NA_REP)
+        self.assertEqual(len(rows), 8)
+        self.assertEqual(rows[4][16], NA_REP)
```
DavisRayM, line 637: Doesn't seem like the fixture changed... Is there a reason for the additional row ? And Columns ?

DavisRayM: From the PR description it doesn't seem like this functionality should have changed... Kindly update the PR with all the changes that are going in if this is an expected effect of adding the additional columns for exports with no data

WinnyTroy: There is no fixture file for the test. The contents in the temp csv file are generated from the export function. Those contents have been updated. That is the cause of the new columns in the row data

onadata/libs/tests/utils/test_csv_builder.py
WinnyTroy: The list defined here is similar to that of the one defined [here](https://github.com/onaio/onadata/blob/a39361f645fa06cb4d6942b623e3b1a17a56198f/onadata/libs/tests/utils/test_csv_builder.py#L622-L631).
I assumed it was a duplication. I didnt notice they were headers vs names.
Returned that segment to the test.

onadata/libs/tests/utils/test_csv_builder.py
DavisRayM: What new columns were added? Seems there are some fixture instances we submit for the export [here](https://github.com/onaio/onadata/pull/2137/files#diff-71defac856691f0cf9680a9c332085b151b84e2222508c82338a6fde859bb98eL556)

_Still a bit confused about what columns have changed... Thought this was only adding functionality to generate export files with headers when no data has been submitted_

WinnyTroy: Ah, I see where the confusion was coming in. Could be something that was updated on the test while testing.
The test doesnt seem to require this changes though at the moment. Thanks for that.

onadata/libs/utils/csv_builder.py
```diff
@@ -552,18 +557,9 @@ def _update_columns_from_data(self, cursor):
         image_xpaths = [] if not self.include_images \
             else self.dd.get_media_survey_xpaths()
```
DavisRayM, line 558: Seems this variable is stored twice in the `_update_ordered_columns_from_data` & `_format_for_dataframe` functions could we perhaps store them in an object attribute like `self.image_xpaths` and set it only once ? So that we avoid the duplicate call to `get_media_survey_xpaths`

Or set it in the `export_to` function? _Might save us a bit more time for some forms; We won't need to traverse the elements to figure out all the media survey xpaths twice_

DavisRayM (commented):
Minor comment but LGTM

onadata/libs/utils/csv_builder.py
```diff
@@ -622,38 +610,44 @@ def export_to(self, path, dataview=None):
...
-            data = self._format_for_dataframe(cursor)
         else:
-            cursor = self._query_data(self.filter_query)
+            try:
+                cursor = self._query_data(self.filter_query)
+            except NoRecordsFoundError:
+                # Set cursor object to an an empty queryset
+                cursor = self.xform.instances.none()
+                # Define export columns using xform schema
+                columns = self.xform.get_headers()
```
DavisRayM, line 630: Seems this variable isn't used anywhere at the moment and is reset at Line 642 ?

onadata/libs/tests/utils/test_csv_builder.py
```diff
@@ -1209,6 +1210,60 @@ def test_show_choice_labels_select_multiple_2(self, mock_query_data):
...
+        # Generate csv export for form
+        csv_df_builder = CSVDataFrameBuilder(
+            self.user.username, self.xform.id_string, include_images=False)
+        temp_file = NamedTemporaryFile(suffix=".csv", delete=False)
+        csv_df_builder.export_to(temp_file.name)
+        csv_file = open(temp_file.name, 'r')
+        csv_reader = csv.reader(csv_file)
+        header = next(csv_reader)
+
+        expected_header = [
```
DavisRayM, line 1233: We are missing some of the repeat headers i.e `kids_name`, `kids_age`... Should this be the case ? Seems okay IMO but curious...

DavisRayM: Was expecting the headers to be almost identical to:

```
'name', 'age', 'has_kids', 'kids_name', 'kids_age', 'kids_name',
            'kids_age', 'gps', '_gps_latitude', '_gps_longitude',
            '_gps_altitude', '_gps_precision', 'web_browsers/firefox',
            'web_browsers/chrome', 'web_browsers/ie', 'web_browsers/safari',
            'instanceID', '_id', '_uuid', '_submission_time', '_date_modified',
            '_tags', '_notes', '_version', '_duration', '_submitted_by',
            '_total_media', '_media_count', '_media_all_received',
            '_review_status', '_review_comment', '_review_date'
```
but with the group names

WinnyTroy: Yes! I would expect this.
The idea is, if the form contains no submissions, then we only generate the xform headers for export.
I have added an extra test that tests if we generate an export for a form with no submission and then later collect submissions, that the column headers are updated successfully.

onadata/libs/utils/csv_builder.py
WinnyTroy: Made this change. Should be good now

onadata/libs/utils/csv_builder.py
WinnyTroy: This variable is passed to the `write_to_csv` function, but we initially define it here, containing a list of strings(column headers) to be used for the export.
While the variable is defined here, it is updated, not reset, on line 640 and again on line 645 with extra columns for the export

DavisRayM: I'm still unsure how line 640 updates the variable....

```
columns = list(chain.from_iterable(
                [[xpath] if cols is None else cols
                    for (xpath, cols) in iteritems(self.ordered_columns)]))
```

The list comprehension used there just seems to override the initial value and doesn't take into account the what the value of `columns` was at the beginning

onadata/libs/tests/utils/test_csv_builder.py
DavisRayM: But in this case, those aren't all the XForm Headers... So should we perhaps add a caveat that repeat headers will not be included in the empty export ?

WinnyTroy: I dont believe repeat headers should be generated within exports for forms with no submissions. Yes.

DavisRayM: I personally think they should; feels a bit weird to export half the headers because in some scenarios and all in other but I'm fine with whichever approach we decide to take.

Thoughts @ukanga @denniswambua ?

onadata/libs/utils/csv_builder.py
WinnyTroy: Ah! I've seen where the confusion here may be coming from. I have made some changes here in regards to this. Thanks!

onadata/libs/tests/utils/test_csv_builder.py
ukanga: The repeat headers tend to be only available when there is data, it is okay if the repeat headers are not included as for CSV exports the repeats are indexed, always.

DavisRayM (approved):

ukanga (approved):

PR #2142, opened 2021-09-07
Query optimization for the Briefcase viewset

DavisRayM:

- Utilize XForm `num_of_submissions` instead of trying to count all the instances when `numEntries` is not specified
  - Count `instances` using `len()` instead of `count()` to avoid an additional Database query & preload the `instances` before the render
- Only retrieve the `pk` & `uuid` columns from the database instead of all the instance table columns; The briefcase viewset does not utilize any of the other columns

- N/A

Improved performance on the briefcase viewset. Performance gain

- Decreased no. of queries for requests with numEntries query param:
- Decreased no. of queries for requests with the numEntries query param:

_Tests were done on a form with 4601 submissions_

#### With changes & numEntries query param

Screenshot from 2021-09-16 14-00-29

#### Without changes & numEntries query param

Screenshot from 2021-09-16 14-05-16

#### With changes & No query param

Screenshot from 2021-09-16 15-01-48

#### Without changes & No query param

Screenshot from 2021-09-16 15-03-07

- [ ] Included tests 
- [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/api/viewsets/briefcase_viewset.py
+++ b/onadata/apps/api/viewsets/briefcase_viewset.py
@@ -146,8 +146,8 @@ class BriefcaseViewset(mixins.CreateModelMixin,
 
         # Using len() instead of .count() to prevent an extra
         # database call; len() will load the instances in memory allowing
-        # as to use them when generating the response and remove the need
-        # to perform a count on the database.
+        # us to pre-load the queryset before generating the response 
+        # and removes the need to perform a count on the database.
         instance_count = len(instances)
 
         if instance_count:
--- a/onadata/apps/api/viewsets/briefcase_viewset.py
+++ b/onadata/apps/api/viewsets/briefcase_viewset.py
@@ -127,8 +127,9 @@ class BriefcaseViewset(mixins.CreateModelMixin,
             xform_kwargs['user__username__iexact'] = username
         xform = get_form(xform_kwargs)
         self.check_object_permissions(self.request, xform)
-        instances = Instance.objects.filter(xform=xform,
-                                            deleted_at__isnull=True)
+        instances = Instance.objects.filter(
+            xform=xform, deleted_at__isnull=True).values(
+                'pk', 'uuid')
         if xform.encrypted:
             instances = instances.filter(media_all_received=True)
         instances = instances.order_by('pk')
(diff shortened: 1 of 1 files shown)


onadata/apps/api/viewsets/briefcase_viewset.py
```diff
@@ -143,10 +144,16 @@ def filter_queryset(self, queryset):
...
             instances = instances[:num_entries]
 
-        if instances.count():
-            last_instance = instances[instances.count() - 1]
-            self.resumptionCursor = last_instance.pk
-        elif instances.count() == 0 and cursor:
+        # Using len() instead of .count() to prevent an extra
+        # database call; len() will load the instances in memory allowing
+        # as to use them when generating the response and remove the need
+        # to perform a count on the database.
```
DavisRayM, line 150:
```suggestion
        # us to pre-load the queryset before generating the response 
        # and removes the need to perform a count on the database.
```

onadata/apps/api/viewsets/briefcase_viewset.py
```diff
@@ -143,10 +144,16 @@ def filter_queryset(self, queryset):
         if num_entries:
             instances = instances[:num_entries]
 
-        if instances.count():
-            last_instance = instances[instances.count() - 1]
-            self.resumptionCursor = last_instance.pk
-        elif instances.count() == 0 and cursor:
+        # Using len() instead of .count() to prevent an extra
+        # database call; len() will load the instances in memory allowing
+        # us to pre-load the queryset before generating the response
```
DavisRayM, line 149:
```suggestion
        # us to pre-load the queryset before generating the response
```

onadata/apps/api/viewsets/briefcase_viewset.py
```diff
@@ -143,10 +144,16 @@ def filter_queryset(self, queryset):
...
+        # and removes the need to perform a count on the database.
+        instance_count = len(instances)
+
+        if instance_count:
+            last_instance = instances[instance_count - 1]
+            self.resumptionCursor = last_instance.get('pk')
+        elif instance_count == 0 and cursor:
```
DavisRayM, line 156:
```suggestion
        if instance_count > 0:
            last_instance = instances[instance_count - 1]
            self.resumptionCursor = last_instance.get('pk')
        elif instance_count == 0 and cursor:
```

onadata/apps/api/viewsets/briefcase_viewset.py
```diff
@@ -127,8 +127,9 @@ def filter_queryset(self, queryset):
             xform_kwargs['user__username__iexact'] = username
         xform = get_form(xform_kwargs)
         self.check_object_permissions(self.request, xform)
-        instances = Instance.objects.filter(xform=xform,
-                                            deleted_at__isnull=True)
+        instances = Instance.objects.filter(
+            xform=xform, deleted_at__isnull=True).values(
+                'pk', 'uuid')
```
ukanga, line 132: Are we using the `uuid` anywhere? We could just only return the `pk`.

ukanga: Noted, forgot about that.

PR #2144, opened 2021-09-09
Query data endpoint by submission review. Filter `NULL` review status

### LINKED ISSUE #2112: Filtering Pending review status data using API does not return correct data
When I try to filter data that has "_review_status is null using API, it does not result to data that have pending status
using this query `https://api.ona.io/api/v1/data/form_id.json?query={"_review_status" : "3"}`  I get empty json 

Steps to replicate issue
- Turn on [review status](https://help.ona.io/knowledge-base/submission-review/) settings 
- Approve few records 
- Reject few records and leave others as pending
- Try to filter records with pending status

WinnyTroy:
- Test ability to query data whose submission review status is `NULL`

- [x] Included tests
- [x] Updated documentation
- [ ] QA

### Side effects of having these changes
- Submission review status `PENDING` can be thought of in 2 ways:
    - Status `3` - For instances when the user had already provided a status, i.e approved/rejected and would like to return submission review status to pending.
    - Status `NULL` - By default, if no review has been provided for the submission.

- [x] Included tests 

Closes #2112

### DIFF
--- a/onadata/apps/viewer/parsed_instance_tools.py
+++ b/onadata/apps/viewer/parsed_instance_tools.py
@@ -78,7 +78,7 @@ def _parse_where(query, known_integers, known_decimals, or_where, or_params):
             if field_key in NONE_JSON_FIELDS:
                 where.append("{} = %s".format(NONE_JSON_FIELDS[field_key]))
                 where_params.extend([text(field_value)])
-            else:
+            elif field_value:
                 where.append(u"json->>%s = %s")
                 where_params.extend((field_key, text(field_value)))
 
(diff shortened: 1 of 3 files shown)


onadata/apps/viewer/parsed_instance_tools.py
```diff
@@ -120,24 +120,29 @@ def get_where_clause(query, form_integer_fields=None,
...
-            if isinstance(query, dict) and '$or' in list(query):
-                or_dict = query.pop('$or')
-
-                for or_query in or_dict:
-                    for k, v in or_query.items():
+            if isinstance(query, dict):
+                if '$or' not in list(query):
+                    for k, v in query.items():
                         if v is None:
                             or_where.extend([u"json->>'{}' IS NULL".format(k)])
```
DavisRayM, line 127: Seems like this functionality would not properly handle queries without the `$or` query in it... Was there a need to change the functionality from how it was earlier?

DavisRayM: Seems the query is handled in the `_parse_where` function could we extend that function to handle `field_value` when it's `None` instead of looping through the query here perhaps?

WinnyTroy: Sounds good.
Made that change

DavisRayM (approved):

denniswambua (dismissed):

denniswambua (approved):

denniswambua (approved):

Issue #2089, opened 2021-05-27
Can't get a list of chart field endpoints


faith-mutua:
### Environmental Information

- Onadata version: 2.5.5

### Problem description

The endpoint `https://api.ona.io/api/v1/charts/{formid}` fails with the error `400 Bad Request`

### Expected behavior

A user should be able to view a list of chart field endpoints for a specific form or dataset using the endpoint `/api/v1/charts/{formid}`

### Steps to reproduce the behavior

Try getting a list of chart field endpoints using `https://api.ona.io/api/v1/charts/{formid}`

### Additional Information

_Logs, [related issues](github.com/onaio/onadata/issues), weird / out of place occurrences, local settings, possible approach to solving this..._

LeonRomo:
Closed by another PR

PR #2151, opened 2021-09-16
Fix charts groupby multiple fields and check content type 

LeonRomo:
test on charts by grouping multiple fields.
test on charts with content-type as application/JSON
update the format to JSON in the even content type has been provided

tested locally and used the unit test
N/A

- [ x] Included tests 
- [ ] Updated documentation

Closes #
#2089

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_charts_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_charts_viewset.py
@@ -1,6 +1,7 @@
 import json
 import os
 import mock
+
 from django.utils import timezone
 from django.core.cache import cache
 from rest_framework.test import APIClient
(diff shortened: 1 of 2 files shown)


onadata/apps/api/tests/viewsets/test_charts_viewset.py
```diff
@@ -475,3 +475,69 @@ def test_nan_not_json_response(self):
...
+            'field_label': 'How old are you?',
+            'field_xpath': 'age',
+            'field_name': 'age',
+            'field_type': 'integer',
+            'grouped_by': ['gender', 'pizza_fan'],
+            'xform': self.xform.pk
+            }
+
+        renderer = DecimalJSONRenderer()
+        res = json.loads(renderer.render(response.data).decode('utf-8'))
```
DavisRayM, line 525: I think we should use the default content negotiation to confirm that the viewset will use the correct renderer and return the response we expect.

```suggestion
        response.render()
        res = json.loads(response.content.decode('utf-8'))
```

LeonRomo: Okay, that should be alright.

denniswambua (approved):

PR #2152, opened 2021-09-22
Refactor enketo authentication with onadata while retrieving X-OpenRosa-Accept-Content-Length header

WinnyTroy:
Allow requests where request method is `HEAD` to not require authentication.
These requests don't return any data, but provide necessary response headers to build the enketo forms for submission collection

- Included tests

Closes onaio/zebra#6472

### DIFF
--- a/onadata/apps/api/permissions.py
+++ b/onadata/apps/api/permissions.py
@@ -522,7 +522,10 @@ class IsAuthenticatedSubmission(BasePermission):
                 return False
             profile, _ = UserProfile.objects.get_or_create(user=user)
 
-            if profile.require_auth:
+            # For the two request methods being handled at this point
+            # Only raise permission denied exception for POST requests where
+            # request user is anonymous && profile.require_auth is enabled
+            if profile.require_auth and request.method != 'HEAD':
                 # raises a permission denied exception,
                 # forces authentication
                 return False
--- a/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_submission_viewset.py
@@ -808,6 +808,49 @@ class TestXFormSubmissionViewSet(TestAbstractViewSet, TransactionTestCase):
                     Instance.objects.filter(xform=self.xform).count(),
                     count+1)
 
+    def test_head_submission_request_w_no_auth(self):
+        """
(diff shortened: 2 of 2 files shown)


onadata/libs/utils/logger_tools.py
```diff
@@ -797,6 +798,19 @@ class OpenRosaResponseForbidden(OpenRosaResponse):
...
+    def __init__(self, *args, **kwargs):
+        super(OpenRosaNotAuthenticated, self).__init__(*args, **kwargs)
+
+        self['Content-Type'] = 'text/html; charset=utf-8'
+        self['X-OpenRosa-Accept-Content-Length'] = DEFAULT_CONTENT_LENGTH
+        tz = pytz.timezone(settings.TIME_ZONE)
+        dt = datetime.now(tz).strftime('%a, %d %b %Y %H:%M:%S %Z')
+        self['Date'] = dt
+
+
```
ukanga, line 813: What challenges were there when subclassing `BaseOpenRosaResponse` instead which already as 807 to 811 implemented?

WinnyTroy: The `BaseOpenRosaResponse`  which handles all the other OpenRosa exceptions inherits from django's HttpResponse class.
The response object for requests failing with 401 status code should be an instance of django's default response class.

Another striking difference is the content_type.
The other exceptions would have the response content type set to `text/xml; charset=utf-8` the NotAuthenticated OpenRosa exception requires the content type to be`text/html; charset=utf-8`

DavisRayM: @WinnyTroy Why does it need to be `text/html`? If it needs to be HTML the name might be a bit confusing since the OpenRosa standard mostly responds in `XML`.

Also, why does it need to be a Django response in particular ?

WinnyTroy: Before implementing this class to include the headers, the response object being returned was an instance of django's default response class. The content type for this response object was `text/html`. Keeping in line with existing implementation flow, I didnt want to change that to xml

Also, trying to implement this class using django's HttpResponse class breaks existing tests with `AttributeError: 'OpenRosaNotAuhenticated' object has no attribute 'request'`. So that is why i subclassed this from django's default response class.

DavisRayM (commented):
LGTM; Kindly rebase

PR #2166, opened 2021-11-11
Csv import overwrite enabled

denniswambua:
Fix csv import overwrite which only updated the soft deleted submissions.

Tests included

Uuids will always be generated for all the csv import modes.

- [x] Included tests 
- [x] Updated documentation

Closes #


DavisRayM (commented):
LGTM, Kindly also update the documentation here https://stage-api.ona.io/static/docs/forms.html#import-csv-data-to-existing-form; To highlight that during an overwrite we don't consider the `_uuid` of the imported data.

docs/forms.rst
```diff
@@ -1210,7 +1210,9 @@ Response
 Import CSV data to existing form
 ---------------------------------
 
-- `csv_file` a valid csv file with exported data (instance/submission per row)
+- `csv_file` a valid csv file with exported data (instance/submission per row).
+  _Note:_ uuid column is ignored and is always regenerated.
```
DavisRayM, line 1214:
```suggestion
 
 Use the `overwrite` parameter to clear all previous submissions while importing submissions.
 Note: Overwrites are not reversible & the `uuid` column is ignored during the process and is always regenerated.
```

DavisRayM (approved):

DavisRayM (approved):

PR #2167, opened 2021-11-16
Use auth user model for _submitted_by field

### LINKED ISSUE #2163: High request times for retrieval of statistics & data on forms with a large number of contributors and submissions
## Current situation

The statistics endpoint `/api/v1/stats/submissions` alongside the data endpoint `/api/v1/data` (when used with start and limit query params) and media count endpoints currently take a lot of time(Around 1 - 3 minute depending on server load) to compute the statistics of a given form... This leads to some unpleasant user interactions.

We should look into a way to decrease the amount of processing/query time we take for such requests in order to speed up the request times as the current wait time is too long IMO

image

## Suggested approaches

- Calculate submission stats as submissions are made(Async task post submission perhaps?) & save the statistics to cache

denniswambua:
Queries for a form collaborators with a lot of submissions  are taking  like 8 mins. We can improve this if we can fetch the data from the user model. 

Testing this and the result are as below
```
Querying using the json field took 528732.595 ms 
Querying with left join took 10016.579 ms
```

Tests added and tested on

- [x] Included tests 
- [ ] Updated documentation

Closes #2163

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_stats_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_stats_viewset.py
@@ -57,6 +57,16 @@ class TestStatsViewSet(TestBase):
         }
         self.assertDictContainsSubset(data, response.data[0])
 
+        request = self.factory.get('/?group=_submitted_by', **self.extra)
+        response = view(request, pk=formid)
+        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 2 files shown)


DavisRayM (dismissed):
LGTM

DavisRayM (approved):

PR #2170, opened 2021-11-24
Default response format to JSON

DavisRayM:

- Default the response format on the charts endpoint to `JSON` if not format is passed

- Updated tests

The charts endpoint will always return a JSON response if the `format` query param is empty

- [x] Included tests 
- [ ] Updated documentation

Related to #2151 

_Main reason for the update is because the browser would still return a `Not supported` error with the changes introduced in #2151; Which wasn't the ideal scenario for an API.... If the format is not specified we should always default to JSON IMO_

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_charts_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_charts_viewset.py
@@ -160,16 +160,6 @@ class TestChartsViewSet(TestBase):
         self.assertEqual(response.data['data'][0]['gender'], 'Male')
         self.assertEqual(response.data['data'][1]['gender'], 'Female')
 
-    def test_return_bad_request_on_non_json_request_with_field_name(self):
-        request = self.factory.get('/charts/%s.html' % self.xform.id)
-        force_authenticate(request, user=self.user)
-        response = self.view(
-            request,
-            pk=self.xform.id
-        )
-        self.assertEqual(response.status_code, 400)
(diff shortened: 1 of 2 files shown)


denniswambua (approved):

onadata/apps/api/tests/viewsets/test_charts_viewset.py
```diff
@@ -160,6 +161,7 @@ def test_get_on_categorized_field(self):
         self.assertEqual(response.data['data'][0]['gender'], 'Male')
         self.assertEqual(response.data['data'][1]['gender'], 'Female')
 
+    @skip("https://github.com/onaio/onadata/pull/2170")
```
denniswambua, line 164: @DavisRayM Not sure why we are skipping this

DavisRayM: Skipping it because we default to JSON instead of returning a bad request at the moment

denniswambua: Can we just remove or edit that test?

DavisRayM: I can remove it, it's a bit similar to one of the other tests so I wasn't so sure about editing it. Let me remove it and re-request review

denniswambua: cool

denniswambua (approved):

PR #2176, opened 2022-01-04
Use cache submission stats

denniswambua:
Cache submission stats and chart results for a configurable time (1 hour default).
This will reduce database hits and will improve performance especially during high peek.

Tested in stage

The stats like contributor counts results will be delayed by one hour.


onadata/settings/common.py
```diff
@@ -603,3 +603,8 @@ def configure_logging(logger, **kwargs):
 
 # Project & XForm Visibility Settings
 ALLOW_PUBLIC_DATASETS = True
+
+# Cache xform submission stat by one hour 60 * 60
+XFORM_SUBMISSION_STAT_CACHE_TIME = 3600
+
+XFORM_CHARTS_CACHE_TIME = 3600
```
DavisRayM, line 610: I'm a bit 50/50 on whether these timings should be the default for the application at all times... 1 hour seems to be a bit too much for regular chart calculations(For deployments without a lot of data); IMO by default we should cache for a 1 minute maximum

DavisRayM (dismissed):
LGTM, have one comment

DavisRayM (approved):

PR #2184, opened 2022-01-21
2160 fix users without user profiles

### LINKED ISSUE #2160: Organization Viewset fails when a member of an Organization does not have a user profile
### Environmental Information

- Onadata version: master

### Problem description

The Organization viewset fails to retrieve organizations when a member of the organization does not have a User Profile; Most users created from the legacy UI do not have UserProfile created for them... since the User Profile model was introduced with the API.

### Expected behavior

- We should check whether a users' profile exists before accessing `gravatar` [here](https://github.com/onaio/onadata/blob/master/onadata/libs/serializers/organization_serializer.py#L123)

- Create one user from the API endpoints (`bob`)
- Create another user using the legacy UI (`joe`)
- Using the user created by the API:
    - Create an Organization
    - Add the user `joe` to the organization
    - Try retrieving the organization using `bob` user
   
_NB: During the entire procedure, do not login as `joe` by using the API endpoints; This creates a `UserProfile`_

### Additional Information

_Logs, related issues, weird / out of place occurrences, local settings, possible approach to solving this..._

- Related to [sentry issue](https://sentry.onalabs.org/ona/onadata/issues/90944/?query=is%3Aunresolved)

KipSigei:
- Added management command to add user profiles to user/org accounts with missing profiles. Noticed that this might happen if a user is created on python shell.

- Added tests to ensure a user without a user profile has one assigned to them on running the management command.

- None

- [x] Included tests 
- [ ] Updated documentation

Closes #2160

### DIFF
--- a/onadata/apps/api/management/commands/create_user_profiles.py
+++ b/onadata/apps/api/management/commands/create_user_profiles.py
@@ -1,8 +1,5 @@
 # -*- coding: utf-8 -*-
-"""
-Management Command to add missing user profiles to users
-"""
-
+"""Management Command to add missing user profiles to users."""
 from django.core.management.base import BaseCommand
 from django.contrib.auth.models import User
 from django.utils.translation import gettext as _
@@ -12,19 +9,17 @@ from onadata.libs.utils.model_tools import queryset_iterator
 
 
 class Command(BaseCommand):
+    """Create missing user profiles management command."""
     help = _("Build out missing user profiles")
 
     def handle(self, *args, **options):
         # get all users
-        try:
-            users = User.objects.all()
-            for user in queryset_iterator(users):
-                try:
-                    profile = user.profile
-                except UserProfile.DoesNotExist:
-                    profile, _ = UserProfile.objects.get_or_create(user=user)
-                    profile.save()
-        except User.DoesNotExist:
-            pass
+        users = User.objects.all()
+        for user in queryset_iterator(users):
+            try:
+                profile = user.profile
+            except UserProfile.DoesNotExist:
+                profile, _ = UserProfile.objects.get_or_create(user=user)
+                profile.save()
 
         self.stdout.write("User Profiles successfully created.")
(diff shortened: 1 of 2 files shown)


onadata/apps/api/management/commands/create_user_profiles.py
```diff
@@ -0,0 +1,31 @@
...
+        # get all users
+        try:
+            users = User.objects.all()
+            for user in queryset_iterator(users):
+                try:
+                    profile = user.profile
+                except UserProfile.DoesNotExist:
+                    profile, _ = UserProfile.objects.get_or_create(user=user)
+                    profile.save()
+        except User.DoesNotExist:
```
DavisRayM, line 26: Is there a need to capture the `DoesNotExist` error? Won't the queryset only return User objects that actually exist?

onadata/apps/api/tests/management/commands/test_create_user_profiles.py
```diff
@@ -0,0 +1,37 @@
...
+        out = StringIO()
+        call_command(
+            'create_user_profiles',
+            stdout=out
+        )
+        user.refresh_from_db()
+        try:
+            _ = user.profile
+        except UserProfile.DoesNotExist:
+            assert False
```
DavisRayM, line 32: Instead of capturing the DoesNotExist exception we can let it be raised so that the test fails with the `UserProfile.DoesNotExist` exception

```suggestion
        # Assert profile is retrievable;
        profile = user.profile
        self.assertEqual(profile.user, user)
```

DavisRayM (approved):

KipSigei:
@DavisRayM @pld I think we should disable pep257's `D203` error `1 blank line required before class docstring` since the rest of the code doesn't have a blank space before class docstrings

DavisRayM:
> @DavisRayM @pld I think we should disable pydocstyle's `D203` error `1 blank line required before class docstring` since the rest of the code doesn't have a blank space before class docstrings

Updated the Codacy checks

DavisRayM (approved):

PR #2201, opened 2022-02-23
Add `refresh` query param to invalidate charts cache

DavisRayM:
## Changes

Adds a `refresh` query param that can be used to invalidate the Charts
cache and recalculate the updated cache

Related to: https://github.com/onaio/zebra/issues/6628

### DIFF
--- a/docs/charts.rst
+++ b/docs/charts.rst
@@ -79,6 +79,19 @@ Response
  - ``html`` format response is a html, javascript and css to the chart
  - ``json`` format response is the ``JSON`` data that can be passed to a charting library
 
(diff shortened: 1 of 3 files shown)


docs/charts.rst
```diff
@@ -79,6 +79,19 @@ Response
  - ``html`` format response is a html, javascript and css to the chart
  - ``json`` format response is the ``JSON`` data that can be passed to a charting library
 
+A few things to note about the retrieving a chart for a specific field is that the resulting chart is cached for 10 minutes. The cache can be invalidated by passing in an optional ``refresh`` query parameter.
```
DavisRayM, line 82:
```suggestion
Note: When retrieving a chart for a specific field is that the resulting response is cached for ten minutes. The cache can be invalidated by passing in an optional ``refresh`` query parameter.
```

KipSigei (approved):

PR #2202, opened 2022-02-25
Ensure returned pagination headers are correct when querying for specific values

### LINKED ISSUE #2203: When querying data with pagination, the returned pagination link headers are invalid
### Environmental Information

- Onadata version: master

### Problem description

When trying to query(_using `query` param_) data from the data endpoint the returned pagination headers do not reflect the available pages after the filter. The application ends up returning headers in relation to all the data contained by the form.

### Expected behavior

- When using the `query` param the API should only return pagination links that actually contain data

- Upload form with a text question i.e What's your name ?
- Submit 4 records for `Bob`, `Bob`, `Bob` & `Troy`
- Try and query records for bob `/api/v1/data/<xform_id>?query="Bob"&page_size=1`
- Returned headers should only indicate that the API can return 3 pages; not 4 

### Additional Information

DavisRayM:

- Test that the pagination headers returned while querying are accurate
- Count returned queryset if the `query` param is present

- [x] Updated tests

N/A

- [x] Included tests 

Closes #2203

### DIFF
--- a/onadata/apps/api/viewsets/data_viewset.py
+++ b/onadata/apps/api/viewsets/data_viewset.py
@@ -417,8 +417,12 @@ class DataViewSet(AnonymousUserPublicFormsMixin,
         import math
 
         url = self.request.build_absolute_uri()
+        query = self.request.query_params.get('query')
         base_url = url.split('?')[0]
-        num_of_records = xform.num_of_submissions
+        if query:
+            num_of_records = self.object_list.count()
+        else:
+            num_of_records = xform.num_of_submissions
         next_page_url = None
         prev_page_url = None
         first_page_url = None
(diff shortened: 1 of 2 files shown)


onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -417,8 +417,12 @@ def _set_pagination_headers(
         import math
 
         url = self.request.build_absolute_uri()
+        query = self.request.query_params.get('query')
         base_url = url.split('?')[0]
-        num_of_records = xform.num_of_submissions
+        if query:
+            num_of_records = self.object_list.count()
```
ukanga, line 423: We might need to keep an eye on the performance implication of this count.

ukanga (approved):

PR #2205, opened 2022-03-01
Support split select multiples with randomize=true parameters

### LINKED ISSUE #2185: Fix export options check for XLS forms with split_multiple enabled
### Environmental Information
- Onadata version:
master, prodution

### Problem description
With the split_multiple option enabled on a form it doesn't not retrieve the correct export.

### Expected behavior
With the split_multiple option enabled on a form it should retrieve the correct export.

- Generate XLS export without the split_multiple_enabled option
- Generate another one with the split_multiple option enabled
- You will not get the correct export on the one that has the option enabled

### Additional Information

Related issues https://github.com/onaio/zebra/issues/6673

ukanga:
- Fix split select bug during export when randomize parameter is set to true

- Added a check to correctly build choices for split select multiple fields with randomize param set to true

- [x] Included tests 
- [ ] Updated documentation

Closes #2185


onadata/libs/tests/utils/test_export_builder.py
```diff
@@ -2404,6 +2404,70 @@ def test_select_multiples_choices_with_choice_filter(self):
...
+        self.assertEqual(choices, expected_choices)
+        select_multiples = {
+            'county': [
+                ('county/king', 'king', 'King'),
+                ('county/pierce', 'pierce', 'Pierce'),
+                ('county/king', 'king', 'King'),
+                ('county/cameron', 'cameron', 'Cameron')
+            ]
+        }
+        self.assertEqual(CSVDataFrameBuilder._collect_select_multiples(dd),
```
DavisRayM, line 2468: The test name is a bit confusing; It seems this was supposed to test splitting the select multiples but it tests the `_collect_select_multiples` function instead

DavisRayM: Could we perhaps do with only one test? 

https://github.com/onaio/onadata/pull/2205/files#diff-71defac856691f0cf9680a9c332085b151b84e2222508c82338a6fde859bb98eR1486-R1563 seems to be almost the same as this one with some minor differences

onadata/libs/utils/csv_builder.py
```diff
@@ -229,7 +230,10 @@ def _collect_select_multiples(cls, dd, language=None):
             choices = [(c.get_abbreviated_xpath(), c.name,
                         get_choice_label(c.label, dd, language))
                        for c in e.children]
-            if not choices and e.choice_filter and e.itemset:
+            is_choice_randomized = str_to_bool(
+                e.parameters and e.parameters.get('randomize'))
+            if not choices and e.choice_filter and e.itemset \
+                    or is_choice_randomized:
```
ukanga, line 236: Should this be 
```python
 if ((not choices and e.choice_filter) or is_choice_randomized ) and e.itemset:
   # ....
```
e.itemset should always be True in both cases, I think.

onadata/libs/tests/utils/test_export_builder.py
KipSigei: addressed

DavisRayM (commented):
Is there a need for the additional fixture? Can we get rid of it since it's unused ?

DavisRayM (approved):

onadata/libs/utils/export_builder.py
```diff
@@ -399,8 +400,10 @@ def get_choice_dict(xpath, label):
             }
 
         choices = []
-
-        if not child.children and child.choice_filter and child.itemset:
+        is_choice_randomized = str_to_bool(
+            child.parameters and child.parameters.get('randomize'))
+        if ((not child.children and child.choice_filter)
+                or is_choice_randomized) and child.itemset:
```
ukanga, line 406: I think we are missing a test for this change, we could test `_get_select_mulitples_choices` in the `test_export_builder`.

ukanga (commented):
Looks good

PR #2219, opened 2022-04-06
Return an `AuthenticationFailed` exception instead of a 404 when Enketo token is not retrievable

### LINKED ISSUE #2218: Page not found exception when using two versions of OnaData on the same subdomain
### Environmental Information

- Onadata version: master

### Problem description

A 404 status code page is currently returned whenever a user signs in on two different versions of OnaData that have been deployed to the same cookie domain.

Screenshot from 2022-04-06 15-07-34

### Expected behavior

- A 404 should not be raised; At best the cookies should be cleared or ignored

- Login to ona.io
- Try to login to preview.ona.io

### Additional Information

N/A

DavisRayM:

- Update the `get_api_token` function to raise an `AuthenticationFailed` expection instead of a 404

- [ ] QA

- A 404 status code will no longer be raised if the provided cookie does not point to an existing token object

- [x] Included tests 
- [ ] Updated documentation

Closes #2218

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,5 +76,8 @@ bob
 .vscode/
 tags
 
+# vim
+.vim
+
 .bash_history
 .inputrc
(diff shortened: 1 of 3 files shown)


KipSigei (commented):
could we add a test case for the raised exception?

KipSigei (approved):

Issue #2210, opened 2022-03-11
Upgrade to Django 3.x


KipSigei:
There is need to upgrade to Django 3.x which contains several bug fixes, features, security and overall performance enhancements that we should take advantage of.

imkingrobi:
- [ ] 6554

PR #2230, opened 2022-04-19
Upgrade dependencies for Django 3.x upgrade

ukanga:
Also converted to using pyproject.toml

- SPSS exports will be disabled.

- [ ] Included tests 
- [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2210


setup.cfg
```diff
@@ -0,0 +1,116 @@
+[metadata]
+name = onadata
+description = Collect Analyze and Share Data
```
KipSigei, line 3: app version seems to be missing

setup.cfg
```diff
@@ -0,0 +1,116 @@
...
+    dpath
+    elaphe3
+    httplib2
+    modilabs-python-utils
+    numpy
+    Pillow
+    python-dateutil
+    pytz
+    requests
+    requests-mock
```
KipSigei, line 91: should be moved to `tests_require` section

onadata/__init__.py
```diff
@@ -6,7 +6,7 @@
 """
 from __future__ import absolute_import, unicode_literals
```
KipSigei, line 8: we might not need the `__future__` import

KipSigei (approved):

PR #2237, opened 2022-05-17
Allow geojson file uploads

FrankApiyo:
Use [add_mimetypes](https://docs.python.org/3/library/mimetypes.html#mimetypes.add_type) to add `geojson` type
Added a test case
N/A

- [x] Included tests 
- [ ] Updated documentation

### DIFF
--- a/onadata/libs/serializers/metadata_serializer.py
+++ b/onadata/libs/serializers/metadata_serializer.py
@@ -173,6 +173,8 @@ class MetaDataSerializer(serializers.HyperlinkedModelSerializer):
 
         if data_file:
             allowed_types = settings.SUPPORTED_MEDIA_UPLOAD_TYPES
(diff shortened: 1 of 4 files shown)


ukanga (approved):

onadata/libs/serializers/metadata_serializer.py
```diff
@@ -173,6 +173,8 @@ def validate(self, attrs):
 
         if data_file:
             allowed_types = settings.SUPPORTED_MEDIA_UPLOAD_TYPES
+            # add geojson mimetype
+            mimetypes.add_type('application/geo+json', '.geojson')
```
DavisRayM, line 177: Why do we need to do this ? I've removed this line and the tests still pass. Isn't the change in the settings enough ?

PR #2291, opened 2022-07-13
Fix an issue when trying to access azure attachments with the `suffix` query param

DavisRayM:

- Move `generate_attachments_download_url` to `libs/image_tools.py`
- Add `generate_media_url_with_sas` function
- Update `generate_attachments_download_url`

- [ ] QA

N/A

Closes #


onadata/libs/utils/image_tools.py
```diff
@@ -1,22 +1,96 @@
...
+                    "Key": file_path,
+                    "ResponseContentDisposition": content_disposition,
+                    "ResponseContentType": "application/octet-stream",
+                },
+                ExpiresIn=expiration,
+            )
+        except ClientError as e:
+            logging.error(e)
+            return None
+
```
ukanga, line 65: Could you also split this out to its own function?

PR #2293, opened 2022-07-18
Support `no-store` cache control value

DavisRayM:

- Add `CacheControlMiddleware` class
- Update `CacheControlMixin` class
- Add `set_cache_control` utility function
- Remove `CACHE_MIXIN_SECONDS` setting
- Add `CACHE_CONTROL_DIRECTIVES` setting

- [x] Testing on stage

N/A

  - [ ] Included tests
  - [ ] Updated documentation

### DIFF
--- a/onadata/settings/common.py
+++ b/onadata/settings/common.py
@@ -559,8 +559,6 @@ STATIC_DOC = "/static/docs/index.html"
 
 HOSTNAME = socket.gethostname()
 
-CACHE_MIXIN_SECONDS = 60
-
 TAGGIT_CASE_INSENSITIVE = True
 
 DEFAULT_CELERY_MAX_RETIRES = 3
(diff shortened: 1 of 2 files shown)


onadata/settings/common.py
```diff
@@ -557,8 +557,6 @@ def configure_logging(logger, **kwargs):
 
 HOSTNAME = socket.gethostname()
 
-CACHE_MIXIN_SECONDS = 60
-
```
ukanga, line 561: Could we have a default `CACHE_CONTROL_DIRECTIVES` setting since we are replacing the `CACHE_MIXIN_SECONDS`?
```python
CACHE_CONTROL_DIRECTIVES = {"max_age": 60}
```

ukanga (dismissed):

onadata/settings/common.py
ukanga: We maintain the current behaviour but switch to a new way of defining it.

ukanga (approved):

KipSigei (approved):

Issue #2204, opened 2022-02-25
Add a mechanism by which we can know when a csv import task is being carried on on a form


FrankApiyo:
## Suggested Feature / Enhancement
- currently, it's possible to carry out multiple CSV imports on a form at the same time; this might arguably cause unpredictable behavior
- in addition, it's not easy to determine if a csv import is currently being carried out on a form

PR #2299, opened 2022-08-18
[WIP]Add endpoint for fetching currently running csv imports

FrankApiyo:

  - [ ] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2204


DavisRayM (dismissed):

KipSigei (dismissed):

onadata/apps/api/viewsets/v2/imports_viewset.py
```diff
@@ -0,0 +1,214 @@
...
+        else:
+            if xls_file and xls_file.name.split(".")[-1] in XLS_EXTENSIONS:
+                csv_file = submission_xls_to_csv(xls_file)
+
+            overwrite = request.query_params.get("overwrite")
+            overwrite = (overwrite.lower() == "true" if isinstance(
+                overwrite, str) else overwrite)
+
+            # Block imports from running when an overwrite is ongoing
+            active_tasks = self._get_active_tasks(xform)
```
ukanga, line 133: Why not `get_active_tasks(self.task_names, xform)` it is still clean and there is nothing special about the `_get_active_tasks()` that requires it to be a class method.

onadata/apps/api/urls/v2_urls.py
```diff
@@ -2,9 +2,70 @@
...
+        """
+        self.registry.extend(router.registry)
+
+
+base_router = MultiLookupRouter(trailing_slash=False)
+router = DetailedPostRouter(trailing_slash=False)
+base_router.register(r"open-data", TableauViewSet, basename="open-data")
+router.register(r"imports", ImportsViewSet, basename="imports")
+
+router.extend(base_router)
```
ukanga, line 71: Why is the DetailedPostRouter necessary? Which path is impossible to do with the existing default and custom routers that we need another custom router?

DavisRayM: This was mostly implemented so that we could pass a detail to a POST route; AFAIK the post (create) route on DRF doesn't usually support detail... Is there a particular way we could implement this without having to create our own custom route?

tl;dr This was done so that we could pass detail to the default `create` function; Allowing requests such as `curl -X POST "http://<domain>/api/v2/imports/<xform_id>"`

onadata/apps/api/viewsets/v2/imports_viewset.py
DavisRayM: Will make this change

onadata/apps/api/urls/v2_urls.py
ukanga: We should use create/post on the `/api/v2/imports` to create a new import instead of `/api/v2/imports/<xform_id>`. The `xform_id` should be a mandatory post parameter in the post to create an import. With that approach, there is no need for a custom router. We already take this approach with the `/api/v1/forms`, `/api/v1/dataviews` and other endpoints.

PR #2319, opened 2022-09-27
Add pagination to forms endpoint

KipSigei:
- Enable pagination when listing forms
- Tests

  - [x] Included tests
  - [x] Updated documentation

### DIFF
--- a/docs/forms.rst
+++ b/docs/forms.rst
@@ -173,6 +173,23 @@ Request
 
     curl -X GET https://api.ona.io/api/v1/forms?owner=ona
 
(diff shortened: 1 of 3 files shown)


DavisRayM:
Should we paginate the response by default ? Perhaps only allow 1000 records or something of the sort ? I don't see the need to always return past 1000 records

DavisRayM (approved):

KipSigei:
@DavisRayM The response will always be paginated if there is a form, the condition added is to handle cases where there are no forms and page will end up being None/empty queryset

PR #2320, opened 2022-09-29
Add pagination to projects endpoint

KipSigei:
- Add pagination to projects endpoint
- Tests

  - [x] Included tests
  - [x] Updated documentation

Closes #


DavisRayM (approved):
One thing, Can we add a note, that the default page size for XForms and Projects is 100,000 records ?

Other than that LGTM

DavisRayM (approved):

PR #2335, opened 2022-11-16
Update GeoJSON endpoint to filter by instances with geoms

KipSigei:
- Update geojson enpoint to return 404 if all instances/submissions have empty geoms e.g
## before
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "GeometryCollection",
        "geometries": []
      },
      "properties": {
        "id": 111,
        "xform": 20
      }
    }
  ]
}
```

## now
```
{
  "detail": "Not found."
}
```
- Tests

  - [x] Included tests
  - [ ] Updated documentation

Closes #


DavisRayM (approved):

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -393,6 +393,17 @@ def destroy(self, request, *args, **kwargs):
...
                 delete_instance(self.object, request.user)
+
+                # update xform if no instance has geoms
+                if (
+                    self.object.xform.instances.filter(
+                        deleted_at__isnull=True, geom=None
+                    ).count()
+                    < 1
+                ):
+                    self.object.xform.instances_with_geopoints = False
```
DavisRayM, line 404: Can we have this implemented on the model level instead of the viewset ? _Guessing this field is updated/set on the model level ?_

onadata/apps/logger/models/instance.py
```diff
@@ -794,11 +794,21 @@ def soft_delete_attachments(self, user=None):
...
+        # update xform if no instance has geoms
+        if (
+            instance.xform.instances.filter(
+                deleted_at__isnull=True, geom=None
+            ).count()
+            < 1
+        ):
+            instance.xform.instances_with_geopoints = False
+            instance.xform.save()
+
```
DavisRayM, line 811: Let's have this happen in `update_xform_submission_count` and let's try to avoid a count on the XForm. If possible let's use the Forms `submission_count` property

onadata/apps/api/viewsets/data_viewset.py
KipSigei: @DavisRayM Moved the logic to the instance model post save signal instead to handle the update after a submission is soft deleted. Note that with soft deletion the post_save signal is triggered instead of post_delete

onadata/apps/logger/models/instance.py
```diff
@@ -316,9 +316,22 @@ def update_xform_submission_count_delete(sender, instance, **kwargs):
         safe_delete(f"{DATAVIEW_COUNT}{xform.pk}")
         safe_delete(f"{XFORM_COUNT}{xform.pk}")
 
-        if xform.instances.exclude(geom=None).count() < 1:
-            xform.instances_with_geopoints = False
-            xform.save()
+        # update xform if no instance has geoms
+        if (
+            instance.xform.instances.filter(
+                deleted_at__isnull=True, geom=None
```
DavisRayM, line 322: This query won't do what is expected; It's counts instances that are not deleted with no geoms... Let's use the initial query `xform.instances.exclude(geom=None, deleted_at__isnull=False).count() < 1`

DavisRayM (changes requested):

onadata/apps/logger/models/instance.py
DavisRayM: `xform.instances.filter(~Q(geom=None), deleted_at__isnull=True).count() < 1`

DavisRayM (approved):

PR #2336, opened 2022-11-16
Propagate project permissions to all KPI Assets

### LINKED ISSUE #1060: Form builder forms can not be edited by other admins in org account
All

### Expected behavior
In an org account, any admin should be able to edit a form previously developed by another individual in form builder as this is an option listed for all admins (see screen shot below). 

screen_shot_2017-06-28_at_8_48_10_am

### Actual behavior
If another admin, who did not develop the form in form builder, clicks edit the form, they receive and error message (see screen shot below) and is unable to edit the form.

screen shot 2017-06-27 at 9 17 16 am

The work around would be to use XLSForm syntax form and replace the form. Everyone would have access to that.

- Have 1 admin in an org account create a form in form builder and publish the form
- Have another org admin try to edit the form

Sample Form: https://ona.io/acme/916/216297
### Implementation plan

- 
- 
-

DavisRayM:

- [x] Ensure permissions are created for Owners on KPI published assets
- [x] Ensure permissions are removed from KPI published assets once a users' role is downgraded/removed
- [x] Cleanup

- [x] QA

_TODO_

  - [x] Included tests
  - [ ] Updated documentation

Closes #1060

### DIFF
--- a/onadata/libs/utils/project_utils.py
+++ b/onadata/libs/utils/project_utils.py
@@ -21,9 +21,13 @@ from onadata.apps.api.models.team import Team
 from onadata.apps.logger.models.project import Project
 from onadata.apps.logger.models.xform import XForm
 from onadata.celeryapp import app
-from onadata.libs.permissions import (ROLES, OwnerRole,
-                                      get_object_users_with_permissions,
-                                      get_role, is_organization)
+from onadata.libs.permissions import (
+    ROLES,
+    OwnerRole,
+    get_object_users_with_permissions,
+    get_role,
+    is_organization,
+)
 from onadata.libs.utils.common_tags import API_TOKEN, OWNER_TEAM_NAME
 from onadata.libs.utils.common_tools import report_exception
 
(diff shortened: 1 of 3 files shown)


ukanga (approved):

onadata/libs/utils/project_utils.py
```diff
@@ -138,3 +153,177 @@ def _set_project_perms():
...
+        form_ids: List[int] = list(
+            project.xform_set.filter(deleted_at__isnull=True).values_list(
+                "id", flat=True
+            )
+        )
+
+        # Propagate permissions for XForms that were published by
+        # Formbuilder
+        for form_id in form_ids:
+            asset: XForm = XForm.objects.get(id=form_id)
```
KipSigei, line 284: We should add a try catch block here so that if the xform doesnt exist in replica it can be fetched from master

DavisRayM: I think that might not be something we'd need to handle since [this](https://github.com/onaio/onadata/pull/2336/files/e246be0987dd007094b7e3b606b11591d912d7f6#diff-b788ff4632a7e33981fda5c4e9b7dfe6f2b550fcf2d60b65126fe4d808cb0030R275) line should only return XForm objects that are available in the moment

PR #2346, opened 2022-12-21
Ensure excel exports are in xlsx format

KipSigei:
- Ensure excel exports are in `xlsx` format instead of `xls`
- Tests
- All excel exports will be in the `xlsx` format

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_xform_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_viewset.py
@@ -61,6 +61,7 @@ from onadata.apps.logger.xform_instance_parser import XLSFormError
 from onadata.apps.main.models import MetaData
 from onadata.apps.messaging.constants import FORM_UPDATED, XFORM
 from onadata.apps.viewer.models import Export
+from onadata.apps.viewer.models.export import ExportTypeError
 from onadata.libs.permissions import (
     ROLES_ORDERED,
     DataEntryMinorRole,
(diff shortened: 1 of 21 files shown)


onadata/apps/api/tests/viewsets/test_xform_viewset.py
```diff
@@ -3539,6 +3540,32 @@ def test_export_form_data_async(self, async_result):
                 export = Export.objects.get(task_id=task_id)
                 self.assertTrue(export.is_successful)
 
+    @override_settings(CELERY_TASK_ALWAYS_EAGER=True)
+    def test_raises_exception_for_xls_export_formats(self):
```
DavisRayM, line 3544: Test name is a bit misleading; I don't see an exception raised in this test

onadata/apps/viewer/models/export.py
```diff
@@ -68,6 +68,7 @@ class Export(models.Model):
     """
 
     XLS_EXPORT = "xls"
```
DavisRayM, line 70: Should we remove `xls` exports ?

onadata/apps/viewer/tasks.py
```diff
@@ -127,12 +127,10 @@ def _create_export(xform, export_type, options):
 @app.task(track_started=True)
 def create_xls_export(username, id_string, export_id, **options):
```
DavisRayM, line 128: Should we rename the function ? Or better yet add a new function and deprecate this one ?

onadata/apps/api/tests/viewsets/test_xform_viewset.py
KipSigei: The tests asserts the `ExportTypeError` is raised with .xls formats [here](https://github.com/onaio/onadata/pull/2346/files#diff-a7dc39d301394b01b2cfcd02e9dcfc7b3eadca03097090cd2b3a5bf72fa1cdf6R3554)

DavisRayM (approved):
LGTM, let's test it out on stage

DavisRayM (dismissed):
Seems like a breaking change to me; We'll need to communicate accordingly

DavisRayM (approved):

PR #2349, opened 2023-01-10
Ensure external select to csv conversion works as expected

DavisRayM:

- Update `sheet_to_csv` function; Ensure that cells are only processed if they have a value
- Ensure all cells are checked and converted to data/integer when applicable

```
This PR sorts an issue where an empty column is falsely diagnosed as an integer field which caused the XLSX to be processed for integer & datetime fields.

Note: The integer and datetime processing functionality was broken since it treated the sheet as a 0-based index array of which it's not. https://openpyxl.readthedocs.io/en/latest/api/openpyxl.worksheet.worksheet.html#openpyxl.worksheet.worksheet.Worksheet.iter_rows
```

- Updated Tests
- [x] QA

N/A

  - [x] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/zebra/issues/7312


DavisRayM:
Closes https://github.com/onaio/zebra/issues/7312

onadata/apps/main/tests/test_process.py
```diff
@@ -106,6 +107,12 @@ def test_publish_xlsx_file_with_external_choices(self):
         """Test publishing an XLSX file with external choices"""
         self._publish_xlsx_file_with_external_choices()
 
+    def test_publis_xlsx_file_with_external_choices_with_empty_row(self):
```
DavisRayM, line 110:
```suggestion
    def test_public_xlsx_file_with_external_choices_with_empty_row(self):
```

ukanga (approved):

Issue #2359, opened 2023-01-30
Add endpoints to render MergedXForms and Filtered datasets as geoJSON


FrankApiyo:
## Suggested Feature / Enhancement
There is currently no endpoint to get MergedXForms and Filtered Datasets as geoJSON

PR #2360, opened 2023-01-30
Render filtered datasets and merged datasets as geojson

FrankApiyo:
- Add `GeoJsonRenderer` and a way to return data as geoJSON for:
   - `MergedXFormViewset`
   - `DataViewViewset`
 
- Added test cases

  - [x] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2359

### DIFF
--- a/onadata/apps/api/viewsets/merged_xform_viewset.py
+++ b/onadata/apps/api/viewsets/merged_xform_viewset.py
@@ -18,7 +18,7 @@ from onadata.libs import filters
 from onadata.libs.renderers import renderers
 from onadata.libs.serializers.merged_xform_serializer import MergedXFormSerializer
 from onadata.libs.serializers.geojson_serializer import GeoJsonSerializer
-from onadata.libs.pagination import CountOverridablePageNumberPagination
+from onadata.libs.pagination import StandardPageNumberPagination
 
 
 # pylint: disable=too-many-ancestors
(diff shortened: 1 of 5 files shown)


onadata/apps/api/viewsets/merged_xform_viewset.py
```diff
@@ -70,8 +84,18 @@ def form(self, *args, **kwargs):
...
         merged_xform = self.get_object()
+        export_type = self.kwargs.get("format", request.GET.get("format"))
         queryset = Instance.objects.filter(
-            xform__in=merged_xform.xforms.all()
+            xform__in=merged_xform.xforms.all(),
+            deleted_at__isnull=True
         ).order_by("pk")
 
+        if export_type == "geojson":
+            page = CountOverridablePageNumberPagination(
```
KipSigei, line 94: instead of doing this you can define `pagination_class` within the class and set to `CountOverridablePageNumberPagination`

`pagination_class = CountOverridablePageNumberPagination`

then do

`page = self.paginate_queryset(queryset)`

KipSigei (changes requested):

onadata/apps/api/viewsets/merged_xform_viewset.py
FrankApiyo: Unrelated tests seem to fail when I set a `pagination_class` that applies to the whole viewset

KipSigei (approved):

PR #2362, opened 2023-01-31
Ensure that the `user` row is selected alongside the `Token`

### LINKED ISSUE #2361: User.DoesNotExist: User matching query does not exist.
Sentry Issue: [ONADATA-1T5](https://sentry.k8s.onalabs.org/organizations/ona/issues/41063/?referrer=github_integration)

```
KeyError: 'user'
  File "django/db/models/fields/related_descriptors.py", line 173, in __get__
    rel_obj = self.field.get_cached_value(instance)
  File "django/db/models/fields/mixins.py", line 15, in get_cached_value
    return instance._state.fields_cache[cache_name]

User.DoesNotExist: User matching query does not exist.
(14 additional frame(s) were not displayed)
...
  File "onadata/libs/authentication.py", line 174, in authenticate_credentials
    if not token.user.is_active:
  File "django/db/models/fields/related_descriptors.py", line 187, in __get__
    rel_obj = self.get_object(instance)
  File "django/db/models/fields/related_descriptors.py", line 307, in get_object
    return super().get_object(instance)
  File "django/db/models/fields/related_descriptors.py", line 154, in get_object
    return qs.get(self.field.get_reverse_related_filter(instance))
  File "django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
```

DavisRayM:

- Ensure that the `user` row is selected alongside the token

Closes #2361

### DIFF
--- a/onadata/libs/authentication.py
+++ b/onadata/libs/authentication.py
@@ -157,7 +157,7 @@ class TempTokenAuthentication(TokenAuthentication):
         try:
             if isinstance(key, bytes):
                 key = key.decode("utf-8")
-            token = self.model.objects.get(key=key)
+            token = self.model.objects.select_related("user").get(key=key)
         except self.model.DoesNotExist as e:
             invalid_token = True
             if getattr(settings, "SLAVE_DATABASES", []):


FrankApiyo (dismissed):

KipSigei (approved):

onadata/libs/authentication.py
```diff
@@ -157,7 +157,7 @@ def authenticate_credentials(self, key):
         try:
             if isinstance(key, bytes):
                 key = key.decode("utf-8")
-            token = self.model.objects.get(key=key)
+            token = self.model.objects.select_related("user").get(key=key)
```
KipSigei, line 160: @DavisRayM could we pin it to the master DB?

DavisRayM: I'd suggest we don't; In the except part of this we retrieve from master if we fail to retrieve from the read replica which I feel is a bit better

KipSigei (approved):

PR #2380, opened 2023-02-20
Prefix keys in filtered dataset kwargs with 'instance__'

FrankApiyo:
- [Prefix keys in filtered dataset kwargs with 'instance__'](https://github.com/onaio/onadata/pull/2380/commits/18b95bc49ebed1e2e359d8197f10a627b75b5524)
     - This will allow filtering attachments using the filtered dataset query
- [x] Replicated the issue locally with a failing tests and fixed it
- [x] QAed

  - [x] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2379

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_dataview_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_dataview_viewset.py
@@ -197,11 +197,11 @@ class TestDataViewViewSet(TestAbstractViewSet):
                       'first_name': 'Alice', 'last_name': 'A',
                       'city': 'Nairobi', 'country': 'KE'}
         alice_profile = self._create_user_profile(extra_post_data=alice_data)
+        self.extra = {"HTTP_AUTHORIZATION": f"Token {alice_profile.user.auth_token}"}
 
         # check that user with no permisisons can not list attachment objects
         request = self.factory.get("/?dataview=" + str(self.data_view.pk), **self.extra)
-        request.user = alice_profile.user
-        response = attachment_list_view(request, user=alice_profile.user)
+        response = attachment_list_view(request)
         attachments = Attachment.objects.filter(
             instance__xform=self.data_view.xform)
         self.assertEqual(0, len(response.data))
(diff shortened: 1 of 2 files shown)


onadata/libs/filters.py
```diff
@@ -357,7 +362,10 @@ def _xform_filter(self, request, view, keyword):
             xforms = xform_qs.filter(shared_data=True)
         else:
             xforms = super().filter_queryset(request, xform_qs, view) | public_forms
-        return {f"{keyword}__in": xforms}
+        return {
+            **{f"{keyword}__in": xforms},
+            **(dataview_kwargs or {})
```
DavisRayM, line 367: Does this work when it's not a dataview ? Don't see a default variable... won't the variable be undefined in some scenarios

onadata/libs/filters.py
```diff
@@ -330,18 +337,16 @@ def _xform_filter(self, request, view, keyword):
...
-            kwargs = get_filter_kwargs(self.dataview.query)
-            return {
-                **kwargs,
-                **{f"{keyword}": self.dataview.xform}
-            }
+            # filter with fitlered dataset query
+            dataview_kwargs = self._add_instance_prefix_to_dataview_filter_kwargs(
+                get_filter_kwargs(self.dataview.query))
+            xform_qs = XForm.objects.filter(pk=self.dataview.xform.pk)
         if merged_xform:
```
DavisRayM, line 344: Lets use if/elif. I don't think we expect to have a dataview, merged_xform and xform right ?

DavisRayM: Also the returned queryset for `xform_qs` at the moment will always end up as `XForm.objects.all()` due to line 358 if `xform` is non-truthy

onadata/apps/api/tests/viewsets/test_dataview_viewset.py
```diff
@@ -178,16 +179,63 @@ def test_dataview_with_attachment_field(self):
...
+        self.assertEqual(self.data_view.query,
+                         [{'value': 'no', 'column': 'pizza_fan', 'filter': '='}])
+        self.assertEqual(
+            [],
+            response.data)
+
+        # check that user with no permisisons can not view a specific attachment object
+        attachment_list_view = AttachmentViewSet.as_view({"get": "retrieve"})
+        request = self.factory.get("/?dataview=" + str(self.data_view.pk), **self.extra)
+        request.user = alice_profile.user
```
DavisRayM, line 217: Why not generate a different authentication extra for this request ? `self.extra` usually contains the authentication credentials for the test user, the request authentication class will set the correct `request.user` attribute down the line

KipSigei (approved):

PR #2381, opened 2023-02-21
Ensure default team role is assigned to users

DavisRayM:

- Only assign manager role to projects that the user has created
- Ensure default team role is assigned to new organization members & project managers
- Ensure project managers are only granted the project manager role on projects they've created

- [ ] QA

Will have an impact on the permissions users expect to have in new organizations as of now this won't affect existing permissions unless permissions are re-assigned.


FrankApiyo (approved):

DavisRayM:
Closing in-favor of #2401

PR #2396, opened 2023-03-28
Allow authenticated users to download public form exports

ciremusyoka:
- Allow authenticated users to download public form exports
- Fixes `AssertionError: renderer returned unicode, and did not specify a charset value` for .zip, .savzip and .csvzip exports

- Added tests

  - [x] Included tests
  - [ ] Updated documentation

Part of  https://github.com/onaio/zebra/issues/7439

### DIFF
--- a/onadata/libs/filters.py
+++ b/onadata/libs/filters.py
@@ -361,11 +361,11 @@ class XFormPermissionFilterMixin:
 
         if request.user.is_anonymous:
             xforms = xform_qs.filter(shared_data=True)
+        elif public_export:
+            export = get_object_or_404(Export, pk=view.kwargs.get("pk"))
+            xforms = XForm.objects.filter(pk=export.xform.pk)
         else:
             xforms = super().filter_queryset(request, xform_qs, view) | public_forms
-            if not xforms and public_export:
-                export = get_object_or_404(Export, pk=view.kwargs.get("pk"))
-                xforms = XForm.objects.filter(pk=export.xform.pk)
         return {
             **{f"{keyword}__in": xforms},
             **dataview_kwargs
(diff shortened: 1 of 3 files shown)


onadata/libs/filters.py
```diff
@@ -361,6 +361,9 @@ def _xform_filter(self, request, view, keyword):
 
         if request.user.is_anonymous:
             xforms = xform_qs.filter(shared_data=True)
+        elif public_export:
+            export = get_object_or_404(Export, pk=view.kwargs.get("pk"))
+            xforms = XForm.objects.filter(pk=export.xform.pk)
```
ukanga, line 366: Can we go with `export.xform_id` since `export.xform.pk` has a tendency to result in another hit to the DB that can be avoided here?

```
 xforms = XForm.objects.filter(pk=export.xform_id)
```

onadata/libs/filters.py
```diff
@@ -688,11 +691,11 @@ def filter_queryset(self, request, queryset, view):
         has_submitted_by_key = (
             Q(options__has_key="query") & Q(options__query__has_key="_submitted_by"),
         )
+        xform_public = _is_public_xform(view.kwargs.get("pk"))
```
ukanga, line 694: Perhaps `xform_is_public` for a variable name - much more clear?

ukanga (changes requested):

onadata/libs/filters.py
```diff
@@ -28,13 +28,13 @@
 User = get_user_model()
 
 
-def _is_public_xform(export_id: int):
+def _public_export_id_or_none(export_id: int):
```
ukanga, line 31: Should this be `_public_xform_id_or_none(export_id: int)`? The function would return the `xform_id` for the `XForm` in the given `Export` if it is a public `XForm`.

KipSigei (dismissed):

KipSigei (approved):

PR #2407, opened 2023-04-17
Prevent numeric usernames on user creation

### LINKED ISSUE #2406: Prevent users from creating an all numeric username
Currently users can create usernames with all characters numeric. This should be prevented since these users will not be able to access there profiles on the the profile endpoint.
e.g If a user has `1` as their username, when accessing their profiles `api/v1/profiles/1` they will get the profile of the user with PK `1` and not their profile.

Related to https://github.com/onaio/zebra/issues/7474

KipSigei:
- Display error message when users register using numeric usernames

- Tests
- No user will be allowed to register to the platform with numeric usernames

  - [x] Included tests
  - [ ] Updated documentation

Closes #2406

### DIFF
--- a/onadata/libs/tests/serializers/test_user_profile_serializer.py
+++ b/onadata/libs/tests/serializers/test_user_profile_serializer.py
@@ -104,12 +104,11 @@ class TestUserProfileSerializer(TestAbstractViewSet):
         self.assertEqual(alice_profile.user.username,
                          alice_serializer.data['username'])
 
-    def test_username_not_allowed_if_numeric(self):
+    def test_numeric_user_not_allowed(self):
         """
         Test ValidationError raised for numeric usernames
         """
-        # create a form instance with a purely numeric username
-        request = APIRequestFactory().get('/')
+        # create a user instance with a purely numeric username
         user_data = {
             'username': '1234',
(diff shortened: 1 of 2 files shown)


ukanga (dismissed):

DavisRayM (dismissed):

onadata/libs/tests/serializers/test_user_profile_serializer.py
```diff
@@ -103,3 +103,33 @@ def test_metadata_view_for_owner_only(self):
...
+            'password2': 'num123',
+            'name': 'Num',
+            'city': 'Nai',
+            'country': 'KE',
+            'organization': '',
+            'home_page': '',
+            'twitter': ''
+        }
+        user_prof = self._create_user_profile(extra_post_data=user_data)
+        request.user = user_prof.user
```
DavisRayM, line 126: Doesn't this create a user with the username `1234` either way ?  Seems we want to block it at the serializer level; which is cool but I think it's weird to have a test do the exact thing it wants to block in a roundabout way...

DavisRayM (dismissed):

DavisRayM (approved):

PR #2410, opened 2023-04-19
Bump action versions

DavisRayM:

- Bump action versions

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,18 @@
+# ignore everthing
+**
+# allow the following
+!onadata
+!requirements
+!manage.py
+!LICENSE
+!pyproject.toml
+!setup.cfg
+!setup.py
+!uwsgi.ini
+!extras
+!Makefile
+!docs
+
+# we finally again ignore the following from the folders allowed above
+**/*test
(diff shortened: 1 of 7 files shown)


.dockerignore
```diff
@@ -0,0 +1,22 @@
+**/__pycache__
+**/test_*
```
kelvin-muchiri, line 2: One of the opinionated approaches I prefer is the pessimistic approach. Disallow all, then allow only the folders and files you require e.g

```
# ignore everthing
**
# allow the following
!apps
!requirements
!manage.py
# we finally again ignore the following from the folders allowed above
**/*__pycache__
```

DavisRayM: LGTM, I'll change it to much this

kelvin-muchiri: I forgot to ignore tests folders again. You can confirm if we have missed any other important file or folder at the top level

```
# ignore everthing
**
# allow the following
!onadata
!requirements
!manage.py
!run.sh
# we finally again ignore the following from the folders allowed above
**/*test
**/*__pycache__
```

KipSigei (approved):

PR #2412, opened 2023-04-25
[SAV Exports] Ensure duplicate metadata fields are handled accordingly

DavisRayM:

- Revert "Add test to ensure data within groups is correctly exported in SAV"
- Revert "Skip truncating group titles for SAV Exports"
- Revert "Ensure metadata is correctly set when group names are included"
- [refactor(xform): return actual gps extra field XPath](https://github.com/onaio/onadata/pull/2412/commits/a8fbd52e022dbc4e8e44134f650fe3b792c12010)
- [refactor(csv builder): ensure get_column_names_only function works as expected with GPS Meta Fields](https://github.com/onaio/onadata/pull/2412/commits/4f68e249db9224f6768782914816666506301bc7) 

- [x] QA

Reverts the SAV Functionality to the intial state; Removes group names by default. This change also introduces unique ID suffixes to the additional GPS metadata added by the API.


onadata/apps/logger/models/xform.py
```diff
@@ -561,7 +561,7 @@ def get_additional_geopoint_xpaths(cls, xpath, remove_group_name=False):
         prefix = ""
         name = xpath
         if match:
-            prefix = "" if remove_group_name else match.groupdict()["prefix"]
```
DavisRayM, line 564: Removing this line since we are generating XPaths; If you remove the group name for an XPath it's no longer an accurate XPath to the actual data.

kelvin-muchiri (approved):

KipSigei (approved):

kelvin-muchiri (approved):

PR #2421, opened 2023-05-12
Handle `AttributeError` returned when element is None

DavisRayM:

- Handle `AttributeError` raised for multiple select choice elements

Related to https://sentry.k8s.onalabs.org/organizations/ona/issues/44054/?project=50&query=is%3Aunresolved

### DIFF
--- a/onadata/libs/utils/export_builder.py
+++ b/onadata/libs/utils/export_builder.py
@@ -1207,7 +1207,7 @@ class ExportBuilder:
                 # check if it is a choice part of multiple choice
                 # type is likely empty string, split multi select is binary
                 element = data_dictionary.get_element(xpath)
-                if element.type == SELECT_ONE:
+                if element and element.type == SELECT_ONE:
                     # Determine if all select1 choices are numeric in nature.
                     # If the choices are numeric in nature have the field type
                     # in spss be numeric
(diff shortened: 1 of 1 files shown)


onadata/libs/utils/export_builder.py
```diff
@@ -1207,7 +1207,7 @@ def _is_numeric(xpath, element_type, data_dictionary):
                 # check if it is a choice part of multiple choice
                 # type is likely empty string, split multi select is binary
                 element = data_dictionary.get_element(xpath)
-                if element.type == SELECT_ONE:
+                if element and element.type == SELECT_ONE:
```
kelvin-muchiri, line 1210: Should we add a test to ensure the code `is_all_numeric` is not called if element is none?

DavisRayM: It should be called; the element could be null when the parent is a select multiple

DavisRayM: The function checks the parent element in the next few lines

DavisRayM: What scenario would cause it to be called ? I don't think this scenario in particular requires a test...

DavisRayM: Much like line `1218`; Element are returned as `None` even when in sections in very rare cases... _When the form has a very large choice list e.t.c_

kelvin-muchiri (approved):

PR #2423, opened 2023-05-12
Add filter by email query parameter on users endpoint

### LINKED ISSUE #2422: Enhance the users endpoint to support partial match searching by user email

Enhance the users endpoint to support searching a user by email

The current `GET /api/v1/users?search=janedoe@example.com` works but has the following caveats:

- All fields specified by the `search_fields` attribute will be used in the search
- The current search by email supports exact matches only 

- The current endpoint to get users is `GET /api/v1/users.json`. We can extend the the same endpoint to add search by email `GET /api/v1/users.json?orgs=false&email=foo`
- The current response will not be altered. The current response is:
```
[{"id":2,"username":"clts","first_name":"","last_name":""},{"id":8,"username":"praekelt","first_name":"Praekelt","last_name":""}]
```
- The response will be a list of users whose email **contains** the search string in the `email` query parameter. 
- If no user is found, the response will be an empty array with a HTTP status of `200`

kelvin-muchiri:

Enable partial matching filter by email on endpoint `GET /api/v1/users` endpoint by adding the `email` query param `GET /api/v1/users?email-janedoe`

Tested locally
Added tests

Enables partial filtering by email on the `GET /api/v1/users?email=janedoe` 

  - [x] Included tests
  - [x] Updated documentation

Closes #2422

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -95,3 +95,34 @@ Response
            },
            ...
        ]
+
+
+Filter users using email
+------------------------------
+
+
+Example
+^^^^^^^
+
+::
+
+      curl -X GET https://api.ona.io/api/v1/users?email=demo@email.com
+
+Response
+^^^^^^^^
+
+::
+
+       [
+           {
+               "username": "demo",
+               "first_name": "First",
+               "last_name": "Last"
+           },
+           {
+               "username": "another_demo",
+               "first_name": "Another",
+               "last_name": "Demo"
+           },
+           ...
+       ]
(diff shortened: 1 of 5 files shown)


DavisRayM (changes requested):
Kindly clarify the use case for this; Seems like this should already be supported by the [search filter](https://www.django-rest-framework.org/api-guide/filtering/#searchfilter) not entirely sure why we would need to have both of these.

DavisRayM:
Perhaps we just need to update the search field to not be an exact match search instead ?

PR #2432, opened 2023-06-06
Tag release v3.9.1

DavisRayM:

- Tag release v3.9.1

### DIFF
--- a/onadata/libs/throttle.py
+++ b/onadata/libs/throttle.py
@@ -20,15 +20,18 @@ class RequestHeaderThrottle(SimpleRateThrottle):
         {"HTTP_USER_AGENT": "Google-HTTP-Java-Client/1.35.0 (gzip)"},
     )
 
+    def get_ident_from_header(self, value):
+        cleaned_ident = value.replace(" ", "")
+        return self.cache_format % {"scope": self.scope, "ident": cleaned_ident}
+
     def get_cache_key(self, request, _):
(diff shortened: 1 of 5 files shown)


kelvin-muchiri (dismissed):

FrankApiyo (approved):

onadata/libs/throttle.py
```diff
@@ -20,15 +20,18 @@ class RequestHeaderThrottle(SimpleRateThrottle):
...
-                # remove whitespace from key
-                cleaned_ident = ident.replace(" ", "")
-                return self.cache_format % {
-                    "scope": self.scope,
-                    "ident": cleaned_ident
-                }
+            if isinstance(value, str):
+                if header_value == value:
+                    return self.get_ident_from_header(header_value)
+            elif isinstance(value, list):
```
KipSigei, line 33: can we add a test case for the list of header values [here](https://github.com/onaio/onadata/blob/main/onadata/libs/tests/test_throttle.py#L9)

KipSigei (changes requested):

onadata/libs/throttle.py
DavisRayM: Updated this in the latest commit

KipSigei (dismissed):

KipSigei (dismissed):

KipSigei (approved):

PR #2446, opened 2023-06-26
Add task to permanently delete soft-deleted submissions

KipSigei:
- Add task to delete soft-deleted submissions after a certain period
- Added tests
- Soft deleted submissions will be permanently deleted after a certain configurable period

  - [x] Included tests
  - [ ] Updated documentation

Closes #


ukanga (commented):
Should we also check that the records have been removed from the InstanceHistory?
Also, could we verify attachments have been removed from storage locations?

onadata/apps/api/tasks.py
```diff
@@ -106,3 +110,19 @@ def send_verification_email(email, message_txt, subject):
...
     """Sends account locked email."""
     send_generic_email(email, message_txt, subject)
+
+
+@app.task()
+def delete_inactive_submissions():
+    """
+    Task to periodically delete soft deleted submissions from db
+    """
+    submissions_lifespan = getattr(settings, "INACTIVE_SUBMISSIONS_LIFESPAN", 365)
```
ukanga, line 120: This line seems to define a hard limit of 365 days if the setting does not exist. Could we take the option of simply not taking any action if the setting is not defined/set? That will maintain the status quo as is now instead of on deployment all deleted items older than a year being removed. For some installations, this may not be ideal and prior communication may be necessary.

onadata/apps/api/tests/views/test_tasks.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+        delete_inactive_submissions()
+        # test that the soft deleted submission is deleted
+        # since deleted_at is greater than specified lifespan
+        self.assertEqual(
+            self.xform.instances.filter(deleted_at__isnull=False).count(), 0
+        )
+        # test that the deletion cascades to InstanceHistory & attachments
+        self.assertEqual(Attachment.objects.all().count(), 0)
+        self.assertEqual(Attachment.objects.all().count(), 0)
```
ukanga, line 99: This line is duplicated.

onadata/apps/api/tests/views/test_tasks.py
```diff
@@ -0,0 +1,99 @@
...
+        # create instance history
+        InstanceHistory.objects.create(
+            xml=instance.xml,
+            checksum=instance.checksum,
+            xform_instance=instance,
+        )
+        history_count = InstanceHistory.objects.filter(
+            xform_instance__id=instance.pk
+        ).count()
+        self.assertEqual(history_count, 1)
```
ukanga, line 78: The InstanceHistory record is created when there has been a content edit on a submission. Can we edit instead of directly creating the object? Though I see how this approach makes the tests faster.

ukanga (changes requested):

FrankApiyo (dismissed):

onadata/apps/api/tasks.py
```diff
@@ -106,3 +111,21 @@ def send_verification_email(email, message_txt, subject):
...
+    """
+    Task to periodically delete soft deleted submissions from db
+    """
+    submissions_lifespan = getattr(settings, "INACTIVE_SUBMISSIONS_LIFESPAN", None)
+    if submissions_lifespan:
+        time_threshold = timezone.now() - timedelta(days=submissions_lifespan)
+        # delete instance attachments
+        instances = Instance.objects.filter(
+            Q(deleted_at__isnull=False) | Q(deleted_at__gte=time_threshold),
+            date_created__lte=time_threshold,
```
ukanga, line 127: Should this check be based only on the time since the record was last deleted hence using the `deleted_at` field for checks and not when the record was created?

ukanga (approved):

FrankApiyo (approved):

PR #2456, opened 2023-07-21
Enhance performance of open-data endpoint `api/v2/open-data/<uuid>/data`

kelvin-muchiri:

An improvement to the enhancements implemented [here](https://github.com/onaio/onadata/pull/2435)
 - Use raw SQL queries to improve the performance of endpoint `GET api/v2/open-data/<uuid>/data`
 - Remove ordering of the results by `id` in ascending order since when the number of submissions is large the query is extremely slow. There is also no reason to order the results by default. The consumers of the endpoint can carry out that functionality by themselves

- [x] QA

Faster response times for endpoint `GET api/v2/open-data/<uuid>/data`

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/api/viewsets/v2/tableau_viewset.py
+++ b/onadata/apps/api/viewsets/v2/tableau_viewset.py
@@ -258,10 +258,10 @@ class TableauViewSet(OpenDataViewSet):
                 sql_params = [xform.pk]
 
             if should_paginate:
-                bottom, top = self.paginator.get_offset_limit(
+                offset, limit = self.paginator.get_offset_limit(
                     self.request, self.data_count
                 )
-                sql += f" LIMIT {top} OFFSET {bottom}"
+                sql += f" LIMIT {limit} OFFSET {offset}"
 
             cursor.execute(sql, params=sql_params)
             instances = [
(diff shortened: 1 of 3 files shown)


onadata/apps/api/viewsets/v2/tableau_viewset.py
```diff
@@ -188,39 +205,74 @@ def data(self, request, **kwargs):
...
+                else:
+                    qs_kwargs = {"xform__pk": xform.pk}
+
+                if gt_id:
+                    qs_kwargs.update({"id__gt": gt_id})
+
+                self.data_count = (
+                    Instance.objects.filter(**qs_kwargs, deleted_at__isnull=True)
+                    .only("pk")
+                    .count()
```
KipSigei, line 227: Lets avoid doing an extra count query and use `xform.num_of_submissions` instead

KipSigei (changes requested):

onadata/apps/api/viewsets/v2/tableau_viewset.py
kelvin-muchiri: @KipSigei That would work, but we are applying filters so `xform.num_of_submissions` will not give us the correct result

KipSigei (approved):

PR #2460, opened 2023-08-01
Enhance performance when exporting data on endpoint `api/v1/data/<form_id>.<format>`

kelvin-muchiri:

- Remove the default ordering by id on endpoint `api/v1/data/<form_id>.<format` is making queries run extremely slow when exporting large amounts of data. To sort data by id in _ascending_ order, the query parameter `sort={"_id":1}` will be applied to the endpoint while to sort data by id in _descending_ order the query parameter `sort={"_id"-1}`. For further reference read [Sort submitted data of a specific form using existing fields](https://github.com/onaio/onadata/blob/main/docs/data.rst#sort-submitted-data-of-a-specific-form-using-existing-fields)
- Remove duplicate database queries when exporting CSV. Remove `CSVDataFrameBuilder._query_data` method which was querying the database twice despite data being passed to `CSVDataFrameBuilder.to_flat_csv_export`. Make use of the data passed to ensure lose coupling between the utility class `CSVDataFrameBuilder` and the code responsible for querying the data.
- Remove test case `onadata.apps.api.tests.viewsets.test_note_viewset.TestNoteViewSet.test_csv_export_form_w_notes`. [Notes are excluded when exporting the csv.](https://github.com/onaio/onadata/blob/main/onadata/libs/utils/csv_builder.py#L721) The column `_notes` is usually added in the CSV but the value in each record  is always overriden to be blank as per the implementation in the `CSVDataFrameBuilder` class. So this test case is futile
- Refactor code to optimize performance

- [x] QA

Faster when exporting data on endpoint `api/v1/data/<form_id>.csv`

  - [x] Included tests
  - [ ] Updated documentation

Closes #


onadata/libs/utils/export_tools.py
```diff
@@ -815,6 +844,7 @@ def generate_external_export(  # noqa C901
     filter_query = options.get("query")
     meta = options.get("meta")
     token = options.get("token")
+    sort = options.get("sort")
```
KipSigei, line 847: @kelvin-muchiri Could we add documentation for this change?

onadata/apps/api/viewsets/data_viewset.py
```diff
@@ -653,76 +658,114 @@ def list(self, request, *args, **kwargs):
 
             return Response(serializer.data)
 
-        return custom_response_handler(request, xform, query, export_type)
+        return custom_response_handler(request, xform, query, export_type, sort=sort)
```
KipSigei, line 661: I dont think its necessary to have a new `sort` param since you can get it from request

onadata/libs/utils/export_tools.py
kelvin-muchiri: @KipSigei Added the [documentation](https://github.com/onaio/onadata/pull/2460/files#diff-194a6eb60a3032894acafae20fb216b5c4c69d6eea231ed18bab3e2186abdd15R393)

KipSigei (approved):

PR #2474, opened 2023-09-12
Revert to have data exports default sorting by id

kelvin-muchiri:

- Revert to have data exports default sorting by id. Default ordering by id was removed as part of the changes made on PR https://github.com/onaio/onadata/pull/2460. This is affecting the index column hence record matching is a problem when exporting data in repeat groups.
- Fix bug sorting is not applied when fetching data on select columns that are not json fields
- Fix bug `argument 3: TypeError: wrong type` raised when using including GIS fields in `Manager.raw` when exporting non-OSM data by avoid `SELECT *` and selecting only the fields of interest

- [x] QA

Default ordering on endpoint `api/v1/data/<form_id>.<format>` by id ascending order is restored

  - [x] Included tests
  - [ ] Updated documentation

Closes #


onadata/apps/api/tests/viewsets/test_tableau_viewset.py
```diff
@@ -396,7 +397,7 @@ def test_gt_id_query_param(self):
         self.view = TableauViewSet.as_view({"get": "data"})
         _open_data = get_or_create_opendata(self.xform)
         uuid = _open_data[0].uuid
-        request = self.factory.get("/", data={"gt_id": 100000}, **self.extra)
+        request = self.factory.get("/", data={"gt_id": sys.maxsize}, **self.extra)
```
ukanga, line 400: Why does this need to change?

kelvin-muchiri: @ukanga It works fine, but was thing `sys.maxsize` would be better in trying to simulate an unrealistic number that cannot be achievable when running tests

ukanga: ok

FrankApiyo (approved):

KipSigei (approved):

PR #2477, opened 2023-09-15
Data endpoint enhancements

### LINKED ISSUE #2472: Noticed the following ONA Data API limitations:
Noticed the following ONA Data API limitations:
1. Some submissions don't have the `_date_modified` field. This is a challenge as that's the field we use as a cursor field
	- The workaround of using either the `_last_edited` or the `_submission_time` is patchy and creates loopholes
2. The API returns wrong records against logical filters
	- `?query={"_date_modified": {"$gt": "2023-07-18T19:09:52"}}` returns records with `_date_modified` matching the exact filter
	- `?query={"_date_modified": {"$gte": "2023-07-18T19:09:52"}}` has the same effect
3. The API does not support filtering using date fields with timezones
	- `&query={"_date_modified": {"$gt": "2023-07-05T16:56:32.172+02:00"}}`
	- we have to programmatically remove the timezone info even though it's gotten from the same API (from the cursor fields `_date_modified`, `_last_edited`, and `_submission_time`
4. The API does not return pagination information (the `links` header) when using sorting
	- `?page=1&page_size=2&sort={"_date_modified": 1}` this does not return links header
	- `?page=1&page_size=2` this does

_Originally posted by @machariamuguku in https://github.com/onaio/source-ona-data/issues/19#issuecomment-1702914169_

### LINKED ISSUE #2480: Instance model date_modified and date_created fields are out of sync with their aliases in the json field
### Environmental Information

- Onadata version: <= 3.13.1

### Problem description

`_date_modified` and `_submission_time` for data returned on endpoint `/api/v1/data` are always behind the model's `date_modified` and `date_created` fields respectively.

`auto_now`  and `auto_now_add` are updated when calling Model.save. That means line `[DATE_MODIFIED] = self.date_modified.strftime(MONGO_STRFTIME)` will always be assigned the previous value of `date_modified` during save while model field `date_modified` is now updated with the current time.

Therefore `Instance.json.get("_date_modified")` will always be referencing the previous `date_modified` value which is incorrect.

### Expected behavior

`Instance.date_modified` should match `Instance.json.get("_date_modified")` and `Instance.date_created` should match `Instance.json.("_submission_time")`

Create a submission. On the Django interactive shell, compare `Instance.date_modified` and `Instance.get("_date_modified")`, you'll find that they do not match. Similar findings will be revealed when you compare `Instance.date_created` and `Instance.json.("_submission_time")`.

Call save on the submission. Perform the comparisons again. You'll find the model fields are always ahead in time than the json fields

kelvin-muchiri:

- Fix pagination link header missing in `/api/v1/data/<form_id>` when [sort query parameter](https://github.com/onaio/onadata/blob/main/docs/data.rst#sort-submitted-data-of-a-specific-form-using-existing-fields) is applied with [pagination](https://github.com/onaio/onadata/blob/main/docs/data.rst#paginate-data-of-a-specific-form)
- Disable Python scripts in migration file `onadata/apps/logger/models/migrations/001_pre-django-3-upgrade.py` preventing newer migrations for model XForm from being applied. This [issue](https://github.com/onaio/onadata/issues/2481) tracks this bug.
- Fix #2480 

- [x] QA

No side effects

  - [x] Included tests
  - [x] Updated documentation

Closes #2472 #2480


onadata/apps/api/tasks.py
```diff
@@ -145,3 +150,34 @@ def send_project_invitation_email_async(
...
         email.send()
+
+
+@app.task(track_started=True)
+def regenerate_form_instance_json(xform_id: int):
+    """Regenerate a form's instances json
+
+    Json data recreated afresh and any existing json data is overriden
+    """
+    try:
```
KipSigei, line 161: Can we wrap this block in a condition such that if `xform.is_instance_json_regenerated` is True then we dont need to execute it

KipSigei (changes requested):

docs/forms.rst
```diff
@@ -1628,3 +1628,42 @@ If the upload is still running:
...
+            "status": "SUCCESS"
+        }
+
+
+If the job is still in progress
+
+::
+
+        {
+            "status": "STARTED"
```
ukanga, line 1668: This should be a management command and not an API.

ukanga (changes requested):

KipSigei (approved):

PR #2479, opened 2023-09-18
Log auth details

KipSigei:

  - [ ] Included tests
  - [ ] Updated documentation

Closes #


onadata/libs/authentication.py
```diff
@@ -114,14 +121,23 @@ def authenticate(self, request):
...
             )
+            logger.info(info_str)
+            error_str = _(
+                "Invalid username/password. "
+                f"For security reasons, after {remaining_attempts} more failed "
+                f"login attempts you'll have to wait {lockout_time} minutes "
+                "before trying again."
+            )
+            # log to sentry
+            report_exception("Authentication Failure", error_str)
```
kelvin-muchiri, line 138: Should we log to sentry the `info_str` instead of `error_str`?

onadata/libs/authentication.py
```diff
@@ -114,14 +121,23 @@ def authenticate(self, request):
...
-                    f"login attempts you'll have to wait {lockout_time} minutes "
-                    "before trying again."
-                )
+            ip_address, username = retrieve_user_identification(request)
+            user_agent = request.META.get("HTTP_USER_AGENT", None)
+            info_str = (
+                f"IP: {ip_address}, USERNAME: {username}, "
+                f"REMAINING_ATTEMPTS: {remaining_attempts}, USER_AGENT: {user_agent}"
             )
+            logger.info(info_str)
```
ukanga, line 130: This information would only be relevant when debugging a service and not always. I recommend we use `logger.debug()`.

PR #2496, opened 2023-10-30
Create user auth token if doesn't exist on user get

### LINKED ISSUE #2490: Users without API Token unable to login
### Environmental Information
- Onadata version: v3.13.1

### Problem description
When attempting to log in, users without an API token encounter an error and are unable to proceed. The error message received is
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python3.10/dist-packages/django/db/models/fields/related_descriptors.py", line 421, in __get__
    raise self.RelatedObjectDoesNotExist(
django.contrib.auth.models.User.auth_token.RelatedObjectDoesNotExist: User has no auth_token.
```
### Expected behavior
Users should be able to log in successfully, regardless of whether they have an existing API token or not. If a user does not have a token, the system should automatically generate one during the login process to enable a seamless login experience.

1. Attempt to log in with a user account that does not have an associated API token.

ciremusyoka:
- Catch error experienced when getting user without auth token and create user token if doesn't exist

- Added test

  - [x] Included tests
  - [ ] Updated documentation

Closes #2490

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_connect_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_connect_viewset.py
@@ -105,6 +105,28 @@ class TestConnectViewSet(TestAbstractViewSet):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(dict(response.data), self.data)
 
+    def test_get_profile_user_no_auth_token(self):
+        # delete auth token
+        token = Token.objects.get(user=self.user)
+        old_token_key = token.key
+        token.delete()
+
+        view = ConnectViewSet.as_view(
+            {"get": "list"},
+            authentication_classes=(
+                DigestAuthentication,
+                authentication.BasicAuthentication,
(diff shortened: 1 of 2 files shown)


onadata/apps/api/tests/viewsets/test_connect_viewset.py
```diff
@@ -105,6 +105,28 @@ def test_get_profile(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(dict(response.data), self.data)
 
+    def test_get_profile_user_no_auth_token(self):
+        # delete auth token
```
kelvin-muchiri, line 109: Could we add a doc string?

KipSigei (approved):

PR #2503, opened 2023-11-09
Soft delete xform from legacy UI

FrankApiyo:
Don't call `xform.delete()` method which soft deletes

  - [x] Included tests
  - [x] Updated documentation

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_xform_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_xform_viewset.py
@@ -59,6 +59,7 @@ from onadata.apps.api.viewsets.xform_viewset import XFormViewSet
 from onadata.apps.logger.models import Attachment, Instance, Project, XForm
 from onadata.apps.logger.models.xform_version import XFormVersion
 from onadata.apps.logger.xform_instance_parser import XLSFormError
+from onadata.apps.logger.views import delete_xform
 from onadata.apps.main.models import MetaData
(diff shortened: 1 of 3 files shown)


onadata/apps/api/tests/viewsets/test_xform_viewset.py
```diff
@@ -4351,6 +4352,42 @@ def test_delete_xform_also_deletes_linked_dataviews(self):
...
+                "tests",
+                "utils",
+                "fixtures",
+                "tutorial",
+                "instances",
+                "uuid{}".format(x),
+                "submission.xml",
+            )
+            self._make_submission(path)
+            x += 1
```
kelvin-muchiri, line 4377: Why do we increment?

FrankApiyo: There's no need for, this, I've removed this line

onadata/apps/logger/views.py
```diff
@@ -504,8 +503,8 @@ def delete_xform(request, username, id_string):
         {"user__username__iexact": username, "id_string__iexact": id_string}
     )
 
-    # delete xform and submissions
-    remove_xform(xform)
```
kelvin-muchiri, line 508: Should we also delete the function `remove_xform` since it is no longer required?

FrankApiyo: done!

kelvin-muchiri (approved):

PR #2558, opened 2024-02-22
Improve perfomance of /status endpoint

FrankApiyo:


kelvin-muchiri (approved):

onadata/apps/main/views.py
```diff
@@ -1576,6 +1576,15 @@ def service_health(request):
     else:
         service_statuses["Cache-Service"] = "OK"
 
+    onadata_version = os.environ.get("ONADATA_VERSION")
```
KipSigei, line 1579: you can get the onadata version from the `__version__` attribute

KipSigei (changes requested):

KipSigei (approved):

PR #2565, opened 2024-03-07
Allow unauthenticated HEAD requests in IsAuthenticatedSubmission class

FrankApiyo:
- Allow unauthenticated HEAD requests to f"{form_pk}/submission" endpoint
- Updated tests

  - [x] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2568

### DIFF
--- a/onadata/apps/api/permissions.py
+++ b/onadata/apps/api/permissions.py
@@ -552,7 +552,10 @@ class IsAuthenticatedSubmission(BasePermission):
                 return False
             profile, _ = UserProfile.objects.get_or_create(user=user)
 
-            if profile.require_auth:
+            if profile.require_auth and not (
+                request.path.endswith(f"{form_pk}/submission")
+                and request.method == "HEAD"
+            ):
                 # raises a permission denied exception,
                 # forces authentication
                 return False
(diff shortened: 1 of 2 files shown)


onadata/apps/api/permissions.py
```diff
@@ -535,10 +537,14 @@ class IsAuthenticatedSubmission(BasePermission):
...
     def has_permission(self, request, view):
-        username = view.kwargs.get("username")
         form_pk = view.kwargs.get("xform_pk")
-        project_pk = view.kwargs.get("project_pk")
-        if request.method in ["HEAD", "POST"] and request.user.is_anonymous:
+        if request.method == "HEAD" and request.path.endswith(
+            f"enketo/{form_pk}/submission"
+        ):
+            return True
+        elif request.method in ["HEAD", "POST"] and request.user.is_anonymous:
```
ukanga, line 545: Could we check that the request.user is authenticated? It is by passing permissions checks in the section below. perhaps what needs to be revisited is lines 561 - 564:

```python
...
    if profile.require_auth:
       return False
...
```

I wonder how does it affect head requests only and not post requests as well?

ukanga (changes requested):

onadata/apps/api/permissions.py
ukanga: @FrankApiyo could you add some test(s) - it could provide better clarity on the issue that is being addressed.

ukanga (changes requested):
Let's file a bug report with Enketo. This change is not necessary.

PR #2572, opened 2024-03-21
Django 4 upgrade

ukanga:

Updating dependences and code to work with Django 4+

  - [ ] Included tests
  - [ ] Updated documentation

Closes #


requirements/azure.pip
```diff
@@ -22,20 +22,18 @@ cryptography==42.0.5
     # via
     #   -r requirements/azure.in
     #   azure-storage-blob
-django==3.2.25
+django==4.0
```
KipSigei, line 25: we should update to 4.2 which is currently in LTS, 4.0 ended security support [11 months ago](https://endoflife.date/django)

ukanga: Oh, yes. I just remembered that my strategy was to go through all the versions but start with 4.0. I will take this back to the draft.

setup.cfg
```diff
@@ -102,6 +101,7 @@ install_requires =
     # Deprecation tagging
     deprecated
     # Redis cache
+    redis[hiredis]
     django-redis
```
KipSigei, line 105: Note that in django 4 we dont require the `django-redis` package as its now implemented via `django.core.cache.backends.redis.RedisCache`

ukanga: We may need a bit more extensive testing for the switch. I recommend we do that separately and not as part of this upgrade. For more context on `django-redis` [see](https://github.com/jazzband/django-redis/issues/585#issuecomment-1002051473).

KipSigei (dismissed):

KipSigei (approved):

KipSigei (approved):

PR #2587, opened 2024-04-15
Add xform to attachment model

ukanga:

Add a direct link between an attachment to the XForm (form for the submission the attachment belongs to) and User (submitted user) models.

Existing records will have the field blank and may need to be updated to have this value for the benefit to be optimal for old records.

  - [ ] Included tests
  - [ ] Updated documentation

Closes #


onadata/apps/logger/migrations/0014_populate_attachment_xform.py
```diff
@@ -0,0 +1,35 @@
...
+    print(f"Found {count} records")
+
+    for attachment in queryset.iterator(chunk_size=100):
+        # We do not want to trigger Model.save or any signal
+        # Queryset.update is a workaround to achieve this.
+        # Instance.save and the post/pre signals may contain
+        # some side-effects which we are not interested in e.g
+        # updating date_modified which we do not want
+        Attachment.objects.filter(pk=attachment.pk).update(
+            xform=attachment.instance.xform
```
ukanga, line 20: Can we use `xform = attachment.instance.xform_id` instead? 

Also, can we add the submission instance user i.e. `user=attachment.instance.user_id`? It is the other new field introduced in this PR.

onadata/apps/logger/migrations/0014_populate_attachment_xform.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 4.2.11 on 2024-04-22 06:42
+
+from django.db import migrations
+
+
+def populate_attachment_xform(apps, schema_editor):
+    """Populate xform field for Attachments"""
+    Attachment = apps.get_model("logger", "Attachments")
+    queryset = Attachment.objects.filter(xform__isnull=True)
```
ukanga, line 9: We are only interested in the `pk`, `instance.xform_id`, and `instance.user_id`. Could we request only for those values?

```python
   queryset = Attachment.objects.filter(xform__isnull=True).values('pk', 'instance__xform', 'instance__user')
...
```

To minimise unnecessary queries triggered below when `atachment.instance.xform` and `attachment.instance.user` are accessed.

KipSigei (approved):

PR #2597, opened 2024-05-17
Update Docker build

ukanga:
- Use pyenv installed python 3.10.14 in Docker image
- Use Ubuntu 22.04 base image
- Updated GitHub actions versions

- Image size reduction to 1.6 GB from 1.9GB


kelvin-muchiri (dismissed):

docker/onadata-uwsgi/Dockerfile.ubuntu
```diff
@@ -1,110 +1,102 @@
-FROM python:3.10 as intermediate
+FROM onaio/python-deps:3.10.14 as base
 
+ARG release_version=v4.0.1
```
kelvin-muchiri, line 3: @ukanga Are we using this arg?

ukanga: I don't think we are, and for some reason, I have been unwilling to remove it. I suppose there is a workflow where that may make sense but I have yet to try and make it work.

kelvin-muchiri (approved):

PR #2625, opened 2024-06-27
Create and delete EntityList via API

kelvin-muchiri:

- Ability to create EntityList via API
- Ability to delete EntityList via API

- [ ] QA

  - [x] Included tests
  - [ ] Updated documentation


onadata/apps/api/permissions.py
```diff
@@ -560,11 +560,26 @@ def has_permission(self, request, view):
...
 
+        if request.user.is_authenticated and view.action == "create":
+            project_id = request.data.get("project")
+
+            if project_id:
+                try:
+                    project = Project.objects.get(pk=int(project_id))
+
+                except (ValueError, Project.DoesNotExist):
+                    # Return as validation will take care of this
```
ukanga, line 580: Does returning false cause issues a this stage - could we handle the false that is returned?

kelvin-muchiri: @ukanga If we return `False`, this will return status `403`. A person who is authorized to create and makes a request with an invalid `project`, they should get a `400` response instead of a`403`. I chose to return `True` so that request will reach the serializer and a `400` error is returned.

Maybe for code clarity, we should raise a validation error at this stage?

kelvin-muchiri: @ukanga I have refactored to raise validation error message

ukanga: Why is the permission check happening before the validation check? Is there a different workflow we should consider?

kelvin-muchiri: @ukanga The idea is to return 400 (Project does not exist) if the project literally does not exist, or if the user has no permission to view the project. If the user has permission to view the project but does not have permission to create a dataset for the project, they get 403. 

Maybe I need to add better comments to communicate the message above

ukanga (approved):

PR #2642, opened 2024-07-10
Optimize updates to project `date_modified` field when submitting data

KipSigei:
- Update the project `date_modified` field in batches periodically

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_project_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_project_viewset.py
@@ -42,6 +42,7 @@ from onadata.apps.logger.models import (
     XFormVersion,
 )
 from onadata.apps.main.models import MetaData
+from onadata.apps.logger.tasks import apply_project_date_modified_async
 from onadata.libs import permissions as role
 from onadata.libs.models.share_project import ShareProject
 from onadata.libs.permissions import (
@@ -1945,6 +1946,9 @@ class TestProjectViewSet(TestAbstractViewSet):
         self.assertNotEqual(last_date, current_last_date)
 
         self._make_submissions()
+
+        # run cronjob to update date_modified field
+        apply_project_date_modified_async.delay()
 
         self.project.refresh_from_db()
         self.assertNotEqual(current_last_date, self.project.date_modified)
--- a/onadata/apps/api/tests/viewsets/test_project_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_project_viewset.py
@@ -42,7 +42,6 @@ from onadata.apps.logger.models import (
     XFormVersion,
 )
 from onadata.apps.main.models import MetaData
-from onadata.apps.logger.tasks import apply_project_date_modified_async
 from onadata.libs import permissions as role
 from onadata.libs.models.share_project import ShareProject
 from onadata.libs.permissions import (
@@ -1944,14 +1943,6 @@ class TestProjectViewSet(TestAbstractViewSet):
         current_last_date = self.project.date_modified
 
         self.assertNotEqual(last_date, current_last_date)
-
-        self._make_submissions()
-
-        # run cronjob to update date_modified field
-        apply_project_date_modified_async.delay()
-
-        self.project.refresh_from_db()
-        self.assertNotEqual(current_last_date, self.project.date_modified)
 
     def test_anon_project_form_endpoint(self):
         self._project_create()
--- a/onadata/apps/logger/models/instance.py
+++ b/onadata/apps/logger/models/instance.py
@@ -34,6 +34,7 @@ from onadata.apps.logger.xform_instance_parser import (
 from onadata.celeryapp import app
 from onadata.libs.data.query import get_numeric_fields
 from onadata.libs.utils.cache_tools import (
+    BATCH_PROJECT_IDS_CACHE,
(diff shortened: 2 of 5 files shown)


onadata/apps/logger/models/instance.py
```diff
@@ -391,7 +391,15 @@ def update_project_date_modified(instance_id, _):
...
-        instance.xform.project.save(update_fields=["date_modified"])
+        # update project date_modified using raw SQL
+        with connection.cursor() as c:
+            sql = (
+                "UPDATE logger_project SET "
+                "date_modified = %s "
+                "WHERE id = %s"
+            )
+            params = [timezone.now(), instance.xform.project.pk]
+            c.execute(sql, params)
```
ukanga, line 402: Could we use the Queryset approach which also does not trigger an `post/pre_save` signals?
```python
Project.objects.filter(pk=instance.xform.project_id).update(date_modified=timezone.now())
```

ukanga (commented):
Should we rethink how we apply these updates rather than for every single submission to once every time window we apply the update?
We could log the project ids for a particular time period and then apply the update in batches when the time window ends.

onadata/apps/logger/models/instance.py
```diff
@@ -391,7 +392,13 @@ def update_project_date_modified(instance_id, _):
         if current_task.request.id:
             raise e
     else:
-        instance.xform.project.save(update_fields=["date_modified"])
+        timeout = getattr(settings, "PROJECT_IDS_CACHE_TIMEOUT", 3600)
+        project_id = instance.xform.project_id
+
+        # Log project id and date motified in cache with timeout
+        project_ids = cache.get(BATCH_PROJECT_IDS_CACHE, {})
+        project_ids[project_id] = timezone.now()
```
kelvin-muchiri, line 400: @KipSigei Can we use `instance.date_modified` instead of `timezone.now()`?

onadata/apps/logger/tests/test_tasks.py
```diff
@@ -61,3 +67,29 @@ def test_invalid_pk(self, mock_logger, mock_set_perms):
...
+        apply_project_date_modified_async.delay()
+        self.project.refresh_from_db()
+        current_date_modified = self.project.date_modified
+
+        # check that date_modified has changed
+        self.assertNotEqual(initial_date_modified, current_date_modified)
+
+        # check if current date modified is greater than initial
+        self.assertGreater(current_date_modified, initial_date_modified)
+        cache.delete(BATCH_PROJECT_IDS_CACHE)
```
kelvin-muchiri, line 95: I think we should assert that the cache is None since it is deleted after the task completes

onadata/libs/utils/cache_tools.py
```diff
@@ -60,6 +60,9 @@
 # Cache timeouts used in XForm model
 XFORM_REGENERATE_INSTANCE_JSON_TASK_TTL = 24 * 60 * 60  # 24 hrs converted to seconds
 
+# Batch project IDs
+BATCH_PROJECT_IDS_CACHE = "batch_project_ids"
```
kelvin-muchiri, line 64: IMO, the constant can be renamed to something like `PROJECT_DATE_MODIFIED_CACHE`

onadata/apps/logger/tasks.py
```diff
@@ -25,3 +27,20 @@ def set_entity_list_perms_async(entity_list_id):
...
+    """
+    Batch update projects date_modified field periodically
+    """
+    project_ids = cache.get(BATCH_PROJECT_IDS_CACHE, {})
+    if not project_ids:
+        return
+
+    # Update project date_modified field in batches
+    for project_id, timestamp in project_ids.items():
+        Project.objects.filter(pk=project_id).update(date_modified=timestamp)
```
kelvin-muchiri, line 43: It's very rare that a project may not exist, but is possible that it could have been hard deleted. I think we need to handle for the case where the project does not exist

onadata/apps/api/tests/viewsets/test_project_viewset.py
```diff
@@ -1946,6 +1947,9 @@ def test_last_date_modified_changes_when_adding_new_form(self):
 
         self._make_submissions()
 
+        # run cronjob to update date_modified field
+        apply_project_date_modified_async.delay()
```
kelvin-muchiri, line 1951: @KipSigei Did we make any changes to how a `Project.date_modified` is updated when a form is published?

KipSigei: @kelvin-muchiri On the line above when a submission is made, with the changes introduced in this PR the `date_modified` field won't change until the cronjob is run. This hasn't changed how the date_modified field is updated when publishing a new form

onadata/apps/logger/tasks.py
KipSigei: @kelvin-muchiri The queryset checks for the existence of the project and updates the `date_modified` field if the project exists, and if it doesn't exist the update method will simply not update any rows

kelvin-muchiri: @KipSigei Oh, I see

kelvin-muchiri (dismissed):

kelvin-muchiri (approved):

PR #2651, opened 2024-07-25
Use eventual consistency to update EntityList num_entities

kelvin-muchiri:

Use eventual consistency to update `EntityList` `num_entities` in the database. Store counter in the cache and add a task that should be run periodically to persist the counter in the database.

Add a failover mechanism to persist cached data in the database incase the periodic task is not running or has malfunctioned.

- [x] QA
- [x] Tests

Less load on the database to update `num_entities`

  - [x] Included tests
  - [ ] Updated documentation

Closes #


faith-mutua:
@kelvin-muchiri this is working as expected, the count in the Entity List is incremented correctly when multiple entities are submitted concurrently. The count also decrements correctly when entities are deleted from the Entity List

ukanga (dismissed):

FrankApiyo (approved):

PR #2677, opened 2024-08-25
Fix exception thrown when updating organization profile

FrankApiyo:
Remove email from `validated_data` before updating the organization profile

Updated test cases

  - [x] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2676


onadata/libs/serializers/organization_serializer.py
```diff
@@ -73,6 +73,10 @@ def update(self, instance, validated_data):
             instance.user.last_name = last_name
             instance.user.save()
 
+        if "email" in validated_data:
+            instance.user.email = validated_data.pop("email")
+            instance.user.save()
```
kelvin-muchiri, line 78: We can avoid saving twice by taking `instance.user.save()` out of the `if` block and removing `instance.user.save()` in `if` block above it.

kelvin-muchiri (approved):

PR #2689, opened 2024-08-30
Add CustomScopedRateThrolle throttling class

FrankApiyo:
This will throttle anonymous enketo users more fairly

  - [ ] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/libs/tests/test_throttle.py
+++ b/onadata/libs/tests/test_throttle.py
@@ -20,7 +20,10 @@ class CustomScopedRateThrottleTest(TestCase):
         request.user = AnonymousUser()
         self.throttle.scope = "submission"
         cache_key = self.throttle.get_cache_key(request, None)
-        self.assertEqual(cache_key, "throttle_submission_/enketo/1234/submission")
+        self.assertEqual(
+            cache_key,
+            "throttle_submission_/enketo/1234/submission_127.0.0.1"
+        )
 
     def test_users_get_throttled_based_on_uri_path(self):
         request = self.factory.get("/enketo/1234/submission")
--- a/onadata/libs/tests/test_throttle.py
+++ b/onadata/libs/tests/test_throttle.py
@@ -1,15 +1,37 @@
 from django.core.cache import cache
+from django.contrib.auth.models import AnonymousUser, User
 from django.test import TestCase, override_settings
 
 from rest_framework.test import APIRequestFactory
 
-from onadata.libs.throttle import RequestHeaderThrottle
+from onadata.libs.throttle import RequestHeaderThrottle, CustomScopedRateThrottle
+
+class CustomScopedRateThrottleTest(TestCase):
+    def setUp(self):
+        """
+        Reset the cache so that no throttles will be active
+        """
+        cache.clear()
+        self.factory = APIRequestFactory()
+        self.throttle = CustomScopedRateThrottle()
+
+    def test_anonymous_users_get_throttled_based_on_uri_path(self):
+        request = self.factory.get("/enketo/1234/submission")
+        request.user = AnonymousUser()
+        self.throttle.scope = "submission"
+        cache_key = self.throttle.get_cache_key(request, None)
+        self.assertEqual(cache_key, "throttle_submission_/enketo/1234/submission")
+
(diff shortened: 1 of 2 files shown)


onadata/libs/throttle.py
```diff
@@ -4,7 +4,19 @@
...
+from rest_framework.throttling import SimpleRateThrottle, ScopedRateThrottle
+
+
+class CustomScopedRateThrottle(ScopedRateThrottle):
+    """
+    Custom throttling for fair throttling for anonymous users sharing IP
+    """
+
+    def get_cache_key(self, request, view):
+        if request.user and request.user.is_authenticated:
```
kelvin-muchiri, line 16: Should we do
```
def get_cache_key(self, request, view):
    xff = request.META.get('HTTP_X_FORWARDED_FOR')

    if (request.user and request.user.is_authenticated) or xff is not None:
        return super().get_cache_key(request, view)

    return f'throttle_{self.scope}_{request.path}'

```

FrankApiyo: I'm not sure about this since [get_ident](https://github.com/encode/django-rest-framework/blob/master/rest_framework/throttling.py#L197) is only called when request.user is `None`

I think we should use get_ident to make the cache key more speicif

kelvin-muchiri (approved):

kelvin-muchiri (dismissed):

onadata/libs/tests/test_throttle.py
```diff
@@ -1,15 +1,40 @@
...
 
 from rest_framework.test import APIRequestFactory
 
-from onadata.libs.throttle import RequestHeaderThrottle
+from onadata.libs.throttle import RequestHeaderThrottle, CustomScopedRateThrottle
+
+class CustomScopedRateThrottleTest(TestCase):
+    def setUp(self):
+        """
+        Reset the cache so that no throttles will be active
```
kelvin-muchiri, line 12: @FrankApiyo Looks like this should be a comment and not a docstring

onadata/libs/tests/test_throttle.py
```diff
@@ -1,15 +1,40 @@
...
+        request = self.factory.get("/enketo/1234/submission")
+        request.user = AnonymousUser()
+        self.throttle.scope = "submission"
+        cache_key = self.throttle.get_cache_key(request, None)
+        self.assertEqual(
+            cache_key,
+            "throttle_submission_/enketo/1234/submission_127.0.0.1"
+        )
+
+    def test_users_get_throttled_based_on_uri_path(self):
```
kelvin-muchiri, line 28: @FrankApiyo The name of the test appears incorrect since authenticated users are throttled by id

onadata/libs/throttle.py
```diff
@@ -4,7 +4,19 @@
 
 from django.conf import settings
 
-from rest_framework.throttling import SimpleRateThrottle
+from rest_framework.throttling import SimpleRateThrottle, ScopedRateThrottle
+
+
+class CustomScopedRateThrottle(ScopedRateThrottle):
```
kelvin-muchiri, line 10: @FrankApiyo Maybe we can rename this to `SharedIPScoppedThrottle` ?

kelvin-muchiri (dismissed):

onadata/libs/tests/test_throttle.py
```diff
@@ -1,20 +1,41 @@
...
+from onadata.libs.throttle import RequestHeaderThrottle, CustomScopedRateThrottle
+
+class CustomScopedRateThrottleTest(TestCase):
+    def setUp(self):
+        # Reset the cache so that no throttles will be active
+        cache.clear()
+        self.factory = APIRequestFactory()
+        self.throttle = CustomScopedRateThrottle()
+
+    def test_anonymous_users_get_throttled_based_on_uri_path(self):
```
kelvin-muchiri, line 16: @FrankApiyo One trick to make test names shorter is to make use of a docstring e.g

```
def test_anonymous_users(self):
    """Anonymous users  get throttled base on URI path"""
```

FrankApiyo: Thanks

kelvin-muchiri (approved):

PR #2691, opened 2024-09-02
Add user activity logs to messaging endpoint

ciremusyoka:
Adds message notifications for the following actions:
1. user_created
2. user_updated
3. password_changed

- Added tests

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/api/viewsets/user_profile_viewset.py
+++ b/onadata/apps/api/viewsets/user_profile_viewset.py
@@ -7,6 +7,7 @@ import datetime
 import json
 
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.contrib.auth.password_validation import validate_password
 from django.core.cache import cache
 from django.core.validators import ValidationError
@@ -32,6 +33,10 @@ from onadata.apps.api.tasks import send_verification_email
 from onadata.apps.api.tools import get_baseviewset_class
 from onadata.apps.logger.models.instance import Instance
 from onadata.apps.main.models import UserProfile
+from onadata.apps.messaging.constants import (
+    USER, USER_CREATED, USER_UPDATED, USER_PASSWORD_CHANGED
+)
+from onadata.apps.messaging.serializers import send_message
 from onadata.libs import filters
 from onadata.libs.mixins.authenticate_header_mixin import AuthenticateHeaderMixin
 from onadata.libs.mixins.cache_control_mixin import CacheControlMixin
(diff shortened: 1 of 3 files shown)


onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -228,6 +246,17 @@ def create(self, request, *args, **kwargs):
         profile = response.data
         user_name = profile.get("username")
         cache.set(f"{USER_PROFILE_PREFIX}{user_name}", profile)
+
+        # send notification on creating new user account
+        user = User.objects.get(username__iexact=user_name)
```
ukanga, line 251: Could we fetch user via the `id` instead?

```
user = User.objects.get(id=profile.get("id")
```

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -228,6 +246,17 @@ def create(self, request, *args, **kwargs):
...
+        send_message(
+            instance_id=user_name,
+            target_id=user.id,
+            target_type=USER,
+            user=user,
+            message_verb=USER_CREATED,
+        )
```
ukanga, line 258: Could we move this directly into the `UserProfileSerializer` class instead? I think the above additional query for the user will be unnecessary in such a case.

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -211,6 +219,16 @@ def update(self, request, *args, **kwargs):
         username = kwargs.get("user")
         response = super().update(request, *args, **kwargs)
         cache.set(f"{USER_PROFILE_PREFIX}{username}", response.data)
+
+        # send notification on updating user profile
+        send_message(
```
kelvin-muchiri, line 224: We can also move this to the serializer. Since we are sending the messaging for both `create` and `update`, we can override the serializer's `save` method and call `send_message` there. [Example](https://github.com/onaio/onadata/blob/main/onadata/libs/serializers/share_project_serializer.py#L143)

We can differentiate create from update by checking [self.instance](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L202). When doing an update `self.instance` is not None

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -303,6 +330,16 @@ def partial_update(self, request, *args, **kwargs):
 
             profile.metadata = metadata
             profile.save()
+
+            # send notification on updating user profile
+            send_message(
```
kelvin-muchiri, line 335: `partial_update` will call the serializer's `save` method. If we refactor  to send the message for `update` within the serializer, we don't need this change

PR #2707, opened 2024-09-20
Create composite index on logger_instance fields xform_id, date_created

kelvin-muchiri:
### Changes

Create composite index on logger_instance fields xform_id, date_created


kelvin-muchiri:
The frequency of accessing by xform_id and date_created is high. The cost of maintaining the index will likely outweigh the benefts.

Issue #2679, opened 2024-08-26
Make user active status configurable


ciremusyoka:
Currently, all users created are set to be active by default [here](https://github.com/onaio/onadata/blob/main/onadata/libs/serializers/user_profile_serializer.py#L302). To enhance flexibility, we should introduce a configuration option to set newly created users as active or inactive by default.

PR #2711, opened 2024-09-26
Make user active status configurable

inv-amalusudhakaran:
Make user active status configurable #2679

Changes made:
Added USER_DEFAULT_IS_ACTIVE setting in settings/common.py to control the default active status of newly created users.
Modified the user creation logic to use the USER_DEFAULT_IS_ACTIVE setting.


kelvin-muchiri:
@inv-amalusudhakaran Add tests and also have the default as true if the setting is not available

PR #2715, opened 2024-10-02
Add/remove user from organization synchronously

kelvin-muchiri:

Since an organization is a Group, adding/removing user from an organization does not take long and can be done synchronously. Sharing organization projects will take longer and remains asynchronous.

- [x] QA

No delay in a user being added to an organization. Clients are able to get the updated members list immediately

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/libs/serializers/organization_member_serializer.py
+++ b/onadata/libs/serializers/organization_member_serializer.py
@@ -2,18 +2,18 @@
 """
 The OrganizationMemberSerializer - manages a users access in an organization
 """
+from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.mail import send_mail
 from django.utils.translation import gettext as _
 
 from rest_framework import serializers
 
 from onadata.apps.api.tools import (
     _get_owners,
+    add_user_to_organization,
     get_organization_members,
-)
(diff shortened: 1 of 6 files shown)


onadata/libs/serializers/organization_member_serializer.py
```diff
@@ -108,12 +108,18 @@ def create(self, validated_data):
...
-                )
+                add_user_to_organization(organization, user, role)
+
+                if email_msg and email_subject and user.email:
+                    send_mail(
+                        email_subject,
+                        email_msg,
+                        settings.DEFAULT_FROM_EMAIL,
+                        (user.email,),
+                    )
```
FrankApiyo, line 122: Is send mail, sync or async?

FrankApiyo (approved):

onadata/libs/serializers/organization_member_serializer.py
kelvin-muchiri: @FrankApiyo It's sync, ideally it should be async, but I was indifferent in making the change at the moment. We should have `send_email_async` task should be available for the entire project.

ukanga (approved):

PR #2721, opened 2024-10-14
Customized deleted_by serializer field for Dataviews

FrankApiyo:
- Exclude `deleted_by` serializer field
- Filter out dataview that have already been deleted from the dataviews endpoint

Ran locally

N/A

  - [x] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2720

### DIFF
--- a/onadata/apps/api/viewsets/dataview_viewset.py
+++ b/onadata/apps/api/viewsets/dataview_viewset.py
@@ -159,9 +159,7 @@ class DataViewViewSet(
         """
         Remove deleted DataViews
         """
-        return super().filter_queryset(
-            queryset.filter(deleted_at=None, deleted_by=None)
-        )
+        return super().filter_queryset(queryset.filter(deleted_at=None))
 
     def list(self, request, *args, **kwargs):
         """
--- a/onadata/apps/api/viewsets/dataview_viewset.py
+++ b/onadata/apps/api/viewsets/dataview_viewset.py
@@ -126,7 +126,7 @@ class DataViewViewSet(
     A simple ViewSet for viewing and editing DataViews.
     """
 
-    queryset = DataView.objects.select_related()
+    queryset = DataView.objects.filter(deleted_at__isnull=True).select_related()
     serializer_class = DataViewSerializer
     permission_classes = [DataViewViewsetPermissions]
     lookup_field = "pk"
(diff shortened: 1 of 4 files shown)


onadata/apps/api/viewsets/dataview_viewset.py
```diff
@@ -155,6 +155,14 @@ def get_serializer_class(self):
 
         return serializer_class
 
+    def filter_queryset(self, queryset):
+        """
+        Remove deleted DataViews
+        """
+        return super().filter_queryset(
+            queryset.filter(deleted_at=None, deleted_by=None)
```
kelvin-muchiri, line 163: @FrankApiyo  Why do we need to filter by both, why not just use `deleted_at`?

onadata/libs/tests/serializers/test_dataview_serializer.py
```diff
@@ -46,9 +46,41 @@ def get_has_hxl_support_value(data_view_name):
...
+        self.assertEqual(DataView.objects.count(), 1)
+
+        # delete dataview and check dataviews endpoint
+        dataview = DataView.objects.get(name=dataview_name)
+        dataview.soft_delete(self.user)
+        serialized_dataview = DataViewSerializer(
+            dataview, context={"request": request}
+        ).data
+        self.assertFalse("deleted_by" in serialized_dataview)
+        self.assertFalse("deleted_at" in serialized_dataview)
```
ukanga, line 83: I think the `deleted_*` should not be present in the serializer. We may need to explicitly exclude the fields from the serializer. Note it will show up even for data views that have yet to be soft deleted.

kelvin-muchiri: @ukanga Looks like the fields had been initially included in the `fields` serializer but have now been removed.

kelvin-muchiri: @FrankApiyo @ukanga We do not need this test since we have filtered out deleted records and removed the `deleted_at`, `deleted_by` fields from the serializer

onadata/apps/api/viewsets/dataview_viewset.py
```diff
@@ -155,6 +155,12 @@ def get_serializer_class(self):
 
         return serializer_class
 
+    def filter_queryset(self, queryset):
+        """
+        Remove deleted DataViews
+        """
+        return super().filter_queryset(queryset.filter(deleted_at=None))
```
kelvin-muchiri, line 162: @FrankApiyo I think this is best handled when defining the `queryset` attribute [here](https://github.com/onaio/onadata/pull/2721/files#diff-fe902cbda39612b13651606735c1974aea1e9e919b9c8b0f419589ed9af07d8fR129)

kelvin-muchiri (approved):

PR #2734, opened 2024-11-13
Ensure geojson exports creation process is idempotent

### LINKED ISSUE #2735: GeoJSON export regenerated when created multiple times with the same options
### Environmental Information

- Onadata version: v4.9.1

### Problem description
GeoJSON export regenerated when created multiple times with the same options
### Expected behavior
GeoJSON export generation should be idempotent
Create GeoJSON export by linking a form to another using metadata or by using [this code snippet](https://github.com/onaio/onadata/pull/2734/files#diff-ffa1bd0426855a3433c1683df5a1bd6912e2f11518f14f66b067ccf891393c00R299-R306)
Repeat the above step in the same way
### Additional Information

FrankApiyo:
- Fix GeoJSON export exports bug
- Added a test case 
- N/A

  - [x] Included tests
  - [ ] Updated documentation

Closes #2735

### DIFF
--- a/onadata/libs/utils/api_export_tools.py
+++ b/onadata/libs/utils/api_export_tools.py
@@ -203,6 +203,7 @@ def custom_response_handler(  # noqa: C0901
                 export_type,
                 dataview_pk=dataview_pk,
                 metadata=metadata,
+                options=options,
             )
 
         if should_create_new_export(xform, export_type, options, request=request):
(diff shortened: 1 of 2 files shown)


kelvin-muchiri (approved):

onadata/libs/utils/api_export_tools.py
```diff
@@ -257,6 +264,8 @@ def _generate_new_export(  # noqa: C0901
         "host": request.get_host(),
         "sort": request.query_params.get("sort"),
     }
+    if export_options is not None:
+        options.update(export_options)
```
FrankApiyo, line 268: The reason this works is because previously when creating exports synchronously, we didn't take into consideration the options passed from the requests. 

For synchronous exports, the following `fn` is used when checking if the export has been created but it's not passed to the function that actually generates the export
```python3
parse_request_export_options({})
{'include_labels': False, 'include_reviews': False, 'include_labels_only': False, 'include_hxl': True, 'remove_group_name': False, 'group_delimiter': '/', 'split_select_multiples': True, 'include_images': True, 'win_excel_utf8': False}

```
This causes a mismatch between the generated exports and when checking if the export has been generated or not

This will fix a similar issue for entity lists because the same endpoint is used

kelvin-muchiri (approved):

PR #2748, opened 2024-12-04
Update reserved usernames

kelvin-muchiri:
### Changes

Update reserved usernames


ukanga (dismissed):

kelvin-muchiri:
This change is unnecessary as a workaround is available

PR #2755, opened 2025-01-07
CSV export optimisation

### LINKED ISSUE #2695: Optimize CSV export by improving how CSV headers are created.
The current process of CSV exports in Ona Data requires parsing all records to determine all the column headers specifically for repeat data. There is an opportunity to optimise this process such that not all records need to be parsed to determine all the required columns.

All possible columns for a form in Ona Data can be determined from reading the XForm definition of a form in XML or JSON or even from the source XLSForm. Evaluating the number of repeats whenever each submission is received is possible for repeat questions. This number can be updated if a newer, higher number is identified in an accessible manner for each repeat group. During a CSV export, there should be no need to parse through all submissions to determine the depth of a repeat group.

kelvin-muchiri:

Optimise CSV export process. Avoid traversing data to build  CSV columns for repeat responses. Instead use a register that has a record of how many columns we should have for each repeat question.

If the register for a form is unavailable, fall back to building the columns from data and trigger an asynchronous task to create the register

Fixed flaky test `onadata.apps.api.tests.viewsets.test_organization_profile_viewset.TestOrganizationProfileViewSet.test_add_members_to_owner_role`

- [x] QA

Faster CSV exports and loading of  linked datasets

  - [x] Included tests
  - [ ] Updated documentation

Closes #2695


onadata/libs/utils/logger_tools.py
```diff
@@ -1520,3 +1523,140 @@ def delete_xform_submissions(
...
+                )
+                WHERE id = %s
+                """,
+                [
+                    [repeat],
+                    repeat,
+                    incoming_max,
+                    metadata.pk,
+                ],
+            )
```
FrankApiyo, line 1592:
```suggestion
    with transaction.atomic()
       for repeat, incoming_max in repeat_max.items():
           # Get the maximum between incoming max and the current max
           # Done at database level to gurantee atomicity and
           # consistency. Avoids race conditions if it were done at the
           # application level

           with connection.cursor() as cursor:
            cursor.execute(
                """
                UPDATE main_metadata
                SET extra_data = jsonb_set(
                    COALESCE(extra_data, '{}'::jsonb),
                    %s,
                    GREATEST(
                        COALESCE((extra_data->>%s)::int, 0),
                        %s
                    )::text::jsonb,
                    true
                )
                WHERE id = %s
                """,
                [
                    [repeat],
                    repeat,
                    incoming_max,
                    metadata.pk,
                ],
            )
```

FrankApiyo: What do you think about making sure all the queries yielded by the loop go to the DB in on batch?

FrankApiyo: nvm, the `@transaction.atomic()` decorator below addresses this concern

FrankApiyo (approved):
Looks good to me!

FrankApiyo (approved):

PR #2757, opened 2025-01-08
Add GeoJSON attachment params to EXPORT_OPTIONS

### LINKED ISSUE #2752: GeoJSON export not re-generated when properties change
### Environmental Information

- Onadata version: >= 4.9.2  version <= 4.10.0 

### Problem description

When generating a GeoJSON export, the GeoJSON properties are not taken into account. Incase the properties change, a new export is not generated and the old export is used.

### Expected behavior

When GeoJSON properties are updated, a new GeoJSON export should be created.

### Additional Information

The bug is a side effect of this [change](https://github.com/onaio/onadata/pull/2734). Generally, a new export is only generated if a new submission is received or if an existing submission is edited. However, for GeoJSON exports, we should also generate a new export if the properties change

FrankApiyo:
- Add new `EXPORT_OPTIONS` for GeoJSON exports

- Tested locally
- Updated test cases

N/A

  - [x] Included tests
  - [ ] Updated documentation

Closes #2752

### DIFF
--- a/onadata/libs/tests/utils/test_export_tools.py
+++ b/onadata/libs/tests/utils/test_export_tools.py
@@ -283,12 +283,11 @@ class TestExportTools(TestAbstractViewSet):
         metadata = MetaData.objects.create(
             content_type=ContentType.objects.get_for_model(XForm),
             data_type="media",
-            data_value=f"xform_geojson {self.xform.id} testgeojson",
+            data_value=f"xform_geojson {self.xform.id} testgeojson2",
             extra_data={
                 "data_title": "start",
                 "data_fields": "",
                 "data_geo_field": "qn09",
-                "data_simple_style": True,
             },
             object_id=self.xform.id,
         )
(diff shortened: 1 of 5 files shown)


onadata/libs/utils/export_tools.py
```diff
@@ -641,6 +641,31 @@ def generate_kml_export(
...
 
+def get_query_params_from_metadata(metadata):
+    """
+    Build out query params to be used in GeoJsonSerializer
+    """
+    if metadata is None:
+        return None
+
+    extra_data = metadata.extra_data
+    if extra_data is None:
```
kelvin-muchiri, line 652: We can use on if statement for line 648 and line 652

onadata/libs/tests/utils/test_export_tools.py
```diff
@@ -349,6 +401,50 @@ def test_should_not_create_new_export_when_old_exists(self):
             Export.objects.get(xform=self.xform).options,
         )
 
+        # New metadata will yield a new export
+        metadata.delete()
```
kelvin-muchiri, line 405: @FrankApiyo Will updating `metadata.extra_data` suffice instead of deleting and creating a new one?

FrankApiyo: Whenever you create a new linked dataset, a new metadata object is created; So creating a new metadata object more closely simulates what happens in production. (as compared to editing an existing metadata object)

kelvin-muchiri: @FrankApiyo Noted. Though l believe since this is a unit test and not an integration test, we are only interested in the `metadata` passed to `custom_response_handler` and not creation of the linked dataset.

onadata/libs/tests/utils/test_export_tools.py
```diff
@@ -349,6 +403,49 @@ def test_should_not_create_new_export_when_old_exists(self):
...
+            data_value=f"xform_geojson {self.xform.id} testgeojson2",
+            extra_data={
+                "data_title": "end",
+                "data_fields": "",
+                "data_geo_field": "qn09",
+                "data_simple_style": True,
+            },
+            object_id=self.xform.id,
+        )
+        _response = custom_response_handler(
```
kelvin-muchiri, line 419: Looks like the variable `_response` is unused

kelvin-muchiri: In that case, the convention is usually to use an `_` underscore as the variable name. Though in this case I think we can just ignore the result from the function call and not assign it to a variable

kelvin-muchiri (approved):

PR #2788, opened 2025-03-31
feat: Create index for user_auth

FrankApiyo:


ukanga (approved):

kelvin-muchiri:
In the past, we've been forced to create indices concurrently on tables that have data otherwise the migration might be slow

PR #2791, opened 2025-04-07
Add new can view and can view minor roles

FrankApiyo:
- Add new can view and can view minor roles

Added test cases

  - [x] Included tests
  - [x] Updated documentation

**Todo**
- [ ] Add test cases for new meta permissions

### DIFF
--- a/onadata/apps/logger/models/xform.py
+++ b/onadata/apps/logger/models/xform.py
@@ -20,7 +20,7 @@ from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import models, transaction
 from django.db.models import Sum
-from django.db.models.signals import post_delete, pre_save
+from django.db.models.signals import post_delete, pre_save, post_save
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.html import conditional_escape
@@ -1413,6 +1413,27 @@ def save_project(sender, instance=None, created=False, **kwargs):
 
 
 pre_save.connect(save_project, sender=XForm, dispatch_uid="save_project_xform")
+
+
+def _create_meta_perms(sender, instance, created, **kwargs):
+    metadata = instance.metadata_set.all().filter(data_type="xform_meta_perms").first()
+    if instance.deleted_by is None and metadata is None:
+        from onadata.libs.serializers.metadata_serializer import MetaDataSerializer
+
+        serializer = MetaDataSerializer(
+            data={
+                "data_type": "xform_meta_perms",
+                "data_value": "editor-minor|dataentry-only|readonly-no-download",
+                "xform": instance.pk,
+            }
+        )
+        if serializer.is_valid():
+            serializer.save()
+
+
+post_save.connect(
+    _create_meta_perms, sender=XForm, dispatch_uid="create_xform_meta_permissions"
+)
 
 
 # pylint: disable=unused-argument
--- a/onadata/apps/logger/models/xform.py
+++ b/onadata/apps/logger/models/xform.py
@@ -1415,6 +1415,7 @@ def save_project(sender, instance=None, created=False, **kwargs):
 pre_save.connect(save_project, sender=XForm, dispatch_uid="save_project_xform")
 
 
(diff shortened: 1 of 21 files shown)


onadata/apps/logger/models/xform.py
```diff
@@ -1415,6 +1415,27 @@ def save_project(sender, instance=None, created=False, **kwargs):
 pre_save.connect(save_project, sender=XForm, dispatch_uid="save_project_xform")
 
 
+def _create_meta_perms(sender, instance, created, **kwargs):
+    metadata = instance.metadata_set.all().filter(data_type="xform_meta_perms").first()
```
kelvin-muchiri, line 1420: `.all()` is unnecessary when you have `.filter`

onadata/apps/logger/models/xform.py
```diff
@@ -1415,6 +1415,27 @@ def save_project(sender, instance=None, created=False, **kwargs):
...
+    if instance.deleted_by is None and metadata is None:
```
kelvin-muchiri, line 1421: Should we care whether the object is deleted or not? I think we can omit the check.

onadata/apps/logger/models/xform.py
```diff
@@ -1415,6 +1415,28 @@ def save_project(sender, instance=None, created=False, **kwargs):
 pre_save.connect(save_project, sender=XForm, dispatch_uid="save_project_xform")
 
 
+# pylint: disable=unused-argument,import-outside-toplevel
+def _create_meta_perms(sender, instance, created, **kwargs):
+    metadata = instance.metadata_set.all().filter(data_type="xform_meta_perms").first()
+    if instance.deleted_by is None and metadata is None:
+        from onadata.libs.serializers.metadata_serializer import MetaDataSerializer
+
+        serializer = MetaDataSerializer(
```
kelvin-muchiri, line 1424: In the past, how have created `xform_meta_perms`? It's best to use the model here to created the object directly instead of using the serializer

onadata/libs/permissions.py
```diff
@@ -251,6 +261,33 @@ class DataEntryRole(Role):
     }
 
 
+class EditorNoDownload(Role):
+    """
+    Editor-Minor Role class - user can submit data, read and edit only the data
```
kelvin-muchiri, line 266: Editor-Minor Role class does not match the class name, either way, I think we can omit it the ` Editor-Minor Role class -` prefix altogether

onadata/apps/logger/models/xform.py
FrankApiyo: the `xform_meta_perms` , in the past, has been created by making API requests from the front-end. So this record is normally not created on the back-end service

FrankApiyo: We could use the `MetaData.objects.create` method directly. I think if we use the serializer this way, we will benefit from some validation and logic already in the `create` method of the serializer

onadata/libs/serializers/metadata_serializer.py
```diff
@@ -326,6 +337,11 @@ def update(self, instance, validated_data):
         instance = super().update(instance, validated_data)
 
         if instance.data_type == XFORM_META_PERMS:
-            update_role_by_meta_xform_perms(instance.content_object)
+            xform = instance.content_object
+            update_role_by_meta_xform_perms(xform)
+            safe_delete(f"{PROJ_OWNER_CACHE}{xform.project.pk}")
```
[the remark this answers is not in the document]
FrankApiyo: moved!

onadata/apps/logger/models/xform.py
kelvin-muchiri: @FrankApiyo Noted, though since it's being created directly and not a request from a client, we are assured the values are always correct. But if there is some essential checks that we think applies, it should be moved to a some utiility module accessible by the serializer module and model module

FrankApiyo: I'll move this into a different PR

kelvin-muchiri (approved):

PR #2798, opened 2025-04-21
fix: PyXForm question may have choices as None

ukanga:

Ensure choices is not None and has the attribute options.

Publish a form with a select_multiple_from_file question on onadata versions 5.0.0 - 5.0.3. Export will fail with an AttributeError exception.

None.

  - [x] Included tests
  - [ ] Updated documentation

Closes https://github.com/onaio/onadata/issues/2797

### DIFF
--- a/onadata/libs/tests/utils/test_export_builder.py
+++ b/onadata/libs/tests/utils/test_export_builder.py
@@ -2723,8 +2723,41 @@ class TestExportBuilder(TestBase):
             CSVDataFrameBuilder._collect_select_multiples(dd), select_multiples
         )
 
+    def test_select_multiple_from_file_choices(self):
+        md_xform = """
+        | survey  |
+        |         | type                                 | name  | label  |
+        |         | text                                 | name  | Name   |
+        |         | integer                              | age   | Age    |
+        |         | select_multiple_from_file fruits.csv | fruit | Fruit  |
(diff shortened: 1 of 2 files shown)


onadata/libs/tests/utils/test_export_builder.py
```diff
@@ -2723,8 +2723,41 @@ def test_select_multiples_choices(self):
...
+        ][0]
+        # pylint: disable=protected-access
+        choices = export_builder._get_select_mulitples_choices(
+            child, dd, ExportBuilder.GROUP_DELIMITER, ExportBuilder.TRUNCATE_GROUP_TITLE
+        )
+        self.assertEqual(choices, [])
+
     # pylint: disable=invalid-name
     def test_select_multiples_choices_with_choice_filter(self):
+        """Test select multiples from file choices."""
```
kelvin-muchiri, line 2760: What's the difference between this test and the newly added one?

ukanga: @kelvin-muchiri The new one has no choice filter. The old one has a choice filter, check the choice_filter.xlsx file to compare the structure. The presence of the choice_filter results in a different data structure that does not land in the code area where the changes are made.

kelvin-muchiri (approved):

PR #2815, opened 2025-05-06
Restrict KMS active key info to organization admins

### LINKED ISSUE #2810: Limit organization active key details only to admins

kelvin-muchiri:

Restrict KMS active key info to organization admins

  - [X] Included tests
  - [ ] Updated documentation

Closes #2810

### DIFF
--- a/onadata/apps/api/viewsets/organization_profile_viewset.py
+++ b/onadata/apps/api/viewsets/organization_profile_viewset.py
@@ -83,20 +83,21 @@ class OrganizationProfileViewSet(
 
     def retrieve(self, request, *args, **kwargs):
         """Get organization from cache or db"""
-        cache_key = get_org_profile_cache_key(request.user, self.get_object())
+        organization = self.get_object()
+        cache_key = get_org_profile_cache_key(request.user, organization)
         cached_org = cache.get(cache_key)
 
         if cached_org:
             return Response(cached_org)
 
         is_admin = OwnerRole.user_has_role(
-            request.user, self.get_object()
-        ) or ManagerRole.user_has_role(request.user, self.get_object())
+            request.user, organization
+        ) or ManagerRole.user_has_role(request.user, organization)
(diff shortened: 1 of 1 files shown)


onadata/libs/serializers/organization_serializer.py
```diff
@@ -85,6 +85,16 @@ def __init__(self, *args, **kwargs):
                 for field in getattr(self.Meta, "owner_only_fields"):
                     self.fields.pop(field)
 
+    def to_representation(self, instance):
+        """Override `to_representation`."""
+        data = super().to_representation(instance)
+
+        if not self._is_admin_or_manager(self.context["request"].user, instance):
+            data.pop("active_kms_key")
+            data.pop("inactive_kms_keys")
```
ukanga, line 94: Can we push this to the serializer methods for the relevant fields? Or perhaps have a different serializer for managers that includes these fields so that we don't end up with pop()?

onadata/apps/api/viewsets/organization_profile_viewset.py
```diff
@@ -83,11 +85,22 @@ def retrieve(self, request, *args, **kwargs):
...
         cache_key = get_org_profile_cache_key(request.user, self.get_object())
         cached_org = cache.get(cache_key)
+
         if cached_org:
             return Response(cached_org)
-        response = super().retrieve(request, *args, **kwargs)
-        cache.set(cache_key, response.data)
-        return response
+
+        is_admin = OwnerRole.user_has_role(
```
ukanga, line 93: This should probably be in the override function `get_serializer_class()`. Can we not access the user who has made the request there?

kelvin-muchiri: @ukanga We can't use `self.get_object` inside `get_serializer_class` as it normally results into an infinite recursion

ukanga (approved):

PR #2820, opened 2025-05-09
feat: Update XForm meta-perms post-save signal

FrankApiyo:


onadata/apps/viewer/models/data_dictionary.py
```diff
@@ -257,6 +257,31 @@ def set_object_permissions(sender, instance=None, created=False, **kwargs):
 )
 
 
+# pylint: disable=unused-argument,import-outside-toplevel
+def _create_meta_perms(sender, instance, created, **kwargs):
+    meta_perms_exists = instance.metadata_set.filter(
+        data_type="xform_meta_perms"
+    ).exists()
+
+    xform = XForm.objects.get(pk=instance.pk)
```
kelvin-muchiri, line 266: Can we move this into the if block? That way we avoid getting an object which we might not use

FrankApiyo: done!

kelvin-muchiri (approved):

PR #2860, opened 2025-07-15
Fix submission marked as duplicate if IntegrityError is thrown

kelvin-muchiri:

Fix submission marked as duplicate if `django.db.IntegrityError` is thrown when saving the submission. The 

- [ ] QA

`django.db.IntegrityError` will have a HTTP response code of `500` and not `202`

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/libs/utils/logger_tools.py
+++ b/onadata/libs/utils/logger_tools.py
@@ -594,36 +594,21 @@ def create_instance(
 
         return DuplicateInstance()
 
-    try:
-        with transaction.atomic():
-            if isinstance(xml, bytes):
-                xml = xml.decode("utf-8")
-            instance = save_submission(
-                xform,
-                xml,
-                media_files,
-                new_uuid,
-                submitted_by,
-                status,
-                date_created_override,
-                checksum,
-                request,
-            )
-    except IntegrityError:
-        instance = get_first_record(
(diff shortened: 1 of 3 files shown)


onadata/libs/utils/logger_tools.py
```diff
@@ -594,36 +594,21 @@ def create_instance(
...
 
-        if instance:
-            attachment_names = [
-                a.media_file.name.split("/")[-1]
-                for a in Attachment.objects.filter(instance=instance)
-            ]
-            media_files = [f for f in media_files if f.name not in attachment_names]
-            save_attachments(xform, instance, media_files)
-            instance.save()
-
```
ukanga, line 625: Is saving attachments no longer necessary?

kelvin-muchiri: After testing and carefully looking at the code, in case of a duplicate submission, the code execution will never reach that code since the attachment processing will be handled in the `if existing_instance and (new_uuid or existing_instance.xform.has_start_time)` block, a few lines above the block I removed.

I then updated [Attachment.objects.get_or_create inside save_attachments method](https://github.com/onaio/onadata/pull/2860/files#diff-76a3d3d03f02ac0438c8082454afcf25cea152b40d7430deb3130c7b00d6023dR444) and excluded the file object from the uniqueness check since it's guaranteed the file object references will always be different

PR #2877, opened 2025-07-28
Remove property users, forms from list of projects

kelvin-muchiri:

Properties `users` and `forms` make fetching a list of projects slow. For performance optimization, these properties will only be available when a project is retrieved via endpoint /api/v1/projects/<id>

- [ ] QA

Faster responses when fetching list of projects

  - [ ] Included tests
  - [ ] Updated documentation

Closes #


FrankApiyo:
@kelvin-muchiri, can you add a field that indicates the number of forms in a project? Also, would it be possible to check the role of the `request.user` in each project?

ukanga (approved):

PR #2879, opened 2025-07-31
Make duplicate submission check efficient

kelvin-muchiri:
### Changes/Features implemented

- Make duplicate submission check efficient
- Fix `Attachment.MultipleObjectsReturned` if the number of duplicate attachments for a duplicate submission found is greater than 1. A bug in versions older than [v5.2.0](https://github.com/onaio/onadata/releases/tag/v5.2.0) allowed the same attachment to be saved more than once for duplicate `Instance`s. In pre `v5.2.0`, the `TemporaryUploadedFile` file object was included in a [get_or_create](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#get-or-create) lookup to be compared with the value of the `media_file` field. `TemporaryUploadedFile` is not comparable to the value of the `media_file` field stored in the DB; hence, the comparison would always fail, resulting in the creation of a new `Attachment`

Efficient handling of duplicate submissions

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/libs/utils/logger_tools.py
+++ b/onadata/libs/utils/logger_tools.py
@@ -437,9 +437,9 @@ def save_attachments(xform, instance, media_files, remove_deleted_media=False):
                     defaults={"media_file": f},
                 )
             except Attachment.MultipleObjectsReturned:
-                # In pre v5.2.0, we had a bug where multiple attachments with the
-                # same name could be created. This is a workaround to avoid
-                # the issue.
+                # In pre v5.2.0, we had a bug where duplicate attachments could
+                # be created. This is a workaround to avoid raising an error for
+                # already existing multiple duplicates.
                 pass
 
     if remove_deleted_media:


FrankApiyo (dismissed):

onadata/libs/utils/logger_tools.py
```diff
@@ -549,17 +540,19 @@ def create_instance(
...
     new_uuid = get_uuid_from_xml(xml)
-    filtered_instances = get_filtered_instances(uuid=new_uuid, xform_id=xform.pk)
-    existing_instance = get_first_record(filtered_instances.only("id"))
-    if existing_instance and (new_uuid or existing_instance.xform.has_start_time):
+    duplicate_exists = Instance.objects.filter(
+        xform_id=xform.pk, uuid=new_uuid
+    ).exists()
+
+    if duplicate_exists:
+        duplicate_instance = Instance.objects.get(xform_id=xform.pk, uuid=new_uuid)
```
ukanga, line 549: Can we switch to a try ..except .. else  block and avoid the [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bug?

ukanga (changes requested):

PR #2928, opened 2025-10-16
feat: import entities via CSV

kelvin-muchiri:

Add ability to import entities from CSV

- [x] QA

No side effects

  - [x] Included tests
  - [x] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_entity_list_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_entity_list_viewset.py
@@ -1890,7 +1890,7 @@ class ImportEntitiesTestCase(TestAbstractViewSet):
         request = self.factory.post("/")
         response = self.view(request, pk=self.entity_list.pk)
         self.assertEqual(response.status_code, 401)
-        # Anonymous user cannot into import public EntityList
+        # Anonymous user cannot import into public EntityList
         self.project.shared = True
         self.project.save()
         request = self.factory.post("/")


docs/entities.rst
```diff
@@ -596,3 +596,169 @@ The endpoint is used to delete a single Entity or multiple Entities. The IDs of
 **Response**
 
 Status: ``204 No Content``
+
+Import Entities from CSV
+-------------------------
+
+.. raw:: html
+
+	   <pre class="prettyprint"><b>POST</b> /api/v2/entity-lists/&lt;entity_list_id&gt;/import-entities</pre>
```
ukanga: Since we have entities under the path `/entity-list/<pk>/entities` should we allow import on the same endpoint? If not, perhaps consider `/entity-list/<pk>/entities-import` instead.

ukanga (commented):
Is there any other column required if you have a csv with only the label column? I would expect that the uuid in this case will be generated for each row on the server side.

```CSV
"label"
"Apple"
"Melon"
"Peach"
```
Should this work or what is missing?

The error below is the current result.

```json
{
    "result": {
        "created": 0,
        "errors": [
            [
                2,
                "{'data': [ErrorDetail(string='This field may not be empty.', code='invalid')]}"
            ],
            [
                3,
                "{'data': [ErrorDetail(string='This field may not be empty.', code='invalid')]}"
            ],
```

docs/entities.rst
ukanga: I would expect that it will pick properties from the CSV file being uploaded. Is there an extra step that was skipped API-wise? Is that documented?

kelvin-muchiri: @ukanga Yes properties are picked from the CSV uploaded. Which CSV did you use? The properties must exist prior

kelvin-muchiri: @ukanga You might have missed to create the properties by uploading a registration form

kelvin-muchiri: This [PR adds a properties endpoint](https://github.com/onaio/onadata/pull/2945)

docs/entities.rst
FrankApiyo:
```suggestion
- ``Entity`` - Each item that gets managed by an ODK workflow. Entities are automatically created from submissions received from a form that contains entity definitions.
```

FrankApiyo:
```suggestion
The following endpoints provides access to Entities related data:
```

onadata/apps/api/tests/viewsets/test_entity_list_viewset.py
```diff
@@ -1814,3 +1825,243 @@ def test_download_from_s3(self):
...
+        )
+        self.assertIsInstance(mock_storage_save.call_args.args[1], InMemoryUploadedFile)
+
+    def test_anonymous_user(self):
+        """Authentication is required to import entities"""
+        # Anonymous user cannot import into private EntityList
+        request = self.factory.post("/")
+        response = self.view(request, pk=self.entity_list.pk)
+        self.assertEqual(response.status_code, 401)
+        # Anonymous user cannot into import public EntityList
```
FrankApiyo:
```suggestion
        # Anonymous user cannot import into public EntityList
```

Issue #2803, opened 2025-04-23
Endpoints that query the user model should not be case sensitive


FrankApiyo:
### Environmental Information

- Onadata version: 5.0.3

### Problem description
Endpoints that query the user model using the username are currently case sensitive.
Examples are as follows:
- `/api/v1/profiles`
- `/api/v1/projects/10574/share`

### Expected behavior
We should be able to query for usernames in a case-insensitive manner

### Steps to reproduce the behavior
The following commands should yield the same result:
```bash
➜  curl 'scheme://api-domain/api/v1/profiles.json?users=fapiyo' \
  -H 'accept: application/json' \
  -H 'authorization: TempToken <Your temp token>'
➜  curl 'scheme://api-domain/api/v1/profiles.json?users=FApiyo' \
  -H 'accept: application/json' \
  -H 'authorization: TempToken <Your temp token>'
```

```bash
➜  curl 'scheme://api-domain/api/v1/projects/<project-id>/share' \
  -X 'PUT' \
  -H 'accept: application/json' \
  -H 'authorization: TempToken <Your temp token>' \
  --data-raw 'username=fapiyo&role=readonly'
➜  curl 'scheme://api-domain/api/v1/projects/<project-id>/share' \
  -X 'PUT' \
  -H 'accept: application/json' \
  -H 'authorization: TempToken <Your temp token>' \
  --data-raw 'username=FApiyo&role=readonly'
```

FrankApiyo:
A potential solution is to use citext: https://www.postgresql.org/docs/current/citext.html

FrankApiyo:
https://adamj.eu/tech/2023/02/23/migrate-django-postgresql-ci-fields-case-insensitive-collation/

FrankApiyo:
This issue also affects 3rd party packages interacting with Onadata

FrankApiyo:
We would also like both the email and username fields to be case insensitive

PR #2966, opened 2025-11-29
Added citext migration to make username lookups case-insensitive (#2803)

jestrada51:

Added a database migration that enables the PostgreSQL citext extension (if missing) and converts auth_user username field from varying to citext so username lookups within the API endpoints become case-insensitive.

Added a regression test (TestCaseInsensitiveUserQuery) that confirms /api/v1/profiles.json?users=<username> correctly matches users regardless of casing (bob, Bob, bOb, etc.). 

Manual verification of the migration was done to ensure that citext is included as an extension.
Manual verification of the API endpoints in curl was done in accordance to the example in the original #2803 post to ensure correct behavior
Test case was ran with different variants of the target username in order to make sure applying citext leads for case insensitivity.

Api endpoints such as:
```
    /api/v1/profiles.json?users=<username>
```
and 
```
 /api/v1/profiles/<username>/
```
are now case insensitive

A potential side effect is although citext was mentioned as a potential solution in the #2803 post, Changing the username column from varchar to citext may affect any raw SQL or third-party integrations that implicitly rely on the old column type.   

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_case_insensitive_username_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_case_insensitive_username_viewset.py
@@ -1,0 +1,87 @@
+from django.test import TestCase
+from django.contrib.auth import get_user_model
+from django.db import connection
+from rest_framework.test import APIRequestFactory, force_authenticate
+from onadata.apps.api.viewsets.user_profile_viewset import UserProfileViewSet
+from onadata.apps.main.models import UserProfile
+
+
+# imports the model which we are testing which is user
+User = get_user_model()
+
+class TestCaseInsensitiveUserQuery(TestCase):
+
+
+    ''' 
+        Method runs sql to create citext within test database if the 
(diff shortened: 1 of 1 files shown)


jestrada51:
Hey @FrankApiyo and @kelvin-muchiri. 

I looked into collations over citext and have updated the migration to use the case-insensitive collation as recommended.
However, during migration, I encountered the following error:

django.db.utils.NotSupportedError: nondeterministic collations are not supported for operator class "varchar_pattern_ops"

After doing some research, I found out that this index ""auth_user_username_6821ab7c_like" is clashing with deterministic = false in the created collation which is needed to allow for case insensitivity in the username field. 

I wanted to know if removing this specific index for the sake of having the collation run is a good idea and would not break anything important. I tried it on my local machine and it achieved case insensitivity for the mentioned api endpoints for #2803.

Let me know.

PR #3013, opened 2026-02-10
Switch to Python 3.13 from 3.10.19

ukanga:

- Switched to Python 3.13 from 3.10.19
- Switched to dhi.io hardened images 
- Updated to the latest pinned requirements
- Trivy HTML direct export instead of conversion
- Use GitHub Token instead of SSH private key setup


setup.cfg
```diff
@@ -66,7 +66,7 @@ install_requires =
     pylibmc
     python-memcached
     # XLSForm support
-    pyxform==4.1.0
+    pyxform
```
kelvin-muchiri: Do we not need to point to an exact version anymore?

ukanga: this is now in .pip file

FrankApiyo (approved):

PR #3021, opened 2026-02-20
fix: incorrect cors middlware placement

kelvin-muchiri:

CorsMiddleware should be placed as high as possible, especially before any middleware that can generate responses such as Django’s CommonMiddleware or Whitenoise’s WhiteNoiseMiddleware. If it is not before, it will not be able to add the CORS headers to these responses. Reference: https://pypi.org/project/django-cors-headers/

One side effect of this bug was that `corsheaders`  settings were not being applied e.g `CORS_ORIGIN_ALLOW_ALL=True` wasn't working for local development flow

### DIFF
--- a/onadata/settings/common.py
+++ b/onadata/settings/common.py
@@ -161,9 +161,9 @@ TEMPLATES = [
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 MIDDLEWARE = (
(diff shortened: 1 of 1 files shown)


FrankApiyo:
@kelvin-muchiri, why not have it as the first middleware class in the list?

FrankApiyo (approved):

ukanga (approved):

PR #3066, opened 2026-04-21
WIP: add bbox endpoint for forms, dataviews, and merged datasets

FrankApiyo:
## Summary

Adds `GET /api/v1/forms/{id}/bbox`, `/api/v1/dataviews/{id}/bbox`, and `/api/v1/merged-datasets/{id}/bbox`.

Each returns `{"bbox": [min_lng, min_lat, max_lng, max_lat] | null}` — the extent of geolocated submissions (null when the target has no geolocated rows, so callers can fall back to a default view cleanly).

## Motivation

The frontend's tile-based map view (Martin + `form_tiles()`) cannot compute bounds client-side since features are paged per MVT tile. A dedicated endpoint lets the frontend fit the viewport on first load without fetching the full submission set.

## Design

- `onadata/libs/utils/bbox_tools.py` — shared `compute_instance_bbox(xform_ids, dataview=None)` helper using Django GIS `Extent` aggregate
- Filter shape mirrors the `form_tiles()` PostGIS function (non-deleted, geom non-null, optional dataview query filter) so the fit matches the data Martin serves
- Three viewset `@action` methods (`XFormViewSet`, `DataViewViewSet`, `MergedXFormViewSet`) wrap the helper; permissions inherit from each viewset's existing class

## Test plan

- [x] `test_bbox_returns_extent_of_geolocated_submissions` — XForm with the gps fixture returns a valid 4-tuple bbox
- [x] `test_bbox_null_when_no_geolocated_submissions` — returns `{"bbox": null}` for forms without geoms
- [x] DataView + MergedXForm bbox actions return the correct shape (null-case covered; the helper is exercised via the XForm extent test)
- [ ] Run full onadata test suite in CI
- [ ] Verify against a live form on stage once deployed

## Consumers

- Frontend tile map view (`TileMapView`) → `useFormBbox` hook → `fitBounds` on map load (frontend integration in progress).

## Why WIP

Marking WIP pending:
- Discussion on whether the `bbox_tools` helper should live in `libs/utils/` or somewhere closer to the logger models
- Decision on whether to cache the endpoint (bbox changes slowly; 5-min Redis cache would be cheap)
- Review of whether the null-return for empty datasets is preferable to 404 (I went with null to make the frontend branch simpler)

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_dataview_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_dataview_viewset.py
@@ -2062,3 +2062,18 @@ class TestDataViewViewSet(TestAbstractViewSet):
             self.assertEqual(
                 csv_file.read(), "name,age,gender\nDennis Wambua,28,male\n"
             )
+
+    def test_bbox_returns_null_without_geolocated_submissions(self):
+        """The tutorial fixture has no geom, so bbox is ``null``.
+
+        Verifies the endpoint is wired and returns the expected envelope; the
+        geolocated-extent path is covered by the XFormViewSet bbox test which
+        shares the same underlying helper.
+        """
+        self._create_dataview()
+        view = DataViewViewSet.as_view({"get": "bbox"})
+        request = self.factory.get("/", **self.extra)
+        response = view(request, pk=self.data_view.pk)
+
+        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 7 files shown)


FrankApiyo:
Superseded by #3072 — rebased onto main and renamed branch from `wip/bbox-endpoint` to `bbox-endpoint`.

PR #3067, opened 2026-04-22
fix: disable Google OAuth PKCE code verifier auto-generation

kelvin-muchiri:
### Changes / Features implemented

Google's confidential web-client OAuth flow does not require PKCE, and this codebase does not persist a code_verifier across the redirect, so an auto-generated verifier causes token exchange to fail. Explicitly pass autogenerate_code_verifier=False to lock in the pre-upgrade behavior regardless of upstream google-auth-oauthlib defaults.

### Steps taken to verify this change does what is intended

- [x] QA

### Side effects of implementing this change

**Before submitting this PR for review, please make sure you have:**

  - [x] Included tests
  - [ ] Updated documentation

Closes #3068

PR #3069, opened 2026-04-22
fix: disable Google OAuth PKCE code verifier auto-generation

kelvin-muchiri:
Google's confidential web-client OAuth flow does not require PKCE, and this codebase does not persist a code_verifier across the redirect, so an auto-generated verifier causes token exchange to fail. Explicitly pass autogenerate_code_verifier=False to lock in the pre-upgrade behavior regardless of upstream google-auth-oauthlib defaults.

  - [ ] Included tests
  - [ ] Updated documentation

Closes #


kelvin-muchiri:
This is the canonical PR https://github.com/onaio/onadata/pull/3067

PR #3073, opened 2026-04-22
fix (xform): update XForm schema after PyXForm v4 upgrade

kelvin-muchiri:

- Added logger migration `0038_update_xform_schema_after_pyxform_v4_upgrade` to repair stored XForm JSON after the PyXForm v4 upgrade. Mirrors `0031_update_schema_after_pyxform_upgrade` (helpers are intentionally duplicated — migrations must stay self-contained snapshots and cannot safely import helpers that may be refactored later).
- Handled [XLSForm/pyxform#752](https://github.com/XLSForm/pyxform/issues/752): `OsmUploadQuestion.to_json_dict()` raises `AttributeError: 'OsmUploadQuestion' object has no attribute 'choices'` under pyxform 4.x (still unfixed upstream as of v4.4.1). Added `AttributeError` to the inner `except` so XForms containing OSM upload questions fall back to the in-place JSON patch path instead of aborting the whole migration.

- [ ] QA

- XForm schemas updated to match PyXForm v4
-


kelvin-muchiri:
Closing this as no new changes were introduced in v4 that affect existing schemas

PR #3084, opened 2026-05-04
fix(docker): include /usr/bin/date in runtime image

### LINKED ISSUE #3083: Runtime Docker image is missing /usr/bin/date, breaking cronjob shell scripts
### Environmental Information

- Onadata version: v5.15.3

### Problem description

Shell scripts running inside the runtime container fail with:

```
/path/to/script.sh: line N: date: command not found
```

The runtime stage of `docker/onadata-uwsgi/Dockerfile.ubuntu` is built `FROM dhi.io/python:3.13-debian13`, a hardened/minimal image that ships almost no userland tools. The Dockerfile then explicitly cherry-picks the binaries the runtime needs from the build stage:

```dockerfile
COPY --from=base /usr/bin/env /usr/bin/env
COPY --from=base /usr/bin/bash /usr/bin/bash
COPY --from=base /usr/bin/bash /bin/bash
COPY --from=base /usr/bin/bash /bin/sh
...
COPY --from=base /usr/bin/java /usr/bin/java
```

`/usr/bin/date` is not in that list, so any shell script that invokes `date` (e.g. cronjob scripts that compute "previous month" or a timestamp suffix using `date --date="-1 month"` / `date +"%Y-%m-%d"`) fails immediately on the `date` line.

### Expected behavior

Shell scripts running in the runtime container can use `date` from GNU coreutils, the same way they can use `bash` and `env`.

1. Build / pull `onaio/onadata:v5.15.3` (or any image built from the current Dockerfile.ubuntu).
2. Run any shell command that calls `date`:
   ```
   docker run --rm onaio/onadata:v5.15.3 bash -c 'date'
   ```
3. Observe `bash: date: command not found`.

### Additional Information

Fix is to copy `/usr/bin/date` from the `base` stage into the runtime stage alongside the other coreutils-style binaries. PR to follow.

kelvin-muchiri:

Copy `/usr/bin/date` from the `base` build stage into the runtime stage of `docker/onadata-uwsgi/Dockerfile.ubuntu`, alongside the existing `env`, `bash`, and `java` binaries.

The runtime image is `FROM dhi.io/python:3.13-debian13`, a hardened/minimal image that ships almost no userland tools, so any binary the runtime needs has to be cherry-picked from the build stage. `date` was previously omitted, which broke shell scripts that call it (e.g. cronjob scripts computing "previous month" / a timestamp suffix with `date --date="-1 month"` or `date +"%Y-%m-%d"`).

- `docker build -f docker/onadata-uwsgi/Dockerfile.ubuntu -t onadata:test .`
- `docker run --rm onadata:test bash -c 'date'` — succeeds and prints the current date.
- `docker run --rm onadata:test bash -c 'date --date="-1 month" +"%Y-%m"'` — succeeds and prints last month.
- Existing `bash`, `env`, and `java` invocations still work.

- Adds one extra binary (~120 KB) to the runtime image. No new shared libraries — `date` only depends on `libc`, which is already pulled in by the existing `/usr/lib/` copy.
- Slightly broadens the runtime tool surface beyond the minimum, but `date` is a benign, read-only coreutil with no network or write side effects.

  - [ ] Included tests
  - [ ] Updated documentation

(No tests added — Dockerfile-only change; behavior is verified by running `date` in a built image.)

Closes #3083

### DIFF
--- a/docker/onadata-uwsgi/Dockerfile.ubuntu
+++ b/docker/onadata-uwsgi/Dockerfile.ubuntu
@@ -92,6 +92,7 @@ COPY --from=base /usr/bin/env /usr/bin/env
 COPY --from=base /usr/bin/bash /usr/bin/bash
 COPY --from=base /usr/bin/bash /bin/bash
 COPY --from=base /usr/bin/bash /bin/sh
+COPY --from=base /usr/bin/date /usr/bin/date
 
 # Java is required by pyxform's ODK Validate for XForm validation.
 # The JRE lives under /usr/lib/jvm/ (already covered by /usr/lib/ copy above)


FrankApiyo (approved):

kelvin-muchiri:
Closing this as at the moment we can use python3 for date in cron jobScripts

PR #3091, opened 2026-05-26
fix: resolve email/username login to the correct account

FrankApiyo:
## Problem

`onadata.apps.main.backends.ModelBackend` lets users log in with their **email**
in place of their username. But `email` is **not** unique on the user model, and
the backend matched candidates with a single `.first()`:

```python
user = User.objects.filter(
    Q(username__iexact=username) | Q(email__iexact=username)
).first()
if user and user.check_password(password):
    ...
```

When two accounts share an email, logging in by email would:

1. resolve to an **arbitrary (wrong) account** (lowest PK) if its password
   happened to match, or
2. **reject valid credentials** belonging to a non-first duplicate — the backend
   never tried the second matching account.

This also poisons downstream identity: OnaData brokers into Keycloak, where
`preferred_username` is set from the authenticated account's username, so the
wrong selection here propagates through SSO.

## Change

- Try **username matches first** (username is unique → typing your own username
  always resolves to your account).
- Resolve **email** collisions **by password** — iterate the candidates and
  return whichever one's password verifies.
- The organization-account login block (`is_organization_user`) is preserved on
  the selected account.

## Test plan

- [x] New `onadata/apps/main/tests/test_backends.py`: username login,
      case-insensitivity, email login, shared-email-resolves-by-password,
      non-first duplicate not blocked, and the organization-account block.
- [x] Verified backend logic against the real module in an isolated harness
      (9/9, including the org-account block).
- [x] black (24.8.0) / isort / flake8 clean.

> Note: email-collision resolution is by password, so it cannot disambiguate two
> accounts sharing **both** email and password (vanishingly unlikely; either is
> equally valid). Deduplicating shared-email accounts remains the durable fix.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,6 +1,2 @@
 exclude_paths:
   - ".claude/skills/**"
-  # Literal credentials in these auth backend tests trip Bandit's
-  # "possible hardcoded password" (B105/B106); they're test fixtures,
-  # not secrets, so exclude the file from analysis.
-  - "onadata/apps/main/tests/test_backends.py"
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,2 +1,6 @@
 exclude_paths:
   - ".claude/skills/**"
+  # Literal credentials in these auth backend tests trip Bandit's
+  # "possible hardcoded password" (B105/B106); they're test fixtures,
+  # not secrets, so exclude the file from analysis.
+  - "onadata/apps/main/tests/test_backends.py"
--- a/onadata/apps/main/backends.py
+++ b/onadata/apps/main/backends.py
@@ -35,13 +35,25 @@ class ModelBackend(DjangoModelBackend):
 
         username_matches = User.objects.filter(username__iexact=username)
         email_matches = User.objects.filter(email__iexact=username)
+        # Username is unique, so username matches are tried first; email is not,
+        # so email collisions are resolved by password. Candidates are
+        # de-duplicated (a value can match both a username and an email) so each
+        # account's deliberately-expensive password hash is verified at most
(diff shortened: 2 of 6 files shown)


FrankApiyo:
### Review feedback — resolution summary

Findings from review of the auth-backend change have been addressed across `3e22a71` and `4e0ddce`:

| Finding | Severity | Resolution | Commit |
|---|---|---|---|
| Org-account guard used `return None`, which could block a legitimate human account sharing the email | HIGH | Skip org candidates with `continue` instead of aborting the lookup | `3e22a71` |
| Custom `authenticate` override dropped Django's `is_active` gate (deactivated users could authenticate, reachable via Basic Auth) | HIGH | Restored `self.user_can_authenticate(user)` check | `4e0ddce` |
| Password hash verified more than once when a value matched both a username and an email | LOW | De-duplicate candidates via `seen_ids` | `3e22a71` |
| Per-candidate password-check cost not documented | LOW | Comment clarifying work is bounded by accounts sharing the value | `3e22a71` |
| Empty-string credentials not guarded | LOW | `if not username or not password` | `4e0ddce` |

**New tests:** inactive user rejected with a valid password; a value matching one account's username and another's email resolves correctly (username precedence, email fallback); `test_user_with_unusable_password_cannot_log_in` no longer patches `is_organization_user`, so it actually exercises the password path.

**Deliberately not changed:**
- **Dummy-hash timing parity** (LOW) — skipped by maintainer decision.
- **`ona-oidc` pin kept as a commit SHA** rather than a tag — intentional for stage-api.ona.io testing; should become a tagged release before merge.

ukanga (changes requested):
Let's handle the issue in the DB instead of introducing a code change.

PR #3092, opened 2026-05-26
feat(security): enforce strict file-type validation on uploads

ukanga:

Deny-by-default upload validation across all entry points via a central `upload_validation` module: per-extension MIME allowlist + magic-byte/structural content checks, configurable size caps (`STRICT_UPLOAD_MAX_BYTES`), and UUID storage names so client-supplied filenames never reach disk. Covers XLSForm publishing, CSV/XLS imports, form media, and supporting docs. Adds an RFC 6266 `Content-Disposition` parser and sets `X-Content-Type-Options: nosniff` on file downloads.

`flake8`, `isort`, `black` clean; unit tests for the validator and parser pass.

Stricter uploads: the legacy content-type-only allowlist (and audio/SVG/DOC-OLE/ZIP types) is removed. Operators raising a per-extension cap must also raise nginx `client_max_body_size`.

  - [x] Included tests
  - [ ] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/logger/models/xform.py
+++ b/onadata/apps/logger/models/xform.py
@@ -8,6 +8,7 @@ import hashlib
 import json
 import os
 import re
+import uuid
 from datetime import datetime
 from io import BytesIO
 from pathlib import Path
@@ -139,8 +140,16 @@ def question_types_to_exclude(_type):
 
 
 def upload_to(instance, filename):
-    """Returns the path to upload an XLSForm file to."""
-    return os.path.join(instance.user.username, "xls", os.path.split(filename)[1])
+    """Returns the path to upload an XLSForm file to.
+
+    The basename is randomised so the stored path never includes any
+    client-supplied name. The original filename is preserved on the in-memory
+    file object during ``DataDictionary.save()`` so pyxform can derive
+    ``fallback_form_name`` from it for forms without an explicit
+    ``settings.id_string``.
+    """
+    extension = os.path.splitext(os.path.split(filename)[1])[1].lower()
+    return os.path.join(instance.user.username, "xls", f"{uuid.uuid4().hex}{extension}")
 
 
 def contains_xml_invalid_char(text, invalids=None):
@@ -874,7 +883,9 @@ class XFormMixin:
 
         if not hasattr(self, "_variable_names"):
             # pylint: disable=import-outside-toplevel
-            from onadata.apps.viewer.models.column_rename import ColumnRename
+            from onadata.apps.viewer.models.column_rename import (  # noqa: PLC0415
+                ColumnRename,
+            )
 
             self._variable_names = ColumnRename.get_dict()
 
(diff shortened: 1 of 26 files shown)


onadata/apps/logger/models/xform.py
```diff
@@ -139,8 +140,16 @@ def question_types_to_exclude(_type):
...
+    """Returns the path to upload an XLSForm file to.
+
+    The basename is randomised so the stored path never includes any
+    client-supplied name. The original filename is preserved on the in-memory
+    file object during ``DataDictionary.save()`` so pyxform can derive
+    ``fallback_form_name`` from it for forms without an explicit
+    ``settings.id_string``.
+    """
+    extension = os.path.splitext(os.path.split(filename)[1])[1].lower()
+    return os.path.join(instance.user.username, "xls", f"{uuid.uuid4().hex}{extension}")
```
kelvin-muchiri: What could be the negative implications of doing this? I guess one is that downloading an XLS file now will have the random name instead of the original name?

onadata/apps/main/models/meta_data.py
```diff
@@ -429,26 +448,27 @@ def media_upload(content_object, data_file=None, download=False):
...
-                        "data_file_type": data_content_type,
-                    },
+            try:
+                upload = validate_uploaded_file(
+                    data_file, FORM_MEDIA_ALLOWED_EXTENSIONS, FORM_MEDIA_UPLOAD_CONTEXT
                 )
+            except UploadValidationError as error:
+                raise ValidationError({"data_file": str(error)}) from error
+
+            data_file.name = upload.storage_basename
```
kelvin-muchiri: This implementation appears similar to `supporting_docs` docs implementation. Can we DRY?

onadata/apps/viewer/models/data_dictionary.py
```diff
@@ -126,11 +128,18 @@ def sheet_to_csv(xls_content, sheet_name):
...
+    can derive ``fallback_form_name`` from it during publication, but the
+    stored basename is randomised to prevent client-influenced storage paths
+    and double-extension reads from the filesystem.
     """
     if instance:
         username = instance.xform.user.username
-    return os.path.join(username, "xls", os.path.split(filename)[1])
+    original_basename = os.path.split(filename)[1]
+    extension = os.path.splitext(original_basename)[1].lower()
+    return os.path.join(username, "xls", f"{uuid.uuid4().hex}{extension}")
```
kelvin-muchiri: This implementation is similar to the one in `xform.py`, can we DRY?

onadata/apps/logger/models/xform.py
ukanga: Only if accessing the storage directly, not through the API. By default, the downloaded file will now have the `{id_string}.{extension}` format. For example, the form Registration Form with the `id_string = 'reg_form'`  will result in the XLSForm file reg_form.xlsx.

kelvin-muchiri (approved):

PR #3127, opened 2026-06-11
feat(security): rate-limit HTTP Basic and OAuth2 password-grant logins

kelvin-muchiri:

Stock `BasicAuthentication` and the OAuth2 token endpoint (`/o/token/`) returned errors indefinitely on wrong credentials, so both could be brute-forced at full network speed — no lockout, rate limiting, or CAPTCHA. Usernames are publicly enumerable (public projects endpoint) and a successful basic auth returns the persistent, non-expiring `api_token`, making this **high severity**.

This complements #3116, which added the same lockout to `DigestAuthentication` and the web login form but left the Basic-auth and OAuth2 password-grant paths unprotected. All password vectors (digest, basic, web form, `/o/token/`) now share one IP + username lockout counter, so an attacker can't reset by switching vector.

- **`onadata/libs/authentication.py`** — add `LockoutBasicAuthentication(BasicAuthentication)`. It reuses the existing IP + username lockout machinery (`assert_not_locked_out` / `add_login_attempt`): rejects a locked-out IP + username *before* checking credentials, and records a failed attempt — locking out after `MAX_LOGIN_ATTEMPTS` for `LOCKOUT_TIME` — whenever authentication fails. The submitted identifier is canonicalised via `get_lockout_username`, so case/email variants share one counter. Extracted `is_lockout_excluded_path()` and reused it in `check_lockout` so the OpenRosa/Briefcase exemptions stay DRY.
- **`onadata/apps/api/viewsets/xform_submission_viewset.py`** — replace stock `BasicAuthentication` with `LockoutBasicAuthentication` (the only in-code Basic-auth surface).
- **`onadata/apps/main/views.py`** — add `LockoutTokenView(TokenView)`. The Resource Owner Password Credentials grant accepts a username + password and was otherwise brute-forceable; this rejects a locked-out IP + username before the grant and records a failed attempt on any non-200 password-grant response, returning a `429 locked_out`. Only the password grant is throttled — `authorization_code`, `client_credentials`, `refresh_token`, and `device_code` pass straight through.
- **`onadata/apps/main/urls.py`** — route `o/token/` to `LockoutTokenView` ahead of the `oauth2_provider` include (same precedence pattern as the existing `OnaAuthorizationView` override).

Behaviour mirrors the digest flow: keyed on IP + username, does **not** reset the counter on success, sends the lockout email at the threshold, and exempts the high-frequency OpenRosa/Briefcase endpoints (`LOCKOUT_EXCLUDED_PATHS`) so legitimate ODK Collect/Briefcase clients aren't locked out.

- Added tests

- Failed attempts via digest, the web login form, basic auth, and the OAuth2 password grant all accumulate against the same IP + username counter, settings (`MAX_LOGIN_ATTEMPTS`, `LOCKOUT_TIME`), and cache keys.
- On `XFormSubmissionViewSet`, Basic-auth submissions are now subject to lockout — but the OpenRosa submission paths are in `LOCKOUT_EXCLUDED_PATHS`, so legitimate ODK clients are unaffected (parity with digest).
- `/o/token/` is now served by a project view rather than the bundled `oauth2_provider` `TokenView`; behaviour is identical except for the password-grant lockout.

  - [x] Included tests
  - [ ] Updated documentation

### DIFF
--- a/onadata/libs/tests/test_authentication.py
+++ b/onadata/libs/tests/test_authentication.py
@@ -1,3 +1,4 @@
+import base64
 from datetime import datetime, timedelta
 from unittest.mock import MagicMock, patch
 
@@ -5,16 +6,17 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.cache import cache
 from django.http.request import HttpRequest
-from django.test import TestCase
+from django.test import TestCase, override_settings
 
 import jwt
-from oauth2_provider.models import AccessToken
+from oauth2_provider.models import AccessToken, get_application_model
 from rest_framework.exceptions import AuthenticationFailed
 from rest_framework.test import APIRequestFactory
 
 from onadata.apps.api.models.temp_token import TempToken
 from onadata.libs.authentication import (
     DigestAuthentication,
+    LockoutBasicAuthentication,
     MasterReplicaOAuth2Validator,
     TempTokenAuthentication,
     TempTokenURLParameterAuthentication,
@@ -23,6 +25,7 @@ from onadata.libs.authentication import (
     get_api_token,
     get_client_ip,
     get_lockout_username,
+    is_lockout_excluded_path,
 )
 from onadata.libs.utils.cache_tools import LOCKOUT_IP, safe_cache_set, safe_key
 from onadata.libs.utils.common_tags import API_TOKEN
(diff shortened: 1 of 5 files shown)


onadata/libs/tests/test_authentication.py
```diff
@@ -296,3 +334,113 @@ def is_valid_mock(*args, **kwargs):
...
+            "/", HTTP_AUTHORIZATION=_basic_header("bob", "secret")
+        )
+        with self.assertRaises(AuthenticationFailed) as ctx:
+            self.auth.authenticate(request)
+        self.assertIn("Locked out", str(ctx.exception))
+
+    def test_lockout_is_keyed_per_username(self):
+        """Locking one username does not lock another from the same IP."""
+        User.objects.create_user(username="alice", password="secret")
+        max_attempts = getattr(settings, "MAX_LOGIN_ATTEMPTS", 10)
```
ukanga: Can we use a settings override instead and reduce this down to 3 attempts before it locks out. Fewer iterations and being deliberate and exact in tests.

ukanga (approved):

PR #3155, opened 2026-06-30
feat(email): verified email-change flow

FrankApiyo:

Verified (verify-before-apply) email change: an OTP is sent to the new address and the change is applied only on confirmation. Endpoints, throttling and expiry behavior are documented in `docs/profiles.rst`.

- Unit tests: OTP TTL / attempt cap / hashing, request throttling, confirm applies + invalidates the cached profile, race-on-confirm and replay both rejected
- Manual end-to-end from the SPA (request → OTP email → confirm)

- Schema change: new `PendingEmailChange` table, with a management command to purge expired rows

  - [x] Included tests
  - [x] Updated documentation

Closes [ONADATA-1411](https://linear.app/ona/issue/ONADATA-1411/email-changes-do-not-update-keycloak-email)


onadata/apps/main/migrations/0020_pendingemailchange_and_more.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 5.2.11 on 2026-06-29 10:32
+
```
kelvin-muchiri, line 2: Do we need a new table for this? I think all this info can be encoded and signed within the link we send to the user. Can something like this suffice?

```
class EmailChangeTokenGenerator(PasswordResetTokenGenerator):
    """Token generator for change in email link"""

    def _make_hash_value(self, user, timestamp):
        return (
            six.text_type(user.pk)
            + six.text_type(timestamp)
            + six.text_type(user.email)
        )
```

kelvin-muchiri: But I now see we need a place to store the new email.

FrankApiyo: Yes, we need to keep the new email in the DB and also keep track of the number of attempts the user has taken to reset the email.

Also, I don't think it would be secure to allow users to have access to the hashed OTP in the email alongside the OTP.

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
         user = User.objects.get(username=data["username"])
         self.assertFalse(user.is_active)
+
+    # ---------------------------------------------------------------------------
+    # request_email_change tests
+    # ---------------------------------------------------------------------------
+
+    def test_request_email_change_happy_path(self):
+        """Happy path: returns 200, creates PendingEmailChange, sends 1 email."""
+        cache.clear()  # start with a clean send-throttle counter
```
kelvin-muchiri: I think it would be better to have `cache.clear` in `tearDown` instead, [example](https://github.com/onaio/onadata/blob/main/onadata/libs/utils/tests/test_entities_utils.py#L528)

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+        response = view(request, user="bob")
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(PendingEmailChange.objects.filter(user=self.user).exists())
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertEqual(mail.outbox[0].to, ["new@x.com"])
+        self.assertRegex(mail.outbox[0].body, r"\d{6}")
+
+    def test_request_email_change_bad_password(self):
+        """Wrong password returns 400 with a 'password' error key."""
+        cache.clear()  # clean send-throttle counter
```
kelvin-muchiri: Same as above, let's move `cache.clear` to `tearDown`

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+        view = UserProfileViewSet.as_view({"post": "request_email_change"})
+        request = self.factory.post(
+            "/",
+            data={"new_email": "other@example.com", "password": "wrongpass"},
+            **self.extra,
+        )
+        response = view(request, user="bob")
+        self.assertEqual(response.status_code, 400)
+        self.assertIn("password", response.data)
```
kelvin-muchiri: Can we assert more  info in the response data other than just `password`?

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+            username="claire", email="taken@example.com", password="clairepass"
+        )
+        request = self.factory.post(
+            "/",
+            data={"new_email": "TAKEN@Example.com", "password": "bobbob"},
+            **self.extra,
+        )
+        response = view(request, user="bob")
+        self.assertEqual(response.status_code, 400)
+        self.assertIn("new_email", response.data)
```
kelvin-muchiri: Can we assert more in the response data other than just `new_email`?

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+
+    def test_request_email_change_throttled(self):
+        """OTP send is rate-limited so it can't spam the target address."""
+        from onadata.apps.api.viewsets.user_profile_viewset import (
```
kelvin-muchiri: Can we move this import to the top?

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+            MAX_EMAIL_CHANGE_REQUESTS,
+        )
+
+        cache.clear()
```
kelvin-muchiri: Let's move `cache.clear` to `tearDown`

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
kelvin-muchiri: Instead of importing `MAX_EMAIL_CHANGE_REQUESTS` can we use `override_settings` decorator?

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+        )
+        response = view(blocked, user="bob")
+        self.assertEqual(response.status_code, 429)
+        self.assertEqual(len(mail.outbox), MAX_EMAIL_CHANGE_REQUESTS)
+
+    def test_request_email_change_throttle_gates_entry(self):
+        """The rate limit gates entry, not just the send: even calls that fail
+        the password check count against the budget, so the password/uniqueness
+        checks can't be exercised as unbounded oracles."""
+        from onadata.apps.api.viewsets.user_profile_viewset import (
```
kelvin-muchiri: Let's use `override_settings` instead of importing from the viewset

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+            MAX_EMAIL_CHANGE_REQUESTS,
+        )
+
+        cache.clear()
```
kelvin-muchiri: Let's move `cache.clear` to `tearDown`

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+        view = UserProfileViewSet.as_view({"post": "request_email_change"})
+
+        # Exhaust the budget with wrong-password calls (each returns 400).
+        for _ in range(MAX_EMAIL_CHANGE_REQUESTS):
```
kelvin-muchiri: This is ok but I think for this test we can just set the cache value directly to simulate  exhausted retry limit

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+    # ---------------------------------------------------------------------------
+
+    def test_confirm_email_change_bad_otp(self):
+        """Invalid OTP returns 400 with an 'otp' error key; email is not changed."""
+        view_confirm = UserProfileViewSet.as_view({"post": "confirm_email_change"})
+        PendingEmailChange.start(self.user, "new@x.com")
+        request = self.factory.post("/", data={"otp": "000000"}, **self.extra)
+        response = view_confirm(request, user="bob")
+        self.assertEqual(response.status_code, 400)
+        self.assertIn("otp", response.data)
```
kelvin-muchiri: We can assert more info in the response data other than just "otp"

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+        self.user.refresh_from_db()
+        self.assertNotEqual(self.user.email, "new@x.com")
+
+    def test_confirm_email_change_no_pending(self):
+        """No pending record returns 400."""
+        view_confirm = UserProfileViewSet.as_view({"post": "confirm_email_change"})
+        request = self.factory.post("/", data={"otp": "123456"}, **self.extra)
+        response = view_confirm(request, user="bob")
+        self.assertEqual(response.status_code, 400)
+        self.assertIn("otp", response.data)
```
kelvin-muchiri: We can assert more info from the response data other just "otp"

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+        self.user.refresh_from_db()
+        self.assertEqual(self.user.email, "new@x.com")
+        self.assertFalse(PendingEmailChange.objects.filter(user=self.user).exists())
+        self.user.profile.refresh_from_db()
+        self.assertTrue(self.user.profile.metadata.get("is_email_verified"))
+
+    def test_confirm_invalidates_profile_cache(self):
+        """The cached profile response must be cleared so the new email is
+        served immediately, not stale until the cache TTL expires."""
+        from onadata.libs.utils.cache_tools import (
```
kelvin-muchiri: Let's move the import to the top. I think to avoid coupling between implementation and tests, we can use the literal string for the key e.g `cache_key = f"foo-{self.user.username}{self.user.username}"` and use standard `cache.get` and `cache.set`

onadata/libs/utils/email.py
```diff
@@ -121,6 +121,25 @@ def send_generic_email(email, message_txt, subject):
...
+def send_email_change_code(to_email: str, code: str) -> None:
+    """Send a numeric OTP to *to_email* as part of an email-change request.
+
+    Uses synchronous delivery so Django's test ``mail.outbox`` captures it
+    without requiring Celery workers.
+    """
+    ttl_seconds = getattr(settings, "EMAIL_CHANGE_OTP_TTL_SECONDS", 300)
+    ttl_minutes = max(1, ttl_seconds // 60)
+    subject = "Verify your new email address"
+    body = (
```
kelvin-muchiri: Should we have a HTML template like the rest of the emails?

onadata/apps/main/tests/models/test_pending_email_change.py
```diff
@@ -0,0 +1,74 @@
...
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+from django.utils import timezone
+
+from onadata.apps.main.models import PendingEmailChange
+
+User = get_user_model()
+
+
+class PendingEmailChangeTests(TestCase):
```
kelvin-muchiri, line 16: Could we add docstrings for the test methods?

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -294,6 +316,101 @@ def change_password(self, request, *args, **kwargs):  # noqa
 
         return Response(status=status.HTTP_200_OK, data=data)
 
+    @action(methods=["POST"], detail=True, url_path="request_email_change")
```
kelvin-muchiri: The standard we've used is hyphens  instead of underscors for URLs i. e  `request-email-change` over `request_email_change`

We also need to add documentation for the new endpoint

onadata/apps/api/viewsets/user_profile_viewset.py
```diff
@@ -294,6 +316,101 @@ def change_password(self, request, *args, **kwargs):  # noqa
...
+                {"new_email": _("This email address is already in use.")},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        _pec, code = PendingEmailChange.start(user, new_email)
+        send_email_change_code(new_email, code)
+
+        return Response({"sent": True}, status=status.HTTP_200_OK)
+
+    @action(methods=["POST"], detail=True, url_path="confirm_email_change")
```
kelvin-muchiri: The standard we've used is hyphens instead of underscors for URLs i. e confirm-email-change over confirm_email_change

We also need to add documentation for the new endpoint

onadata/apps/api/tests/viewsets/test_user_profile_viewset.py
```diff
@@ -2122,3 +2124,195 @@ def test_profile_create_user_inactive_when_disabled(self):
...
+        view_confirm = UserProfileViewSet.as_view({"post": "confirm_email_change"})
+        pec, code = PendingEmailChange.start(self.user, "raced@x.com")
+        # Another user registers the target email before bob confirms
+        User.objects.create_user(
+            username="racer", email="raced@x.com", password="racerpass"
+        )
+        request = self.factory.post("/", data={"otp": code}, **self.extra)
+        response = view_confirm(request, user="bob")
+        self.assertEqual(response.status_code, 400)
+        self.assertIn("new_email", response.data)
```
kelvin-muchiri: Can we assert more from the response other than `new_email`?

PR #3172, opened 2026-07-10
feat(v2): server-side project search, sort & role/starred/shared filters

FrankApiyo:

Server-side `search`, `ordering`, `shared`, `starred` and `role` query params on `GET /api/v2/projects`, so clients can query projects without fetching the full list. v1 is untouched. Parameter behavior is documented in `docs/projects.rst`.

- Unit tests covering each param, combined params, both ordering directions, anonymous requests and invalid-input rejection

- None expected: the new filters only narrow the already permission-scoped queryset, and the `last_submission_date` annotation is applied only when requested for ordering

  - [x] Included tests
  - [x] Updated documentation

### DIFF
--- a/onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
+++ b/onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
@@ -678,3 +678,32 @@ class ProjectSearchTestCase(TestAbstractViewSet):
         response = self.view(request)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.data, [])
+
+
+@override_settings(TIME_ZONE="UTC")
+class ProjectOrderingTestCase(TestAbstractViewSet):
+    """?ordering= sorts by name / created."""
+
+    def setUp(self):
+        super().setUp()
+        # Created oldest → newest: Banana, Apple, Cherry
+        self.banana = Project.objects.create(
+            name="Banana", organization=self.user, created_by=self.user)
+        self.apple = Project.objects.create(
+            name="Apple", organization=self.user, created_by=self.user)
+        self.cherry = Project.objects.create(
+            name="Cherry", organization=self.user, created_by=self.user)
+        self.view = ProjectViewSet.as_view({"get": "list"})
+
+    def _names(self, response):
+        return [p["name"] for p in response.data]
+
+    def test_order_by_name_asc(self):
+        request = self.factory.get("/", {"ordering": "name"}, **self.extra)
+        response = self.view(request)
+        self.assertEqual(self._names(response), ["Apple", "Banana", "Cherry"])
+
+    def test_order_by_created_desc(self):
+        request = self.factory.get("/", {"ordering": "-created"}, **self.extra)
+        response = self.view(request)
+        self.assertEqual(self._names(response), ["Cherry", "Apple", "Banana"])
--- a/onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
+++ b/onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
@@ -878,7 +878,7 @@ class ProjectStarredFilterTestCase(ProjectListFilterTestBase):
 
 @override_settings(TIME_ZONE="UTC")
 class ProjectRoleFilterTestCase(ProjectListFilterTestBase):
-    """?role= returns projects where the requesting user's role is in the set."""
+    """?role= returns projects where the user holds that role or higher."""
 
     def setUp(self):
         super().setUp()
@@ -903,11 +903,6 @@ class ProjectRoleFilterTestCase(ProjectListFilterTestBase):
         """?role=owner returns only projects where the user is an owner."""
         response = self._list({"role": "owner"}, extra=self.member_extra)
         self.assertEqual(self._names(response), ["Owner Proj"])
-
-    def test_role_owner_or_manager(self):
-        """?role=owner,manager returns manager-and-above projects."""
-        response = self._list({"role": "owner,manager"}, extra=self.member_extra)
-        self.assertEqual(self._names(response), ["Manager Proj", "Owner Proj"])
 
     def test_role_editor_includes_higher_roles(self):
         """?role=editor means editor-and-above: manager and owner projects
--- a/onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
+++ b/onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
@@ -692,6 +692,17 @@ class ProjectSearchTestCase(ProjectListFilterTestBase):
         response = self._list({"search": self.user.username})
         self.assertEqual(self._names(response), ["Household Census", "Rainfall Survey"])
 
+    def test_search_by_owner_username_case_insensitive_substring(self):
+        """?search= matches a case-variant substring of the owner's username."""
+        response = self._list({"search": self.user.username[:-1].upper()})
+        self.assertEqual(self._names(response), ["Household Census", "Rainfall Survey"])
+
+    def test_search_no_match(self):
+        """?search= returns an empty list when nothing matches."""
+        response = self._list({"search": "no-such-project"})
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data, [])
+
     def test_search_combines_with_ordering(self):
         """?search= and ?ordering= can be combined in a single request."""
         response = self._list({"search": self.user.username, "ordering": "-name"})
@@ -703,7 +714,14 @@ class ProjectSearchTestCase(ProjectListFilterTestBase):
 
 @override_settings(TIME_ZONE="UTC")
 class ProjectOrderingTestCase(ProjectListFilterTestBase):
-    """?ordering= sorts by name / created."""
+    """?ordering= sorts by name / created / category / last submission.
+
+    ``setUp`` only creates the three bare projects; tests that sort on
+    ``metadata__category`` or ``last_submission_date`` add those fixtures
+    via ``_set_categories`` / ``_make_forms``. The fixture orders are
+    chosen so no two sort fields agree — a test can't pass by falling
+    back to another ordering.
+    """
 
     def setUp(self):
         super().setUp()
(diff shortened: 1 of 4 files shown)


onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
```diff
@@ -641,3 +642,282 @@ def test_non_member_denied_public_project(self):
...
+        self.assertEqual(self._names(response), ["Plain One", "Starred One"])
+
+    def test_filter_starred_invalid_value(self):
+        """A ?starred= value other than true/false is a 400, not silently false."""
+        response = self._list({"starred": "banana"})
+        self.assertEqual(response.status_code, 400)
+
+
+@override_settings(TIME_ZONE="UTC")
+class ProjectRoleFilterTestCase(ProjectListFilterTestBase):
```
kelvin-muchiri, line 912: We can add a test case for when the requesting user has no role in the project

FrankApiyo: done!

onadata/libs/filters.py
```diff
@@ -142,6 +152,76 @@ class Meta:
...
+            return queryset.none()
+        allowed = get_objects_for_user(
+            request.user,
+            perms,
+            klass=queryset,
+            any_perm=False,
+            accept_global_perms=False,
+            with_superuser=False,
+        )
+        return queryset.filter(pk__in=allowed.values("pk"))
```
kelvin-muchiri: `get_objects_for_user` returns a filtered queryset, is there need to further filter?

onadata/libs/filters.py
```diff
@@ -142,6 +152,76 @@ class Meta:
...
+            return queryset.none() if str2bool(starred) else queryset
+
+        if str2bool(starred):
+            return queryset.filter(user_stars=request.user)
+
+        return queryset.exclude(user_stars=request.user)
+
+
+# pylint: disable=too-few-public-methods
+class ProjectRoleFilter(filters.BaseFilterBackend):
```
kelvin-muchiri, line 197: Our intention for adding the role filter was that it would allow us to filter projects where the requesting user is the owner.  As much as this adds additional support for filtering by _any_ role, will [ProjectOwnerFilter](https://github.com/onaio/onadata/blob/main/onadata/libs/filters.py#L261) suffice and we can drop this for now?

onadata/libs/filters.py
FrankApiyo: No, in this case, there's no need for any further filters because the queryset that this filter is operating on has already been filtered and narrowed down by all the filters configured in `filter_backends` for the `ProjectViewset` above.

The most important filter is the `AnonUserProjectFilter` that inherits from `ObjectPermissionsFilter`; which implies that the queryset we are dealing with here has already been filtered down to what projects `request.user` has permissions for.

onadata/libs/filters.py
FrankApiyo: The `ProjectOwnerFilter` and the `ProjectRoleFilter` are subtly different.

`ProjectOwnerFilter` is checking the `organization` field of the project to filter for projects that a user owns
`ProjectRoleFilter` on the other hand, filters by permissions to find which project a user has specific permissions in. Keep in mind that it only filters projects that the user already has access to, so that it can narrow down to a specific permission level

FrankApiyo: To add more context to this, both filters are needed for different functionality:

When you want all projects belonging to an organization,  we will use v2/projects?owner=acme (this is the ProjectOwnerFilter)
When you want to fetch projects that the user has manager or owner permissions to(because you want to transfer a form to the new project), now we will use our new filter (ProjectRoleFilter)

onadata/libs/filters.py
kelvin-muchiri: @FrankApiyo Can we do, Why do we need to `queryset.filter(pk__in=allowed.values("pk"))`

```
return get_objects_for_user(
            request.user,
            perms,
            klass=queryset,
            any_perm=False,
            accept_global_perms=False,
            with_superuser=False,
        )
```

onadata/libs/filters.py
```diff
@@ -142,6 +152,76 @@ class Meta:
...
+    """Project `role` filter using the `role` query parameter."""
+
+    # pylint: disable=unused-argument
+    def filter_queryset(self, request, queryset, view):
+        """Filter by the requesting user's role (from object permissions).
+
+        ``?role=editor,owner`` returns projects where the user holds the
```
kelvin-muchiri: Any particular reason why we need to support an array of values?

FrankApiyo: We don't need to support an array of values.

fixed!

onadata/apps/api/viewsets/v2/project_viewset.py
```diff
@@ -28,6 +39,44 @@ class ProjectViewSet(ProjectViewSetV1):
...
+
+        Overrides super().get_queryset()
+        """
+        queryset = super().get_queryset()
+        ordering = self.request.query_params.get("ordering", "")
+        requested = {field.strip() for field in ordering.split(",")}
+        if requested & {"last_submission_date", "-last_submission_date"}:
+            # Annotation is only for ordering; the serializer still computes
+            # last_submission_date itself. Exclude soft-deleted forms.
+            queryset = queryset.annotate(
```
kelvin-muchiri, line 73: Why do we need this implementation?

FrankApiyo: I believe it's needed because projects don't have a `last_submission_date` field, and so when you do `projects?ordering=last_submission_date`, it needs to annotate the projects before sorting using `OrderingFilter`

At this point, the serializer hasn't serialized the projects, so they still don't have that field(it's usually added by the serializer)
....and we only need the annotation when we need to do ordering this way

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
```diff
@@ -641,3 +642,289 @@ def test_non_member_denied_public_project(self):
...
+            extra = self.extra
+        return self.view(self.factory.get("/", params, **extra))
+
+    def _names(self, response, sort=True):
+        names = [project["name"] for project in response.data]
+        return sorted(names) if sort else names
+
+
+@override_settings(TIME_ZONE="UTC")
+class ProjectSearchTestCase(ProjectListFilterTestBase):
```
kelvin-muchiri, line 669: Looks like we need to add tests for _all_ search fields. e.g tests for searching by organization username are missing

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
```diff
@@ -641,3 +642,289 @@ def test_non_member_denied_public_project(self):
...
+        """?search= and ?ordering= can be combined in a single request."""
+        response = self._list({"search": self.user.username, "ordering": "-name"})
+        self.assertEqual(
+            self._names(response, sort=False),
+            ["Rainfall Survey", "Household Census"],
+        )
+
+
+@override_settings(TIME_ZONE="UTC")
+class ProjectOrderingTestCase(ProjectListFilterTestBase):
```
kelvin-muchiri, line 716: Looks like we need to add tests for _all_ ordering fields. e.g tests for ordering by `last_submission_date` ascending and descending are missing

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
```diff
@@ -641,3 +642,289 @@ def test_non_member_denied_public_project(self):
...
+    def test_order_by_created_asc(self):
+        """?ordering=date_created sorts the oldest-created project first."""
+        response = self._list({"ordering": "date_created"})
+        self.assertEqual(
+            self._names(response, sort=False), ["Banana", "Apple", "Cherry"]
+        )
+
+
+@override_settings(TIME_ZONE="UTC")
+class ProjectOrderingDerivedTestCase(ProjectListFilterTestBase):
```
kelvin-muchiri: Why are the tests for this test case not in `ProjectOrderingTestCase` instead?

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
FrankApiyo: Actually, that's covered by the test case named `test_search_by_owner_username` 🤓 

Keep in mind that the `.organization` field is the project owner field. So `organization__username` is the owner username

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
FrankApiyo: `last_submission_date`(ascending and descending) are in the `ProjectOrderingDerivedTestCase` test class

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
FrankApiyo: These classes(`ProjectOrderingTestCase` and `ProjectOrderingDerivedTestCase`) have been separated into 2 for clarity/simplicity because they have different test setup requirements. 

If we mix both, the `setUp` mtd. is going to be larger and more complex because it would be doing too much

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
```diff
@@ -641,3 +642,289 @@ def test_non_member_denied_public_project(self):
...
+    def test_filter_shared_true(self):
+        """?shared=true returns only public projects."""
+        response = self._list({"shared": "true"})
+        self.assertEqual(self._names(response), ["Public One"])
+
+    def test_filter_shared_false(self):
+        """?shared=false returns only private projects."""
+        response = self._list({"shared": "false"})
+        self.assertEqual(self._names(response), ["Private One"])
+
```
ukanga, line 862: @FrankApiyo Can we add a test case where the shared parameter might not be the expected `true` or `false`? For example `shared=banana`?

ukanga (changes requested):

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
kelvin-muchiri: I missed that

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
```diff
@@ -641,3 +642,289 @@ def test_non_member_denied_public_project(self):
...
+            metadata={"category": "agriculture"},
+        )
+        self.p_gov = Project.objects.create(
+            name="Gov",
+            organization=self.user,
+            created_by=self.user,
+            metadata={"category": "governance"},
+        )
+
+    def _make_form(self, project, id_string, last_submission_time):
```
kelvin-muchiri: Conventionally, we've leaned on publishing a form by markdown in tests i. e using method `_publish_markdown`. I don't think we've used this method of creating a form from XML

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
FrankApiyo: done!

onadata/apps/api/tests/viewsets/v2/test_project_viewset.py
FrankApiyo: done!

kelvin-muchiri (approved):

ukanga (approved):

PR #3183, opened 2026-07-20
fix: soft delete suffix handling and name reuse for EntityList and XForm

### LINKED ISSUE #3182: Soft delete suffix is truncated or dropped for long EntityList and XForm names
### Environmental Information

- Onadata version: 5.22.3

### Problem description

`EntityList.soft_delete()` appends `-deleted-at-<epoch>` to `name` and then truncates to the field's 255-character limit. `XForm.soft_delete()` does the same to `id_string` and `sms_id_string` with a 100-character limit (`MAX_ID_LENGTH`). When the original value is near the limit, the suffix is partially or entirely cut off:

- Value exactly at max length: the suffix is dropped entirely — the soft-deleted row is saved under its unchanged `name`/`id_string`.
- Value just under max length: only a deterministic fragment of the suffix fits (e.g. `-deleted-at-` or a prefix of it, with no epoch digits). Soft-deleting a re-created object with the same name produces an identical mangled value and raises `IntegrityError` under the uniqueness constraint. `restore()` also cannot reliably strip a partial suffix.

### Expected behavior

- If appending the full deletion suffix would exceed the field's max length, do not append it at all (never store a truncated suffix).
- Override `__str__` on `EntityList` and `XForm` to include the deletion suffix when the object is soft-deleted but the stored `name`/`id_string` does not contain it, so soft-deleted objects remain distinguishable in display output.

1. Create an `EntityList` whose name is 250 characters long.
2. Call `soft_delete()` — the name is stored as the 250 characters plus `-dele` (partial suffix).
3. Create a new `EntityList` with the same 250-character name in the same project.
4. Call `soft_delete()` on the new one — `IntegrityError`, since the mangled name is identical to step 2's.

Alternatively, create an `EntityList` with a 255-character name and call `soft_delete()` — the stored name is unchanged.

### Additional Information

- Applies to `EntityList.name` (255 chars) and `XForm.id_string`/`sms_id_string` (100 chars).
- Known limitation of the chosen approach: a name at exactly max length still occupies the name while soft-deleted (`unique_together` on `name`/`project`), so re-creating it remains blocked until the object is restored or hard-deleted. Freeing the name fully would require a conditional unique constraint (partial index on `deleted_at IS NULL`), which is out of scope here.

kelvin-muchiri:

- Never store a truncated deletion suffix
- Free EntityList name that does not have deletion suffix for reuse after soft delete
- Free XForm id_string that does not have deletion suffix  for reuse after soft delete

- [x] Added tests

- Migrations `0039` and `0040` create partial unique indexes on the `logger_entitylist` and `logger_xform` tables

  - [x] Included tests
  - [ ] Updated documentation

Closes #3182


ukanga (changes requested):
Is this changing id_string uniqueness to be project level rather than account level?

ukanga (dismissed):

ukanga (approved):

PR #3176, opened 2026-07-15
chore(deps): bump ona-oidc for the account proxy (blocked on ona-oidc#124)

FrankApiyo:
### Changes / Features implemented

Bump ona-oidc to pull in the Keycloak Account REST proxy (sessions / linked-accounts / credentials endpoints + CSRF gating) from onaio/ona-oidc#124.

### Steps taken to verify this change does what is intended

- Smoke-tested against the account proxy on stage
- Covered by the ona-oidc test suite

### Side effects of implementing this change

- Blocked on onaio/ona-oidc#124: the pin targets that PR's head; retarget to the release SHA/tag once it merges and a version is cut, before taking this out of draft

**Before submitting this PR for review, please make sure you have:**

  - [x] Included tests
  - [ ] Updated documentation

PR #3185, opened 2026-07-24
fix(oidc): serve the ona-oidc account-proxy routes

FrankApiyo:

Serve ona-oidc's Keycloak account-proxy routes (`/oidc/<server>/sessions`, `/linked-accounts`, `/credentials`) so the SPA's Account tab works, and bump ona-oidc to pick up onaio/ona-oidc#124.

- **Inherit ona-oidc's URLconf instead of copying it.** #3090 replaced `include("oidc.urls")` with an inline copy so routes could point at `OnaOpenIDConnectViewset`, which refuses SSO login for organization accounts. ona-oidc now takes `VIEWSET_CLASS`, so the blanket include is restored without losing that rejection. The copy had already drifted — the `session` route lost its `$` anchor and prefix-matched `/sessions`.
- **Add an opt-in `OnaKeycloakOpenIDConnectViewset`.** ona-oidc moved the proxy actions into `KeycloakAccountMixin` and derives routes from the `@action` decorators on whichever class `VIEWSET_CLASS` names. The proxy is Keycloak-specific and keeps the user's access/refresh tokens in the session, so it is **not** on the default viewset: `OnaOpenIDConnectViewset` is unchanged, and a deployment opts in by pointing `VIEWSET_CLASS` at the new subclass. The organization-account rejection is inherited either way.

- `test_urls_without_admin.py` pins both halves of the opt-in: the default viewset exposes none of the seven proxy actions and its paths 404, the Keycloak subclass exposes all of them with `stash_oidc_tokens=True`, and the subclass still refuses organization accounts. Also pins that `session` does not prefix-match `sessions`.
- Whether the proxy is routed is silent either way — no error, the routes simply exist or don't — which is why these are asserted rather than left to a smoke test.
- Covered upstream by ona-oidc's suite (294 tests, green on the pinned commit).

- Depends on ona-oidc `0697676`, pinned here. That change renames the OIDC session keys with no fallback read, so **live sessions get one forced re-login** at deploy.
- The `/oidc/<server>/account` route is gone; ona-oidc removed that action.
- ona-oidc now requires a server-side `SESSION_ENGINE` for the proxy viewset and applies its `Origin` allowlist to reads, so `LOGIN_REDIRECT_ALLOWED_HOSTS` must contain the SPA's exact netloc — host **and** port — in every environment.
- No behaviour change for existing deployments: the default viewset is untouched and `VIEWSET_CLASS` still points at it. Deployments wanting the Account tab set it to `OnaKeycloakOpenIDConnectViewset` via their own settings overlay.
- `str_to_bool` upstream now reads `"false"`, `"0"`, `""`, `"no"` and `"off"` as false; previously only the exact literal `"False"` was. Settings here are real booleans, so nothing changes for this repo — but a deployment that feeds any `OPENID_CONNECT_VIEWSET_CONFIG` flag from an env var was getting the opposite of what it asked for, and will now get what it asked for.
- Supersedes the bump in #3176, which can be closed.

  - [x] Included tests
  - [x] Updated documentation

Closes #

### DIFF
--- a/onadata/apps/main/oidc_viewsets.py
+++ b/onadata/apps/main/oidc_viewsets.py
@@ -7,13 +7,20 @@ establish a session via SSO. The base viewset looks a user up by the email or
 username claim coming from the IdP and calls ``django.contrib.auth.login``
 directly, bypassing ``onadata.apps.main.backends.ModelBackend``. Guarding only
 the backend is therefore not enough; the rejection must also happen here.
+
+The Keycloak Account REST proxy is *not* part of the default viewset. It is
+Keycloak-specific and stores the user's access and refresh tokens in the
+session, so a deployment opts in by pointing ``VIEWSET_CLASS`` at
+``OnaKeycloakOpenIDConnectViewset`` below.
 """
 
 from django.conf import settings
 from django.utils.translation import gettext as _
 
+from oidc.keycloak import KeycloakAccountMixin
 from oidc.viewsets import SSO_COOKIE_NAME, UserModelOpenIDConnectViewset
 from rest_framework import status
+from rest_framework.decorators import action
 from rest_framework.response import Response
 
 from onadata.apps.main.models.user_profile import UserProfile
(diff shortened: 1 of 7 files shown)


FrankApiyo:
Superseded. These routes served Keycloak's Account REST API for the linked-accounts and sessions UI, both of which have been dropped. The one fix worth keeping — re-declaring the `@action` that `OnaOpenIDConnectViewset.login()` dropped — is carried in onaio/onadata#3207. Note the ona-oidc bump this PR folded in is *not* carried; #3207 still pins 8f39ec02.

PR #3186, opened 2026-07-24
feat(entities): create/update Entities from repeats

### LINKED ISSUE #3184: Create/update Entities from repeats

[Repeats](https://xlsform.org/en/#repeats) make it possible to capture data about multiple things of the same kind within the context of a single form.

Users want to create Entities from those repeat so that they can follow up with them. Currently, a **single form submission** can only create or update a **single** Entity. 

We propose, in addition to the current feature, to create/update **MULTIPLE** Entities in a **SINGLE** Entity List from a repeat

Ability to create multiple Entities from a repeat so that users can follow up with them.

## Suggested implementation plan (Steps to be taken to implement feature)

Read the [documentation](https://docs.getodk.org/entities-quick-reference/#create-or-update-multiple-entities-in-the-same-list) and add support for the feature.

Constraints to beware of:

- **ALL** `save_to` values **MUST** be in the **SAME** repeat (or groups in that repeat)
- `label` and other expressions on the entities sheet may **ONLY** reference fields in the repeat (or groups in that repeat).

kelvin-muchiri:

Add support for creating/updating Entities  from repeats

**Constraints**

- **ALL** `save_to` values **MUST** be in the **SAME** repeat (or groups in that repeat)
- `label` and other expressions on the entities sheet may **ONLY** reference fields in the repeat (or groups in that repeat). 

- [x] QA

None

  - [x] Included tests
  - [ ] Updated documentation

Closes #3184

---
<a href="https://app.blacksmith.sh/onaio/codesmith/onadata/pr/3186"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787489489&installation_model_id=422582&pr_number=3186&repository=onaio%2Fonadata&return_to=https%3A%2F%2Fgithub.com%2Fonaio%2Fonadata%2Fpull%2F3186&signature=baedbebb609c4711c308b3f12c059d114d8aa3483f8197c5d61c1e8581f708cd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/onadata/apps/logger/tests/test_parsing.py
+++ b/onadata/apps/logger/tests/test_parsing.py
@@ -2,25 +2,26 @@
 import json
 import os
 import re
+
+from django.test import SimpleTestCase
+
 from defusedxml import minidom
 
-from onadata.apps.main.tests.test_base import TestBase
-from onadata.apps.logger.xform_instance_parser import (
-    XFormInstanceParser,
-    xpath_from_xml_node,
-)
-from onadata.apps.logger.xform_instance_parser import (
-    get_uuid_from_xml,
-    get_meta_from_xml,
-    get_deprecated_uuid_from_xml,
-)
-from onadata.libs.utils.common_tags import XFORM_ID_STRING
 from onadata.apps.logger.models.xform import XForm
 from onadata.apps.logger.xform_instance_parser import (
+    XFormInstanceParser,
     _xml_node_to_dict,
     clean_and_parse_xml,
+    get_deprecated_uuid_from_xml,
+    get_entity_group_data,
+    get_entity_label_from_node,
+    get_entity_nodes_from_xml,
+    get_meta_from_xml,
+    get_uuid_from_xml,
+    xpath_from_xml_node,
 )
-
+from onadata.apps.main.tests.test_base import TestBase
+from onadata.libs.utils.common_tags import XFORM_ID_STRING
 
 XML = "xml"
 DICT = "dict"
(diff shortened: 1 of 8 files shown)


ukanga (changes requested):
1. **[P1] Empty repeat rows shift Entity/property association** (`onadata/apps/logger/xform_instance_parser.py`, around lines 548-552).

`_get_node_index()` counts every repeat in the XML, but `instance.get_dict()` omits repeats containing only empty fields and a self-closing update Entity node. For a valid update where Entity A has a blank optional property and the following Entity B has `year_planted=2016`, `repeat_data[0]` contains B data while the XML index for A is also 0. A is therefore updated with B properties, and B receives an empty patch.

Please extract values from the Entity node containing XML repeat, or preserve positional placeholders in the parsed repeat data. Add a regression test with a blank repeat before a populated repeat.

2. **[P2] Nested repeats lose Entity properties** (`onadata/apps/logger/xform_instance_parser.py`, around lines 545-552).

For an Entity inside a nested repeat such as `household/member`, the member list is nested inside each household item. Looking up `instance_data.get("household/member")` fails at the top level, so the function ascends to `household` and returns the outer repeat dictionary. `get_entity_json_from_instance()` then cannot see the inner member fields, producing Entities with labels but without their saved properties.

ukanga:
### Medium severity

- [`onadata/apps/logger/xform_instance_parser.py:543`](https://github.com/onaio/onadata/blob/3544b42a57e7c59254ad68f7ad5495f73f3e0af3/onadata/apps/logger/xform_instance_parser.py#L543) — Entity properties represented as CDATA are silently omitted because collection only accepts `TEXT_NODE`, despite the existing instance parser supporting `CDATA_SECTION_NODE`.

  **Suggested fix:** Treat both text and CDATA nodes as scalar values and add create/update coverage for CDATA-backed entity properties.

FrankApiyo (approved):

PR #3207, opened 2026-08-10
feat(2fa): two-factor authentication with secrets encrypted at rest

FrankApiyo:

Two-factor authentication moves into onadata with **both second-factor secrets encrypted at rest**. Login runs through the wizard at `/account/login/` (`/accounts/login/` redirects to it); enrolment stays off until a deployment sets `ENABLE_TWO_FACTOR`.

- **Encrypted at rest** — the authenticator seed and recovery codes are stored under a Fernet key held in config (`TWO_FACTOR_FIELD_ENCRYPTION_KEYS`), not the database. Both are encrypted rather than hashed because the seed must be recoverable to compute codes and unspent recovery codes are re-viewable. `rotate_two_factor_encryption_key` re-encrypts under a new key in batches; a startup check refuses to boot with `ENABLE_TWO_FACTOR` on and no usable key.
- **`/api/v1/totp/`** — `status`, `enroll/start`, `enroll/confirm`, `disable`, `verify`, `recovery/generate`, `recovery`. Weakening or replacing a factor needs a current code or a `verify` grant minted for that one operation; responses are `Cache-Control: no-store`; the owner is emailed on factor changes and repeated failures.
- **First enrolment** demands the account password (`TWO_FACTOR_ENROLMENT_REQUIRES_PASSWORD`, on by default) and accepts only the interactive sign-in credential, so an API token cannot enrol a factor and lock the owner out.
- CSRF is enforced on the ambient sign-in cookie (header/API-token credentials stay exempt); the failed-login lockout now covers both wizard steps; a partial unique constraint enforces one authenticator per user.

`django-two-factor-auth`, `django-otp` and `qrcode` become declared dependencies.

Driven through the wizard's own POSTs with codes from a real HOTP implementation; single-use and enrolment races covered by `TransactionTestCase` tests; encryption, key rotation, the startup check and fail-closed decryption each have tests. Exercised by hand against a running stack.

- An account with no factor sees the plain password form — nothing changes until they enrol.
- `ENABLE_TWO_FACTOR` requires a valid key with no plaintext fallback; existing plaintext django-otp enrolments must re-enrol.
- `/admin/` still needs only staff permission, so an un-enrolled staff user gets in on a password alone.
- #3215 is stacked on this branch.

  - [x] Included tests
  - [x] Updated documentation

Closes #


onadata/apps/api/viewsets/totp_viewset.py
```diff
@@ -0,0 +1,673 @@
...
+
+    def _spend_code_attempt(self, request):
+        """Count a failed code against the bounded login allowance.
+
+        A threshold hit raises, which the next request sees as ``locked_out``;
+        the current response stays the standard invalid-code message.
+        """
+        try:
+            add_login_attempt(get_client_ip(request), request.user.username)
+        except AuthenticationFailed:
```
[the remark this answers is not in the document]
FrankApiyo: Addressed in 4747b2e8e: kept the `except AuthenticationFailed` catch but replaced the bare `pass` with an explanatory comment and an explicit `return`, so the control flow is unchanged and the security rationale is inline. Applied the same treatment to the matching broker-guard in `notify_owner` to pre-empt the duplicate finding.

ukanga (dismissed):

onadata/apps/api/tasks.py
```diff
@@ -137,6 +137,12 @@ def send_account_lockout_email(email, message_txt, subject):
     send_generic_email(email, message_txt, subject)
 
 
+@app.task(base=AutoRetryTask)
+def send_two_factor_changed_email(email, message_txt, subject, message_html=None):
```
kelvin-muchiri, line 141: The convention we've used for naming async tasks, albeit not always respected, is appending `_async` suffix. i.e `send_two_factor_changed_email_async`

onadata/libs/utils/email.py
```diff
@@ -85,6 +88,32 @@ def get_account_lockout_email_data(username, ip_address, end=False):
...
+    """Generates the email notifying a two-factor account event.
+
+    ``event`` names the template set under ``two_factor/``; extra context
+    reaches the templates. A plain-text and an HTML rendering are returned so
+    the message goes out as multipart/alternative.
+    """
+    now = timezone.now().astimezone(dt_timezone.utc)
+    ctx_dict = {
+        "username": username,
+        "support_email": getattr(settings, "SUPPORT_EMAIL", "support@example.com"),
```
kelvin-muchiri, line 101: Is `support@example.com` a good default. We'd rather not display it in the email or raise an `ImproperlyConfigured` error

onadata/libs/utils/email.py
```diff
@@ -110,13 +139,15 @@ def get_account_deactivation_email_data(
     }
 
 
-def send_generic_email(email, message_txt, subject):
-    """Sends an email."""
+def send_generic_email(email, message_txt, subject, message_html=None):
```
kelvin-muchiri, line 142: Did we update unit tests for method `send_generic_email` after this change?

onadata/apps/main/registration_urls.py
```diff
@@ -48,15 +47,13 @@
...
+    # password-only bypass of the second factor. The name is kept so existing
+    # reverse("auth_login") call sites still resolve. Declared first so it
+    # takes precedence over the ``login/`` route in ``registration.auth_urls``.
     path(
         "login/",
-        LoginView.as_view(
-            template_name="registration/login.html",
-            authentication_form=LoginLockoutAuthenticationForm,
-        ),
+        RedirectView.as_view(pattern_name="two_factor:login", query_string=True),
```
kelvin-muchiri, line 56: @FrankApiyo Will this continue to work even with `ENBALE_TWO_FACTOR=False`

onadata/apps/api/tests/viewsets/test_totp_concurrency.py
```diff
@@ -0,0 +1,355 @@
+"""Concurrency tests for one-time-code verification.
```
kelvin-muchiri, line 1: Do we have a corresponding `totp_concurrency_viewset` in package `onadata/apps/api/viewsets` ? I couldn't find it hence this test module might be misplaced

requirements/base.pip
```diff
@@ -119,6 +119,8 @@ django-debug-toolbar==7.1.1
 django-digest @ git+https://github.com/onaio/django-digest.git@6bf61ec08502fd3545d4f2c0838b6cb15e7ffa92
```
kelvin-muchiri, line 119: `requirements/dev.pip` not yet updated with the new requirements

onadata/apps/api/models/encrypted_recovery_device.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
+"""Recovery codes, encrypted at rest so the owner can view them again.
```
kelvin-muchiri, line 2: Let's add a corresponding test module at `onadata/apps/api/tests/models`

onadata/apps/api/models/encrypted_totp_device.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
+"""A TOTP authenticator whose seed is encrypted at rest.
```
kelvin-muchiri, line 2: Let's add a corresponding test module at `onadata/apps/api/tests/models`

onadata/apps/api/tests/test_encrypted_two_factor.py
```diff
@@ -0,0 +1,246 @@
+"""Encrypted authenticator seed and recovery codes, plus key rotation."""
+
```
kelvin-muchiri, line 2: What module does this file test?

onadata/libs/utils/email.py
FrankApiyo: Going to remove the default

onadata/apps/api/tasks.py
FrankApiyo: Done — renamed to `send_two_factor_changed_email_async` (definition, import, and call site).

onadata/libs/utils/email.py
FrankApiyo: Dropped the placeholder default: `get_two_factor_email_data` now requires `SUPPORT_EMAIL`, and a startup system check (`api.E003`) fails fast when two-factor is enabled without it — rather than mailing users an address they cannot reach.

onadata/libs/utils/email.py
FrankApiyo: Added unit tests for `send_generic_email`: plain-text only, the HTML alternative part, and the required-args guard.

onadata/apps/main/registration_urls.py
FrankApiyo: Yes. `two_factor:login` is registered unconditionally (its own URLConf, not gated by `ENABLE_TWO_FACTOR`) and served by `LockoutLoginView`, so this redirect resolves and password-only login still works with the flag off. `ENABLE_TWO_FACTOR` only gates the enroll/manage API.

onadata/apps/api/tests/viewsets/test_totp_viewset_concurrency.py
```diff
@@ -0,0 +1,355 @@
+"""Concurrency tests for one-time-code verification.
```
FrankApiyo: Renamed to `test_totp_viewset_concurrency.py` — it exercises `TOTPViewSet` verification under concurrent connections, kept as its own module because it needs `TransactionTestCase`.

requirements/base.pip
FrankApiyo: Regenerated `dev.pip`; the two-factor deps are now present and pinned to the same versions as `base.pip`.

onadata/apps/api/models/encrypted_recovery_device.py
FrankApiyo: Added `onadata/apps/api/tests/models/test_encrypted_recovery_device.py`.

onadata/apps/api/models/encrypted_totp_device.py
FrankApiyo: Added `onadata/apps/api/tests/models/test_encrypted_totp_device.py`.

onadata/apps/api/tests/test_encrypted_two_factor.py
FrankApiyo: Split it: the per-model tests moved to `tests/models/test_encrypted_totp_device.py` and `test_encrypted_recovery_device.py`; this file was renamed to `test_two_factor_key_rotation.py` and now holds only the key-rotation command and the startup-check tests.

kelvin-muchiri (approved):

PR #3217, opened 2026-08-17
feat(messaging): audit log profile changes

kelvin-muchiri:

- Changes to a user's or organization's profile are now recorded, capturing who made the change and which details changed, with the previous and new values. Contact details (email, phone number) are recorded as changed without their values. The record is available through the messaging endpoint.
- User and organization profiles are now version tracked, so earlier versions of a profile can be inspected and restored.

- [x] Added tests

None.

  - [x] Included tests
  - [x] Updated documentation

### DIFF
--- a/docs/messaging.rst
+++ b/docs/messaging.rst
@@ -39,6 +39,8 @@ This endpoint provides access to event messages sent for a specific target. Wher
 
     - ``entitylist_deleted``
 
+    - ``require_auth_changed``
+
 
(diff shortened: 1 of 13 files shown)


ukanga (commented):
Can it be just general? That a change happened to the user profile, but perhaps detect what changes so that it covers not just require_auth but other changes?

ukanga (approved):

PR #3242, opened 2026-08-31
fix: make Enketo request timeout configurable and handle request failures

FrankApiyo:

Production requests to Enketo (e.g. generating a submission edit URL) can exceed the hardcoded 20s timeout; the resulting `ReadTimeout` propagates uncaught and clients get a 500 ([Sentry event](https://ona-systems.sentry.io/issues/124831738/events/1548f7be8a4d4d18bd7b4b765f8ee079/)).

- New `ENKETO_API_REQUEST_TIMEOUT` setting (default 20s) replaces the hardcoded timeout, so slow Enketo deployments can raise it without a code change.
- `get_enketo_urls()` now reports request failures (timeout, connection error) to Sentry and raises `EnketoError`, which callers already handle — clients get the friendly "Sorry, we cannot load your form right now" message instead of a 500.
- Non-object JSON bodies from Enketo (in error or success responses) also divert into the `EnketoError` path instead of crashing on `.get()`.

The first commit is lint/format churn (pre-commit hooks had not previously run on these files).

Mocked `SimpleTestCase` tests: `ReadTimeout` and `ConnectionError` raise `EnketoError` with the generic message and Sentry reference, and the configured timeout is passed to the request. Verified via CI.

Enketo network failures now surface as a 400 with the generic Enketo error message rather than a 500.

Note for reviewers: since #3213 the ruff and isort pre-commit hooks disagree on Django import sorting and undo each other on any file with Django imports. This branch keeps the repo-standard isort layout; aligning ruff's isort config is left for a separate PR.

  - [x] Included tests
  - [ ] Updated documentation

### DIFF
--- a/onadata/apps/api/tests/viewsets/test_data_viewset.py
+++ b/onadata/apps/api/tests/viewsets/test_data_viewset.py
@@ -68,6 +68,7 @@ from onadata.libs.serializers.submission_review_serializer import (
 )
 from onadata.libs.utils.common_tags import MONGO_STRFTIME
 from onadata.libs.utils.logger_tools import create_instance
+from onadata.libs.utils.viewer_tools import ENKETO_GENERIC_ERROR
 
 
 @urlmatch(netloc=r"(.*\.)?enketo\.ona\.io$")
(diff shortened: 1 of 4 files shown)


ukanga:
Can we preserve lazy evaluation of `response.text` here? Using `data.get("message", response.text)` reintroduces a regression previously fixed in [PR #3031](https://github.com/onaio/onadata/pull/3031), specifically commit [`a5eb24aaf`](https://github.com/onaio/onadata/commit/a5eb24aaf957f151cf75bf638e3f12e121110162). Because `dict.get()` eagerly evaluates its default argument, this currently causes `TypeError: string argument without an encoding` in three existing tests.

Could we retain the new non-dictionary handling with:

```python
message = (
    data["message"]
    if isinstance(data, dict) and "message" in data
    else response.text
)
```

Can we also add an endpoint-level regression test that makes `requests.post` raise `ReadTimeout`, calls `/api/v1/data/{pk}/{dataid}/enketo`, and asserts the established 400 response with the generic Enketo error message?

ukanga (approved):

PR #3243, opened 2026-08-31
feat(2fa): encrypt the authenticator seed and recovery codes at rest

FrankApiyo:

Encrypts both second-factor secrets at rest with a symmetric key held in application config (no external KMS/Vault dependency), the way GitLab holds its OTP secret:

- **Authenticator seed** — a project-owned `EncryptedTOTPDevice` replaces django-otp's `TOTPDevice`, whose seed column is plaintext. The seed is decrypted only to compute a code; `verify_token`/`config_url` mirror django-otp so the login wizard and enrolment payload are unchanged. A data migration re-encrypts existing enrolments in place.
- **Recovery codes** — stored encrypted rather than hashed, so the owner can re-view the unspent codes (as GitHub and Google allow). The `recovery` endpoint now returns the unspent codes behind the same step-up every code-weakening route takes (a current code/grant), with `no-store` and an audit entry on each view.
- **Keys** — `TWO_FACTOR_FIELD_ENCRYPTION_KEYS` is an ordered Fernet key list (first encrypts, all decrypt), so a new key rolls in ahead of the old and `manage.py rotate_two_factor_encryption_key` re-encrypts every seed and code before the old key is dropped. Replaces the previous recovery-code pepper.

- New tests for the encryption util (round-trip, per-call IV, rotation), both devices (seed encrypted-at-rest, TOTP verify + replay rejection, recovery re-view/spend/case-fold), the data migration, and the rotation command.
- Existing second-factor tests adapted to the encrypted models; full second-factor suite green (122 tests).

- `TWO_FACTOR_FIELD_ENCRYPTION_KEYS` must be set wherever `ENABLE_TWO_FACTOR` is on (and before migrating any existing enrolments).
- Recovery codes generated before this change were stored as one-way hashes and cannot be migrated; those users regenerate their codes once. Authenticators keep working.

  - [x] Included tests
  - [x] Updated documentation

---
<a href="https://app.blacksmith.sh/onaio/codesmith/onadata/pr/3243"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790752267&installation_model_id=422582&pr_number=3243&repository=onaio%2Fonadata&return_to=https%3A%2F%2Fgithub.com%2Fonaio%2Fonadata%2Fpull%2F3243&signature=4cea17c577b158b19b1b26264ac5c8708648e8cf80e560d6c8cf57c3dfba0140"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


FrankApiyo:
Collapsed directly into the base branch (feat/django-two-factor-auth): the second-factor secrets are now encrypted from a single migration, with no hashed model or create-then-drop migration left behind. Superseded by that change.

PR #3260, opened 2026-09-17
fix: let Briefcase download attachments of private forms

### LINKED ISSUE #3259: Briefcase attachment downloads fail with 403 Forbidden on private forms
### Environmental Information

- Onadata version: v5.26.0 and later (reproduced by code inspection on v5.29)

### Problem description

ODK Briefcase can no longer pull attachments from private forms. Submissions download normally, but every attachment request fails and Briefcase logs:

```
Error downloading attachment 1 of 1 of submission 189 of 316: Forbidden
```

The attachment links handed to Briefcase in the submission download response point at the `/attachment/original?media_file=...&attachment_id=...` endpoint. Since v5.26.0 that endpoint checks that the requester has access to the form, but it only recognises a browser session. Briefcase authenticates with Digest credentials, which the endpoint ignores, so the request is treated as anonymous and answered with `403 Not shared.`.

Briefcase only sends its credentials after receiving a `401` with a Digest challenge. Because the endpoint answers `403` with no challenge, Briefcase never gets the chance to authenticate and gives up on the attachment.

The `/media/<path>` endpoint has the same behaviour, which affects deployments that serve attachments from local file storage.

Forms with data sharing turned on are not affected.

### Expected behavior

A Briefcase user with access to a private form can download its attachments:

- An unauthenticated attachment request for a private form is answered with `401` and a Digest challenge, so that clients such as Briefcase can retry with credentials.
- A request carrying valid Digest (or API token) credentials for a user with access to the form gets the attachment.
- An authenticated user without access to the form still gets `403`.

1. Create a private form that collects an attachment (e.g. a photo) and make a submission with one.
2. In ODK Briefcase, connect to the server as a user who has access to the form.
3. Pull the form.
4. Open the detailed status: submissions download, every attachment fails with `Forbidden`.

### Additional Information

Possible approach: have the attachment and media endpoints accept the same credentials as the API (Digest and token) in addition to the browser session, and answer anonymous requests to private attachments with a `401` Digest challenge instead of `403`.

Workaround until fixed: temporarily turn on data sharing for the form while pulling, which exposes the data publicly for that period.

kelvin-muchiri:

ODK Briefcase could not pull attachments from private forms. Submissions downloaded normally, but every attachment failed with "Forbidden".

Attachment links only recognised people signed in through the browser. Briefcase signs in a different way, so it was treated as an anonymous visitor and turned away, without ever being asked for its username and password.

Attachment and media links now ask for a username and password when an anonymous visitor requests something private, and accept the same sign-in methods as the rest of the API. Briefcase users with access to a form can download its attachments again. People who are signed in but do not have access to the form are still refused.

- [x] Added tests

These include one that pulls a submission the way Briefcase does and then downloads its attachment. It fails with the reported "Forbidden" error without this change.

Someone who is not signed in and opens a link to a private attachment, export or form file in their browser now sees the browser's sign-in prompt instead of a "Not shared" message.

Rejected sign-ins on these links do not count towards account lockout, the same as the other links Briefcase pulls from. A large pull makes one request per attachment, and a request can be rejected for reasons other than a wrong password, so counting them could lock a legitimate user out.

  - [x] Included tests
  - [ ] Updated documentation

Closes #3259


kelvin-muchiri:
### Code review

Found 1 issue:

1. Attachment and media links now run Digest authentication but are not exempt from failed-login lockout, unlike every other Briefcase endpoint. `authenticate_media_request` calls `DigestAuthentication.authenticate`, which counts any Digest validation failure as a failed login unless the path is in `LOCKOUT_EXCLUDED_PATHS`. That list covers `downloadSubmission`, `submissionList`, `formList` and the rest because ODK clients hit them constantly, but not `/attachment/` or `/media/`. A failure does not have to be a wrong password: an expired nonce or an out-of-order nonce count from interleaved downloads is counted the same way. Briefcase requests one attachment link per file, so a large pull could reach the attempt limit and lock the account out of everything, including the endpoints this PR is meant to unblock. Not reproduced against a real Briefcase client; the added tests perform a fresh handshake per request, so they never reuse a nonce. (bug due to the lockout exemption added in e45f1e588, "Ignore ODK APIs when checking for user lockout")

https://github.com/onaio/onadata/blob/53dd7f592d7a49a66ac4890d424c49721ea01373/onadata/libs/authentication.py#L296-L306

https://github.com/onaio/onadata/blob/53dd7f592d7a49a66ac4890d424c49721ea01373/onadata/libs/authentication.py#L60-L76

https://github.com/onaio/onadata/blob/53dd7f592d7a49a66ac4890d424c49721ea01373/onadata/libs/authentication.py#L578-L588

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

FrankApiyo (approved):

